PR #18, opened 2022-04-04
Updated action to run on demand

TheDanMiller:
Updating this to be manually dispatched


gmrabian (approved):
https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

PR #31, opened 2022-04-08
Configure a11y testing

braxex:
## Description

Just setting up accessibility testing per [MDCT-95](https://qmacbis.atlassian.net/browse/MDCT-95), which calls out `axe` and `pa11y` specifically. 

Axe testing already existed within Cypress thanks to `axe-core` and `cypress-axe`. So I just added the other two using the `cypress-audit` lib.

### How to test

1. `cd tests/cypress`
2. `yarn install`
3. `yarn test` & follow prompts per usual for E2E testing

### Dependencies

Added `@cypress-audit/pa11y` & `@cypress-audit/lighthouse`. 

## Code author checklist
- [x] I have added [thorough tests](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research)

^^ Not yet, as I'm looking for comments first.

### DIFF
--- a/tests/cypress/support/commands.js
+++ b/tests/cypress/support/commands.js
@@ -1,7 +1,9 @@
+import "cypress-file-upload";
+import "@cypress-audit/pa11y/commands";
+
 before(() => {
   cy.visit("/", { timeout: 60000 * 5 });
 });
-import "cypress-file-upload";
 
 const emailForCognito = "//input[@name='email']";
 const passwordForCognito = "//input[@name='password']";
@@ -40,6 +42,8 @@ function terminalLog(violations) {
 // axe api documentation: https://www.deque.com/axe/core-documentation/api-documentation/
 Cypress.Commands.add("checkA11yOfPage", () => {
   cy.wait(3000);
+
+  // check accessibility using axe-core
   cy.injectAxe();
   cy.checkA11y(
     null,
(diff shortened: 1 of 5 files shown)


tests/cypress/support/commands.js
```diff
@@ -56,4 +61,15 @@ Cypress.Commands.add("checkA11yOfPage", () => {
      */
     true // does not fail tests for ally violations
   );
+
+  // check accessibility using pa11y
+  cy.pa11y({
+    threshold: 10,
```
braxex, line 67: Can set a threshold of number of errors that you're ok with to still have it pass. We can have this be whatever number we want, but I wanted to start a discussion about it first. You can also set limits for notices and warnings. 

See the whole list of [pa11y options here](https://github.com/pa11y/pa11y#configuration).

tests/cypress/support/commands.js
```diff
@@ -56,4 +61,15 @@ Cypress.Commands.add("checkA11yOfPage", () => {
...
+    standard: "WCAG2AA",
+  });
+
+  // check accessibility using lighthouse
+  cy.lighthouse({
+    accessibility: 90,
```
braxex, line 73: Same concept as above. The lighthouse plugin can check lots of other things, but I've got it set currently to only care about accessibility. Right now it's set to fail if the a11y check scores less than 90/100, but again, wanted to start a discussion. We may decide to have different limits for local checks vs. CI checks, too.

BearHanded (approved):

tests/cypress/support/commands.js
BearHanded: 90 seems like a good one to aim for right now

PR #34, opened 2022-04-15
Cleanup

braxex:
## Description

There was a bunch of stuff in here that we did not want to be in here. So we removed it.

### How to test

1. Pull down branch
2. `nvm use`
3.  `./dev local`

Everything should still be up and running. Maybe not looking the greatest, but we'll get to that later.

### Dependencies

No changed deps.

## Code author checklist
- [x] ~~I have added [thorough tests](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research)~~


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -9,7 +9,7 @@
 
 ## Code author checklist
 - [ ] I have performed a self-review of my code
-- [ ] I have added [thorough tests](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research)
+- [ ] I have added [thorough](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research) tests
```
TheDanMiller, line 12: :trollface:

TheDanMiller (approved):

karla-vm (approved):
Looks good and runs on my local environment 👍🏽

.github/PULL_REQUEST_TEMPLATE.md
braxex: 😂

package.json
```diff
@@ -13,14 +13,15 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/CMSgov/macpro-quickstart-serverless.git"
+    "url": "https://github.com/CMSgov/mdct-mcr"
```
bodnarbm, line 16: nit: The repository should be the value that you would pass to git clone. For `https` clones that would be `https://github.com/CMSgov/mdct-mcr.git`
```suggestion
    "url": "https://github.com/CMSgov/mdct-mcr.git"
```

bodnarbm (commented):
My nit below is just a nit, but 

I would highly encourage renaming `.scripts` to `scripts` or leaving the scripts at the root of the repo. This is to increase visibility and discoverability as placing them in the a `.` folder will make them hidden by default to most terminal commands.

Additionally, I recommend against the file renames proposed here generally unless we plan to make these changes across all the MACPro repos. Otherwise developers across the repos will have to remember which repos they run which commands in and increases the complexity of sharing changes to common workflows / scripts as need to remember changing the names to match the naming convention for that repo.

package.json
braxex: Good catch. That one's my bad for just copy/pasting and forgetting to add the `.git`.

braxex:
@bodnarbm 
>I would highly encourage renaming .scripts to scripts or leaving the scripts at the root of the repo. This is to increase visibility and discoverability as placing them in the a . folder will make them hidden by default to most terminal commands.

Totally willing to change from `.scripts` --> `scripts`. 👍

>Additionally, I recommend against the file renames proposed here generally unless we plan to make these changes across all the MDCT repos. Otherwise developers across the repos will have to remember which repos they run which commands in and increases the complexity of sharing changes to common workflows / scripts as need to remember changing the names to match the naming convention for that repo.

Yeah, I see what you're saying. We were annoyed at the differing naming conventions so made some changes to standardize, but it's not a massive issue to me. I'd like to clean up all the repos but that doesn't seem particularly feasible at this point in the engagement. Let me make some changes and I'll ping you for a second look.

bodnarbm (approved):

PR #42, opened 2022-04-29
add integration & a11y tests for login/logout functionality

braxex:
## Description

Adds integration & accessibility testing for the Cognito login/logout functionality, and remediates existing basic accessibility issues.

Satisfies the following ticket: [MDCT-260: Integration tests for login functionality](https://qmacbis.atlassian.net/browse/MDCT-260)

Note: Incidentally had to add a new colorScheme for buttons to the theme object to remediate a color contrast accessibility issue.

### How to test

Prerequisite: Ensure correct `.env` files (see [readme](https://github.com/CMSgov/mdct-mcr#environment-configuration)).

Integration tests:
1. `cd` && `nvm use`
2. `yarn test`
3. When cypress window launches, select one of the added test suites (`accessibility > login.spec.js`, `e2e > login.spec.js`).
4. All tests should be passing and there should be no accessibility errors or warnings.

### Changed Dependencies

No changed deps.

## Code author checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,44 @@ When run locally, auth bypasses Cognito. The frontend mimics login in local stor
 
 ### Running the app locally
 
-- Populate .env files in the root directory and in the `ui-src` directory.
+- Populate .env files in the root directory and in the `ui-src` directory. See Environment Configuration below for up-to-date files.
 - Run all the services locally with the command `./dev local`
 - Troubleshooting: See the Requirements section if the command asks for any prerequisites you don't have installed.
+
+#### Environment Configuration
+
+Root `.env`
+```
+SKIP_PREFLIGHT_CHECK=true
+LOCAL_LOGIN=true
+MEASURE_TABLE_NAME=local-measures
+MEASURE_TABLE_ARN=local_nonsense_if_unset_we_search_CF_for
+measureTableName=local-measures
+DYNAMODB_URL=http://localhost:8000
+API_URL=http://localhost:3030/local
+S3_LOCAL_ENDPOINT=http://localhost:4569
+S3_ATTACHMENTS_BUCKET_NAME=local-uploads
+URL=http://localhost/3000
+DISABLE_ESLINT_PLUGIN=true
+```
+
+/services/ui-src `.env`
+```
+LOCAL_LOGIN=true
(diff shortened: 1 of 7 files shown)


gmrabian (approved):
lgtm - tests passing

README.md
```diff
@@ -40,10 +40,45 @@ When run locally, auth bypasses Cognito. The frontend mimics login in local stor
...
+DISABLE_ESLINT_PLUGIN=true
+```
+
+/services/ui-src `.env`
+```
+LOCAL_LOGIN=true
+API_REGION=us-east-1
+API_URL=http://localhost:3030
+COGNITO_REGION=us-east-1
+COGNITO_IDENTITY_POOL_ID=us-east-1:61c849d8-f95c-45af-8131-15b75c525034
```
bodnarbm, line 70: All the information here is for a different project than MCR. Can you please correct with the appropriate information for MCR.

services/ui-src/src/components/logins/LoginCognito.tsx
```diff
@@ -40,30 +40,34 @@ export const LoginCognito = () => {
...
-      </Heading>
-      <Input
-        id="password"
-        name="password"
-        type="password"
-        value={fields.password}
-        onChange={handleFieldChange}
-        className="field"
-      />
+      <label>
```
bodnarbm, line 43: Blocking: As far as I understand the a11y documentation, headings should not be nested under label's as it can break some navigation aids. Thus I think we need to either please the label inside the heading and than target the input or need to use something other than an heading for the label styling.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#headings

bodnarbm (changes requested):
Just the two blocking requests and then this looks good.

services/ui-src/src/components/logins/LoginCognito.tsx
braxex: Ah yeah good catch. I'll swap it out.

bodnarbm (approved):
@braxex this LGTM. Could we update the branch with the latest from master to make sure the cypress jobs run in CI. And then I think we are good.

bodnarbm (approved):
Nice!

PR #2601, opened 2022-05-26
add error handling

braxex:
## Description

Adds some basic error handling so we can catch those errors and show 'em to the people.

### How to test

1. Pull and run.
2. Try to log in with invalid credentials. You should get a browser alert with an error message.

### Changed Dependencies

No changed deps.

## Code author checklist

### DIFF
--- a/services/.sechub/handlers/sync.js
+++ b/services/.sechub/handlers/sync.js
@@ -59,6 +59,16 @@ async function getAllActiveFindings() {
                 Value: "ACTIVE",
               },
             ],
+            WorkflowStatus: [
+              {
+                Comparison: "EQUALS",
+                Value: "NEW",
+              },
+              {
+                Comparison: "EQUALS",
+                Value: "NOTIFIED",
+              },
+            ],
             SeverityLabel: severityLabels,
           },
           MaxResults: 100,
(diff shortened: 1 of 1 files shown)


services/ui-src/src/utils/auth/useUser.ts
```diff
@@ -4,10 +4,11 @@ import { UserContext } from "./userContext";
 export const useUser = (): any => {
   const context = useContext(UserContext);
 
-  if (context === undefined) {
+  if (typeof context === "undefined") {
```
bodnarbm, line 7: Why this change? Just wondering for my own learning. I have a preference for direct checks against undefined or loose equality check against `null` (e.g., `null == context`). But just wondering what this typeof covers over the string equality with `undefined`?

gmrabian (approved):

bodnarbm (approved):

services/ui-src/src/utils/auth/useUser.ts
braxex: They're very similar, but using `typeof` it won't throw an error if the variable hasn't been declared, so I always default to using `typeof` and strict equality when checking for undefined. It's an especially common pattern in SSR apps to guard server-side vs. client-side code by checking for the window obj, like:

```
if (typeof window !== 'undefined') {
  // do whatever
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#typeof_operator_and_undefined

PR #2614, opened 2022-05-26
fix rule of hooks issue (+0.1%)

braxex:
## Description

There was a bug introduced in `eb055df` where we were violating the rule of hooks by conditionally rendering the AdminBanner provider. This PR fixes that by moving the provider post-auth.

Screen Shot 2022-05-26 at 18 46 18

### How to test

1. Pull and run. The error should not appear in the console. Everything should still work.

### Changed Dependencies

No changed deps.

## Code author checklist
- [x] ~~I have added [thorough](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research) tests~~

### DIFF
--- a/services/ui-src/src/components/app/AppRoutes.tsx
+++ b/services/ui-src/src/components/app/AppRoutes.tsx
@@ -1,32 +1,44 @@
 import { Navigate, Route, Routes } from "react-router-dom";
 import { Admin, Help, Home, NotFound, Profile } from "../../views";
-// components
-import { AdminBanner } from "components/banners/AdminBanner";
 // utils
 import { UserRoles } from "utils/types/types";
+import {
+  AdminBannerProvider,
+  AdminBannerContext,
+} from "components/banners/AdminBannerProvider";
 
 export const AppRoutes = ({ userRole }: Props) => {
   const isAdmin = userRole === UserRoles.ADMIN;
-  const adminBanner = AdminBanner();
 
   return (
     <main id="main-wrapper">
-      <Routes>
-        <Route path="/" element={<Home adminBanner={adminBanner} />} />
-        <Route
-          path="/admin"
-          element={
-            !isAdmin ? (
-              <Navigate to="/profile" />
-            ) : (
-              <Admin adminBanner={adminBanner} />
-            )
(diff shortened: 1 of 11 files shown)


services/ui-src/src/components/app/AppRoutes.tsx
```diff
@@ -1,10 +1,14 @@
...
+// components
+import { AdminBanner } from "components/banners/AdminBanner";
 // utils
-import { AdminBannerShape, UserRoles } from "utils/types/types";
+import { UserRoles } from "utils/types/types";
 
-export const AppRoutes = ({ adminBanner, userRole }: Props) => {
+export const AppRoutes = ({ userRole }: Props) => {
   const isAdmin = userRole === UserRoles.ADMIN;
+  const adminBanner = AdminBanner();
```
bodnarbm, line 10: Injecting the component down the stack seems odd to me in the scenario. What about taking the fetching and other state logic out of the component and placing it in a context instead? Then the context can be grabbed anywhere further down in the stack as needed, and the component itself becomes pure presentation.

gmrabian (approved):

gmrabian:
@braxex also wanted to call out App.test.tsx currently mocks out an AdminBanner for the initialization. To keep things clean we should remove that, and consider creating an AppRoutes.test file that can reuse it now.

services/ui-src/src/components/app/AppRoutes.tsx
braxex: Yeah that's a good idea. I made that change, but then have been running into some weird issues related to [act()](https://reactjs.org/docs/test-utils.html#act) in the testing. Ideally I'd like to use context, but there's something above my head going on with the hooks that makes React upset. You can get the tests to pass locally but it still throws errors about it.

braxex: Gonna hold off on useContext for now and re-evaluate later.

braxex:
Added AppRoutes.test.tsx file.

gmrabian (approved):
me gusta

PR #2747, opened 2022-06-09
fix up data-testid passing & references

braxex:
## Description

We've been overusing data-testids and in some cases passing them explicitly as props, which doesn't make a ton of sense. Nothing really changes functionally here, but it cleans things up a bit. 👍 

Note: In some places I've changed testids from being passed on every instance of a component and pulled them into only being placed in test files as needed.

### How to test

Just see that all the tests pass.

### Changed Dependencies

No changed deps.

## Code author checklist


gmrabian (approved):
so far so good. just need to merge main

PR #2762, opened 2022-06-09
MDCT-374 Auth fix for Cognito vulnerability

gmrabian:
## Description
[MDCT-374](https://qmacbis.atlassian.net/browse/MDCT-374)

Fixes Synack finding **[AWS Access Token configuration allows write access to Access Management Information](https://client.synack.com/vulnerabilities/anubis-w031-2)**

Removes `aws.cognito.signin.user.admin` scope from authenticated users. Now when they log in they can no longer exploit the access token to modify their own user attributes. 

Also updated user info fetch pattern because the scope is what allowed us to pull it directly. Now we parse it out of the id token. This is safe because the details from this token are only used for frontend display. It cannot be exploited to change roles or escalate permissions in backend api calls. Fixed various locations in frontend where user info is referenced to match updated fetch pattern.

_Note: we cannot fully test this on the branch because the aws-amplify library automatically adds that role when we log in with username + password. To test it we need to login with IDM which we can only do once its merged. Brandon has confidence this will fix it, though we can't know yet._

### How to test

Need to merge to test :/

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/app/App.test.tsx
+++ b/services/ui-src/src/components/app/App.test.tsx
@@ -7,12 +7,15 @@ import {
   mockNoUser,
   RouterWrappedComponent,
 } from "utils/testing/setupJest";
-import { useUser } from "utils/auth";
 //components
 import { App } from "components";
 
-jest.mock("utils/auth");
-const mockedUseUser = useUser as jest.Mock<typeof useUser>;
+jest.mock("utils/auth", () => ({
+  useUser: jest
+    .fn()
+    .mockReturnValueOnce(mockStateUser)
+    .mockReturnValue(mockNoUser),
+}));
 
 const appComponent = (
   <RouterWrappedComponent>
@@ -20,13 +23,8 @@ const appComponent = (
   </RouterWrappedComponent>
 );
 
-beforeEach(() => {
-  mockedUseUser.mockImplementation((): any => mockStateUser);
-});
-
 describe("Test App", () => {
   test("App is visible", async () => {
-    mockedUseUser.mockImplementation((): any => mockStateUser);
     await act(async () => {
       await render(appComponent);
     });
@@ -34,7 +32,6 @@ describe("Test App", () => {
   });
 
   test("App renders local logins if there is no user", async () => {
-    mockedUseUser.mockImplementation((): any => mockNoUser);
     await act(async () => {
       await render(appComponent);
     });
--- a/services/ui-src/src/components/app/App.test.tsx
+++ b/services/ui-src/src/components/app/App.test.tsx
@@ -7,15 +7,12 @@ import {
   mockNoUser,
   RouterWrappedComponent,
 } from "utils/testing/setupJest";
+import { useUser } from "utils/auth";
 //components
 import { App } from "components";
 
-jest.mock("utils/auth", () => ({
-  useUser: jest
-    .fn()
-    .mockReturnValueOnce(mockStateUser)
-    .mockReturnValue(mockNoUser),
-}));
+jest.mock("utils/auth");
+const mockedUseUser = useUser as jest.MockedFunction<typeof useUser>;
 
 const appComponent = (
   <RouterWrappedComponent>
@@ -25,6 +22,7 @@ const appComponent = (
 
 describe("Test App", () => {
   test("App is visible", async () => {
+    mockedUseUser.mockReturnValue(mockStateUser);
     await act(async () => {
       await render(appComponent);
     });
@@ -32,6 +30,7 @@ describe("Test App", () => {
   });
 
   test("App renders local logins if there is no user", async () => {
+    mockedUseUser.mockReturnValue(mockNoUser);
     await act(async () => {
       await render(appComponent);
     });
--- a/services/ui-src/src/components/app/App.tsx
+++ b/services/ui-src/src/components/app/App.tsx
@@ -5,7 +5,7 @@ import { Container, Divider, Flex, Heading, Stack } from "@chakra-ui/react";
 import { AppRoutes, Footer, Header, LoginCognito, LoginIDM } from "components";
 
 export const App = () => {
-  const { logout, user, showLocalLogins, loginWithIDM } = useUser();
+  const { logout, user, showLocalLogins } = useUser();
   return (
     <div id="app-wrapper">
       {user && (
--- a/services/ui-src/src/components/app/App.tsx
+++ b/services/ui-src/src/components/app/App.tsx
@@ -5,14 +5,14 @@ import { Container, Divider, Flex, Heading, Stack } from "@chakra-ui/react";
 import { AppRoutes, Footer, Header, LoginCognito, LoginIDM } from "components";
 
 export const App = () => {
-  const { logout, user, userRole, showLocalLogins, loginWithIDM } = useUser();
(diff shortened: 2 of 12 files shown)


braxex (approved):

services/ui-src/src/components/app/App.tsx
```diff
@@ -5,14 +5,14 @@ import { Container, Divider, Flex, Heading, Stack } from "@chakra-ui/react";
...
-  const { logout, user, userRole, showLocalLogins, loginWithIDM } = useUser();
+  const { logout, user, showLocalLogins, loginWithIDM } = useUser();
   return (
     <div id="app-wrapper">
       {user && (
         <Flex sx={sx.appLayout}>
           <Header handleLogout={logout} />
           <Container sx={sx.appContainer} data-testid="app-container">
-            <AppRoutes userRole={userRole} />
+            <AppRoutes userRole={user?.userRole} />
```
bodnarbm, line 15: `user?.userRole` may be undefined here, but `userRole` for the `AppRoutes` props is defined as requiring a string. This would be a type error if `useUser` was property defined to return a `UserContextInterface` instead of the `any`. Suggest removing the `any` from the return type for the `useUser` and then changing `userRole` in `AppRoutes` props to take an optional string.

bodnarbm: Also the `?` here is not needed as this part of the code will run only if `user` is truthy and thus must be a defined `MCRUser`

services/ui-src/src/utils/auth/userContext.tsx
```diff
@@ -1,22 +1,20 @@
 import { createContext } from "react";
-import { CognitoUser } from "@aws-amplify/auth";
-
-export interface CustomCognitoUser extends CognitoUser {
-  role: string;
+export interface MCRUser {
+  email: string;
+  given_name: string;
+  family_name: string;
+  userRole: string | undefined;
```
bodnarbm, line 6: Prefer `userRole?: string` to match our other optional use below (`state?`).

Also, I don't think userRole should ever not be set. Maybe when we construct the `MCRUser` object if `userRole` is not present in IDM then we present some error page to the user asking them to contact the help desk for an issue with their role? I think that is work for later though and not part of this PR.

services/ui-src/src/utils/auth/userProvider.tsx
```diff
@@ -35,8 +34,27 @@ export const UserProvider = ({ children }: Props) => {
 
   const checkAuthState = useCallback(async () => {
     try {
-      const authenticatedUser = await Auth.currentAuthenticatedUser();
-      setUser(authenticatedUser);
+      const session = await Auth.currentSession();
+      const { email, given_name, family_name } = await session.getIdToken()
```
bodnarbm, line 38: Don't need the `await` here.

services/ui-src/src/utils/auth/userProvider.tsx
```diff
@@ -35,8 +34,27 @@ export const UserProvider = ({ children }: Props) => {
...
+        .payload;
+      // "custom:cms_roles" is an string of concat roles so we need to check for the one applicable to MCR
+      const userRole = (
+        (await session.getIdToken().payload?.["custom:cms_roles"]) as
+          | string
+          | undefined
```
bodnarbm, line 44: Since we are calling pulling the payload off multiple times, it may help to store it in it's own local variable for readability.

Something like this maybe

```
const payload = session.getIdToken().payload;
const { email, given_name, family_name } = payload;
....
const cms_role = payload["custom:cms_roles"] as string ?? "";
const userRole = cms_role.split(",").find((r) => r.includes("mdctmcr"));
...
```

bodnarbm (changes requested):
Overall this approach looks great. Just have a few code change requests to help with readability. Good to merge after those changes.

Also, I tested this with the IDM flow having the local react application point to the main cognito pool / client, and this worked / closed the issue as expected. Nice work.

services/ui-src/src/components/app/App.test.tsx
```diff
@@ -7,34 +7,31 @@ import {
...
-import { useUser } from "utils/auth";
 //components
 import { App } from "components";
 
-jest.mock("utils/auth");
-const mockedUseUser = useUser as jest.Mock<typeof useUser>;
+jest.mock("utils/auth", () => ({
+  useUser: jest
+    .fn()
+    .mockReturnValueOnce(mockStateUser)
```
bodnarbm, line 16: This will depend on test execution order. When mock setup varies between test, it is better to do the setups within the individual tests themselves.

gmrabian: It was that way before. Tests were actually passing with the changes, but typescript didn't like it
I'll give it another shot

services/ui-src/src/views/Profile/Profile.test.tsx
```diff
@@ -18,32 +17,18 @@ const profileView = (
 
 // MOCKS
 
-jest.mock("utils/auth");
-const mockedUseUser = useUser as jest.Mock<typeof useUser>;
+jest.mock("utils/auth", () => ({
+  useUser: jest
+    .fn()
+    .mockReturnValueOnce(mockAdminUser)
```
bodnarbm, line 23: See other comments about test execution order.

services/ui-src/src/views/Profile/Profile.tsx
```diff
@@ -5,16 +5,9 @@ import { Button, Flex, Heading, Text } from "@chakra-ui/react";
...
-  const { email, given_name, family_name } = userInfo.user.attributes;
-  const { userRole } = userInfo;
-  const state = userInfo.user.attributes?.["custom:cms_state"] || "";
-  return { email, given_name, family_name, userRole, state };
-};
-
 export const Profile = () => {
-  const { email, given_name, family_name, userRole, state } = userDetails();
+  const { user } = useUser();
+  const { email, given_name, family_name, userRole, state } = user!;
```
bodnarbm, line 10: Prefer

```
const { email, given_name, family_name, userRole, state } = user ?? {};
```

over the ! post-fix. Using the ! post-fix will force the type compiler to assume this is never undefined, and thus the types of required properties like email will be typed as string and not possibly undefined. This can lead to unexpected runtime errors as the type system won't know email could be undefined and will let you run unchecked methods on it. Using `?? {}` instead will ensure that `email` will be typed as `string | undefined` and thus require handing / checking before calling string methods on it.

bodnarbm (changes requested):
Some minor changes related to testing and type safety.

services/ui-src/src/utils/auth/userProvider.test.tsx
```diff
@@ -10,25 +10,23 @@ import { RouterWrappedComponent } from "utils/testing/setupJest";
...
   Auth: {
-    currentAuthenticatedUser: jest
+    currentSession: jest
       .fn()
-      .mockImplementationOnce(() => {
-        throw new Error("failed!");
-      })
-      .mockImplementationOnce(() => {
-        throw new Error("failed!");
+      .mockReturnValueOnce({
```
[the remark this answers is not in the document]
gmrabian: This one was more complicated so I left it out for now 😓

bodnarbm (approved):
Looks great @gmrabian!

braxex (approved):

PR #2866, opened 2022-06-15
MDCT-368 Add scroll to top for route change and submission

gmrabian:
## Description
[MDCT-368](https://qmacbis.atlassian.net/browse/MDCT-368). Add scroll to top ability. On route change page will reset to top with no delay or visible movement.

On form submission (admin banner is the only form right now), the movement is visible. (Note there is a delay here as it waits for the write to finish)

### How to test
Run locally. Navigate to the help page and scroll to the bottom. Click on the logo to go to the home. Notice you're at the top. You can toggle between these, scrolling down in between (compare to main).

Submit banner. Wait and see it scroll up.

### Changed Dependencies
none

## Code author checklist


gmrabian:
Don't let the yarn numbers fool you. This is small.

services/ui-src/package.json
```diff
@@ -34,6 +34,7 @@
     "react-query": "^3.34.5",
     "react-router-dom": "6.0.2",
     "react-scripts": "^5.0.0",
+    "react-scroll": "^1.8.7",
```
bodnarbm, line 37: Do we need this additional library? Looking at [the react router guide](https://v5.reactrouter.com/web/guides/scroll-restoration), it seems like we should be able to handle this with existing dom methods.

services/ui-src/src/utils/scroll/scrollToTop.ts
```diff
@@ -0,0 +1,27 @@
...
+  return null;
+};
+
+const scrollOnRouteChange = () => {
+  animateScroll.scrollToTop({
+    duration: 0,
+  });
+};
+
+export const scrollToTopFormSubmit = () => {
```
bodnarbm, line 21: If we are keeping the user on the same page post submit, should we be scrolling the user to the top? It seems like unexpected behavior. If we are bouncing them back to the top (like old form submits that had to hit a backend server and then reload the page), I think we are also fine with no animation here and just setting them to top of page (just like the page reloaded).

services/ui-src/package.json
braxex: This is something we discussed a bit, but `react-scroll` allows you greater control over the scroll behavior ([options](https://github.com/fisshy/react-scroll#propsoptions)) and will be useful down the line when design wants to target specific elements as scroll destinations and handle fancy scrolling behaviors in plan-specific drawers within the form.

services/ui-src/src/utils/scroll/scrollToTop.ts
braxex: Hmm... interesting thought. I agree that we definitely need to put them at the top of the page so it's evident that the operation succeeded. But it doesn't necessarily have to be a scroll and we could just set them to the top. Might be something to run past the designers.

braxex (approved):

services/ui-src/package.json
gmrabian: (See Brax's comments below):

We did originally have it using window.scrollTo, like react router, but we used this because it more easily exposes options for animations, duration, delay, etc. If you think there's never a need to animate, I'll change it back. Just did this because it'd be easier for form stuff in the future (related to your question below: we may not need that anyway)

bodnarbm: Could we defer bringing in the additional dependency until its features are needed? I think the dom's scrollTo on both the window and element objects (along with the optional `smooth` behavior) covers everything that is currently needed.

bodnarbm (approved):

services/ui-src/src/utils/scroll/scrollToTop.ts
bodnarbm: We can test it out and adjust as needed. I'm good the current animated scroll.

gmrabian:
Closing to open a new request - resolve yarn changes

PR #2908, opened 2022-06-17
MDCT 397: Fix avscan

gmrabian:
## Description
[MDCT 397](https://qmacbis.atlassian.net/browse/MDCT-397)

### How to test
- Log into MCR dev AWS
- Go to `uploads-fix-scan-attachments-446712541566` bucket in S3
- Drag a file in and upload it
- Navigate to CloudWatch and find the avscan lambda logs `/aws/lambda/uploads-fix-scan-avScan`
- Find the log associated with your upload (should be most recent)
- Confirm scan passed 🎉 

For additional confirmation, open Cloudshell and enter this command
```
aws s3api get-object-tagging --bucket uploads-fix-scan-attachments-446712541566 --key {YOUR_FILE_NAME}
```
See if the tag comes back CLEAN

### Changed Dependencies
None

## Code author checklist


gmrabian:
Editing in three parts - Managed Care Review (a _different_ MCR) has fixed these things already.

1. [Sanitizing file path](https://github.com/CMSgov/managed-care-review/commit/8e9a68c961f7dc8456dd8d0606bb00c30059dbf7). Rather than giving the uploaded file name (which should be safe because it comes from an admin, but this is good security), this sanitizes the name to only be alphanumeric. S3 still recognizes the file without the `.` between the name and file type.
2. In the same PR above you can see their exec statement contains `-d /tmp/`. Previously we only had `-d`. This just runs the daemon in the tmp directory.
3. Adding the `s3:DeleteObject` permission to the lambdas [since we need that permission for the virus scanner](https://github.com/CMSgov/mdct-mcr/blob/main/services/uploads/src/clamav.js#L109-L142)

services/uploads/serverless.yml
```diff
@@ -219,91 +222,6 @@ resources:
...
-                    - s3:PutObjectAcl
-                    - s3:PutObjectTagging
-                    - s3:PutObjectVersionTagging
-                    - s3:ListBucket
-                  Resource: !Sub ${ClamDefsBucket.Arn}/*
-                - Effect: "Allow"
-                  Action:
-                    - s3:ListBucket
-                  Resource: !GetAtt ClamDefsBucket.Arn
-
```
gmrabian, line 306: These are deleted because they're unused, and have been as long as they've been in this file. The permissions are defined above under provider.statements

services/uploads/serverless.yml
```diff
@@ -24,16 +24,19 @@ provider:
...
+            - s3:GetObject
+            - s3:GetObjectTagging
+            - s3:PutObject
+            - s3:PutObjectAcl
+            - s3:PutObjectTagging
+            - s3:PutObjectVersionTagging
+            - s3:DeleteObject
+            - s3:ListBucket
+          Resource:
+            - !Sub arn:aws:s3:::${self:service.name}-${self:custom.stage}-attachments-${AWS::AccountId}/*
```
gmrabian, line 40: Obviously don't need dynamo permissions here. Changing to S3 with the proper ones we need on that bucket

services/uploads/src/antivirus.js
```diff
@@ -32,13 +32,26 @@ async function isS3FileTooBig(s3ObjectKey, s3ObjectBucket) {
...
+ * @param {string} s3ObjectKey Key of the s3 object
+ * @return {string} Path of the s3 object on the local fs
+ *
+ */
+function pathFromObjectKey(s3ObjectKey) {
+  // remove problematic characters from filename
+  let sanitizedFilename = s3ObjectKey.replace(/[^a-zA-Z0-9]/g, "");
+  return `${downloadDir}/${path.basename(sanitizedFilename)}`;
+}
+
```
gmrabian, line 48: Sanitizes input file name just in case. Removes `.` too but S3 still reads the file with the name and extension concatenated.

services/uploads/serverless.yml
```diff
@@ -219,91 +217,6 @@ resources:
             - ServerSideEncryptionByDefault:
                 SSEAlgorithm: AES256
         AccessControl: Private
-    BucketAVScanRole:
```
gmrabian, line 222: These two roles are deleted because they're unused, and have been as long as they've been in this file. The permissions are defined above under provider.statements

gmrabian:
Also tested against the first MCPAR template provided and it passed the scan.

services/uploads/src/utils.js
```diff
@@ -96,6 +96,8 @@ function extractBucketFromApiEvent(s3Event) {
  */
 function generateSystemMessage(systemMessage) {
   let finalMessage = `--- ${systemMessage} ---`;
+  // eslint-disable-next-line no-console
+  console.log(finalMessage);
```
braxex, line 100: Still need this?

services/uploads/src/clamav.js
```diff
@@ -191,10 +191,11 @@ async function uploadAVDefinitions() {
  */
 function scanLocalFile(pathToFile) {
   try {
-    execSync(
+    let avResult = execSync(
       `${constants.PATH_TO_CLAMAV} -v -a --stdout -d /tmp/ ${pathToFile}`
     );
 
+    console.log(avResult.toString()); // eslint-disable-line no-console
```
braxex, line 198: Still need this?

braxex (approved):
Looks great!

services/uploads/src/utils.js
gmrabian: Without this we don't see any logs on what it's doing

services/uploads/src/clamav.js
gmrabian: Yep this is just a good double check that the virus scan worked and didn't have a wacky error that didn't throw on our end

services/uploads/src/antivirus.js
```diff
@@ -32,13 +32,26 @@ async function isS3FileTooBig(s3ObjectKey, s3ObjectBucket) {
   return fileSize > constants.MAX_FILE_SIZE;
 }
 
+let downloadDir = constants.DOWNLOAD_DIR;
```
bodnarbm, line 35:
```suggestion
const downloadDir = constants.DOWNLOAD_DIR;
```

Do we alter this anywhere, or good to make this a const?

bodnarbm (approved):

PR #3029, opened 2022-07-07
merge dev to val

braxex:
Merge dev to val. Contains role mapping fix.


gmrabian (approved):
DON'T SQUASH!!! BE WARY THE MINDLESS CLICK

PR #3064, opened 2022-07-09
add hydration mechanism to formJson parsing

braxex:
## Description

To display fields that have been pre-filled with data (whether disabled or not) while maintaining the static nature of the form json, we need a hydration mechanism to fill the field values with dynamic data.

Changes:
- Added hydration utility function to handle hydrating form fields.
- Added temporary fake data to `McparReportPage` for state and program name. These are both things we can get from the database once we have program creation figured out.
- Modified existing forms to conform to modified types and add in necessary hydration data.
- Tweaked some related styles and types.

### How to test

1. Navigate to `/mcpar/program-information/point-of-contact`. Observe that field A1 has been hydrated with dynamic data.
2. Complete form with valid fill and continue. Observe that field A6 has been hydrated with dynamic data.
3. All tests pass.

### Changed Dependencies

n/a

## Code author checklist

- [x] I have done the lgtm review


services/ui-src/src/verbiage/forms/adminBannerForm.ts
```diff
@@ -152,6 +157,7 @@ export const form = {
     {
       id: "abf-endDateYear",
       type: "child",
+      props: null,
```
gmrabian, line 160: Hmm I wish we didn't have to add these. Is there a way around it? It creates noise in these files, unless I'm not understanding the benefit.

services/ui-src/src/utils/forms/forms.test.ts
```diff
@@ -1,32 +1,79 @@
...
+
+  it("Hydrates with 'ERROR' if data unavailable", () => {
+    const hydratedFormFields = hydrateFormFields(
+      mockFormFields.filter((field) => field.id === "mock-field-2"),
+      mockHydrationData
+    );
+    const hydratedFieldValue = hydratedFormFields.find(
+      (field) => field.id === "mock-field-2"
+    )?.props!.value;
+    expect(hydratedFieldValue).toEqual("ERROR");
```
gmrabian, line 77: I don't quite understand how this works. Shouldn't the hydration value be empty for mock-field-2

gmrabian (approved):

services/ui-src/src/utils/forms/forms.test.ts
gmrabian: Sure thing:
For history's sake my question was actually about having the hydrate value populated when there was no data available anyway (why not take hydrate out for that object?). Brax clarified the hydrate value is a key lookup for what will eventually be a database key, that will contain the real value. So in this case since "unavailableData" doesn't exist in the array { mockData: "the mocked data" };, it returns ERROR (the array being the stand-in for the database)

PR #3069, opened 2022-07-10
remove need for duplicate name/id

braxex:
## Description

It was annoying to have to add `id` and `name` to fields when realistically they should always be the same (`label` is what display the visible field name, not `name`). Now that's solved for. The added complexity of the "is type child" `if/else` can be removed after we rework the date field and no longer need `React.Fragment` for unseen child components.

Changes: 
- Added duplication of `field.id` as both `id` and `name` in the form factory so it doesn't have to be done manually.
- Removed the now superfluous `name` key from fields in all current forms.

### How to test

1. Log in as admin and nav to /admin. Test banner validation and update. Everything still works.

### Changed Dependencies

n/a

## Code author checklist
- [x] ~~I have added [thorough](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research) tests~~


ntsummers1 (approved):
LGTM!

PR #3096, opened 2022-07-13
MDCT-471: ChoiceList component does not correctly update form state/field value onChange

ntsummers1:
## Description
Fixes [MDCT-471](https://qmacbis.atlassian.net/browse/MDCT-471). The Choicelist component is updated to keep track of its individual choices and return them

### How to test

Before testing, add:

```
{
      id: "abf-radio",
      type: "radioList",
      required: true,
      props: {
        name: "Choice List Radio Name",
        type: "radio",
        label: "Choice List Radio Label",
        choices: [
          { label: "Choice 1", value: "A" },
          { label: "Choice 2", value: "B" },
          { label: "Disabled choice 3", value: "C" },
        ],
      },
      validation: {
        type: "array",
        options: { min: 1 },
        errorMessages: {
          min: "Description text is required",
        },
      },
    },
    {
      id: "abf-check",
      type: "checkboxList",
      props: {
        name: "Choice List Checkbox Name",
        type: "checkbox",
        label: "Choice List Checkbox Label",
        choices: [
          { label: "Choice 1", value: "A" },
          { label: "Choice 2", value: "B" },
          { label: "Disabled choice 3", value: "C" },
        ],
      },
      validation: {
        type: "array",
        options: { min: 1 },
        errorMessages: {
          min: "Description text is required",
        },
      },
    },
```

to adminBannerForm.ts. You'll want to add it as part of the "Fields" array

- ./dev local
- Sign in as an admin
- My Account -> Manage Account -> Banner Editor
- Play with the checkboxes and console log form.getValues() in the useEffect hook

## Code author checklist


services/ui-src/src/verbiage/forms/adminBannerForm.ts
```diff
@@ -4,6 +4,40 @@ export const form = {
...
+      id: "abf-radio",
+      type: "choiceList",
+      props: {
+        name: "Choice List Radio Name",
+        type: "radio",
+        label: "Choice List Radio Label",
+        choices: [
+          { label: "Choice 1", value: "A", defaultChecked: true },
+          { label: "Choice 2", value: "B" },
+          { label: "Disabled choice 3", value: "C", disabled: true },
```
braxex, line 17: I've talked to design and we will not be using `defaultChecked` or `disabled` on ChoiceList choices, so no need to support those for MVP. 👍

services/ui-src/src/styles/index.scss
```diff
@@ -55,6 +55,12 @@ a {
   background-color: var(--chakra-colors-palette-gray_lightest);
 }
 
+// ChoiceList Override
+
+.ds-c-choice[type="checkbox"]:checked::after {
+  box-sizing: content-box;
+}
+
```
braxex, line 63: We can prob do this in the `sx` in ChoiceList. I'll help.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -15,14 +16,31 @@ export const ChoiceListField = ({
...
+  const onChangeHandler = (event: InputChangeEvent) => {
+    const choiceSelected: ChoiceListSelected = {
+      value: event.target.value,
+      checked: event.target.checked,
+    };
+    setChoicesChosen((prevState) => {
+      return prevState.map((choice) => {
+        return choice.value === choiceSelected.value ? choiceSelected : choice;
+      });
+    });
```
braxex, line 43: The plan has been to make a `CheckboxField` component and a `RadioField` component, both comprised of a `ChoiceListField` component after the `ChoiceList` was done. We may want to do that in this PR and consider having separate `onChangeHandler` functions for each.

services/ui-src/src/styles/index.scss
ntsummers1: Yeah... this was a weird one. I decided on this method since it'll occur every single time we want to use a checkbox choicelist or choice and it felt weird to include that as an sxOveride prop every time. Or if we didn't want to do that, create another box to simply apply this sx onto. Happy to go either way tbh

services/ui-src/src/components/fields/ChoiceListField.tsx
ntsummers1: Is there a different output we would like to have? Right now both Radio and Checkbox's return this format:

```
abf-radio: Array(3)
0: {value: 'A', checked: false}
1: {value: 'B', checked: true}
2: {value: 'C', checked: false}
```

ntsummers1: Settled on this format over a conversation with Brax and Garrett:

Radio:
```
abf-a1o1: Array(1)
0: {value: 'C', id: 'abf-radio03'}
```

Checkbox
```
abf-a2o1: Array(3)
0: {value: 'A', id: 'abf-checkboxo1'}
1: {value: 'B', id: 'abf-checkbox02'}
2: {value: 'C', id: 'abf-checkbox03'}
```

Both of these cases are only showing array values that are checked. Unchecked is not shown in the array.

ntsummers1:
@braxex I need to write some better tests (They're currently pretty wack) for this PR, but hows this PR now look overall? I don't want to hold you up from your work on this.

Also the updated output for radio and checkbox should now be the expected format:

```
Radio
abf-radio: Array(1)
0: {value: 'A', id: 'radio_abf-radio_8'}

Checkbox
abf-check: Array(2)
0: {value: 'A', id: 'checkbox_abf-check_12'}
1: {value: 'C', id: 'checkbox_abf-check_14'}
```

services/ui-src/src/styles/index.scss
braxex: What I would do is build it out as an `sx` object in the component file, like: 

```
".ds-c-choice[type='checkbox']:checked::after":  {
  box-sizing: content-box;
}
```

and then have it style the component alongside the override, like:

```
<Component sx={{ ...sx, ...sxOverride }} />
```

services/ui-src/src/verbiage/forms/adminBannerForm.ts
```diff
@@ -4,6 +4,40 @@ export const form = {
...
+        choices: [
+          { label: "Choice 1", value: "A" },
+          { label: "Choice 2", value: "B" },
+          { label: "Disabled choice 3", value: "C" },
+        ],
+      },
+      validation: {
+        type: "array",
+      },
+    },
```
braxex, line 40: Remember to take these out before you finish up. 

Also will you check out validation for these? Adding in `required` might not work directly, but we could try using `minLength` of 1 or something similar.

services/ui-src/src/components/fields/CheckboxField.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  useEffect(() => {
+    form.setValue(name, choicesChosen, { shouldValidate: true });
+  }, [choicesChosen]);
+
+  // get the form context
+  const form = useFormContext();
+
+  // update form data
+  const onChangeHandler = (event: InputChangeEvent) => {
```
braxex, line 33: The handler is what CodeClimate wants to see tested. Inclusive lines 34-45.

services/ui-src/src/styles/index.scss
ntsummers1: Fixed!

services/ui-src/src/verbiage/forms/adminBannerForm.ts
ntsummers1:
```
validation: {
  type: "array",
  options: { min: 1 },
  errorMessages: {
    min: "Field is required",
  },
},
```
This will check to make sure that the choicelist is filled out with at least 1 option

ntsummers1: minLength does not work here. Despite the docs telling me it would 🙃

braxex: So this validation does not technically work because the field is not being registered with the form correctly, which is kind of a larger issue we need to look into. That's also why the date field validation is not working correctly. We also don't want fields to be showing error messages before the user interacts with them. For right now we can merge this PR and move forward but we need to come back to this next.

braxex (approved):

PR #3121, opened 2022-07-18
a couple thoughts

braxex:
proof of concept for how we can have both regex and end date >= start date validation living side by side.


services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -1,11 +1,11 @@
-import { useEffect, useState } from "react";
+import { useState } from "react";
```
braxex, line 1: we can avoid using `useEffect` here if we do a direct check in the `onChangeHandler` before setting any form values.

services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -1,11 +1,11 @@
...
 import { useFormContext } from "react-hook-form";
 // components
 import { SingleInputDateField as CmsdsDateField } from "@cmsgov/design-system";
 import { Box } from "@chakra-ui/react";
 // utils
 import { AnyObject } from "types";
-import { makeMediaQueryClasses } from "utils";
+import { checkDateCompleteness, makeMediaQueryClasses } from "utils";
```
braxex, line 8: New method from time utils.

services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -14,16 +14,20 @@ export const DateField = ({ name, label, sxOverride, ...props }: Props) => {
   const form = useFormContext();
   form.register(name);
 
-  const [displayValue, setDisplayValue] = useState("");
+  const [displayValue, setDisplayValue] = useState<string>("");
+  const [formattedValue, setFormattedValue] = useState<string>("");
```
braxex, line 18: tracking these separately so we can access the formatted valued in `onBlur`, which is necessary so we can check for incomplete dates `onBlur` and tell users the entered date is invalid.

services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -14,16 +14,20 @@ export const DateField = ({ name, label, sxOverride, ...props }: Props) => {
...
-      form.setValue(name, displayValue || "", { shouldValidate: true });
+  const onChangeHandler = (rawValue: string, formattedValue: string) => {
+    setDisplayValue(rawValue);
+    setFormattedValue(formattedValue);
+    const completeDate = checkDateCompleteness(formattedValue);
+    if (completeDate) {
+      form.setValue(name, formattedValue, { shouldValidate: true });
     }
-  }, [displayValue]);
+  };
```
braxex, line 27: setting both values in local state, then checking if the date is complete. only if it is complete will it run a validation check. this is to stop the previous behavior of yelling at a user immediately when they start entering a date, for instance just typing `9`. now it will wait to set the form value and validate until it is either 1) a complete date or 2) they leave the field prior to entering a complete date, triggering `onBlur`

services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -14,16 +14,20 @@ export const DateField = ({ name, label, sxOverride, ...props }: Props) => {
...
 
-  const onChangeHandler = async (inputtedString: string) => {
-    setDisplayValue(inputtedString);
+  const onBlurHandler = () => {
+    form.setValue(name, formattedValue, { shouldValidate: true });
   };
```
braxex, line 31: allows us to validate the field on blur so we can tell users that partial dates are not valid.

services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -34,7 +38,8 @@ export const DateField = ({ name, label, sxOverride, ...props }: Props) => {
         name={name}
         label={label}
         onChange={onChangeHandler}
-        value={displayValue || props.hydrate}
+        onBlur={onBlurHandler}
+        value={displayValue || props.hydrate || ""}
```
braxex, line 42: added the empty string here because without it the empty string was not getting set as the initial state and it was throwing a nasty React controlled vs uncontrolled switch error.

services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -45,6 +50,7 @@ export const DateField = ({ name, label, sxOverride, ...props }: Props) => {
 interface Props {
   name: string;
   label: string;
+  timetype?: string;
```
braxex, line 53: added for ease in using a revised method in `utils/time`

services/ui-src/src/components/forms/AdminBannerForm.tsx
```diff
@@ -5,7 +5,7 @@ import { ErrorAlert, Form, PreviewBanner } from "components";
 // utils
 import { bannerId } from "../../constants";
 import { REPLACE_BANNER_FAILED } from "verbiage/errors";
-import { convertDatetimeStringToNumber } from "utils/time/dateTimeConversion";
+import { convertDatetimeStringToNumber } from "utils";
```
braxex, line 8: changed import path to just `utils`

services/ui-src/src/components/forms/AdminBannerForm.tsx
```diff
@@ -21,11 +21,11 @@ export const AdminBannerForm = ({ writeAdminBanner, ...props }: Props) => {
       link: formData["abf-link"],
       startDate: convertDatetimeStringToNumber(
         formData["abf-startDate"],
-        "abf-startDate"
+        formJson.fields.find((el) => el.id === "abf-startDate")!.props.timetype!
       ),
       endDate: convertDatetimeStringToNumber(
         formData["abf-endDate"],
-        "abf-endDate"
+        formJson.fields.find((el) => el.id === "abf-endDate")!.props.timetype!
```
braxex, line 28: A changed method in `utils/time` now needs the "timeType" rather than the field name. More on this below.

services/ui-src/src/constants.ts
```diff
@@ -1,2 +1,6 @@
 // BANNERS
 export const bannerId = "admin-banner-id";
+
+// REGEX
+export const dateFormatRegex =
+  /^((0[1-9]|1[0-2])\/(0[1-9]|1\d|2\d|3[01])\/(19|20)\d{2})|((0[1-9]|1[0-2])(0[1-9]|1\d|2\d|3[01])(19|20)\d{2})$/;
```
braxex, line 6: Just moved to constants so we can reference it later. _Excellent_ regex-ing my man. 👍

services/ui-src/src/forms/internal/abf.json
```diff
@@ -32,14 +32,16 @@
...
+        "timetype": "startDate"
       }
     },
     {
       "id": "abf-endDate",
       "type": "date",
       "props": {
-        "label": "End date"
+        "label": "End date",
+        "timetype": "endDate"
```
braxex, line 44: added the optional "timetype" prop so we can use it later

services/ui-src/src/forms/internal/abf.schema.ts
```diff
@@ -12,5 +10,15 @@ export default object({
...
+    .test(
+      "is after start date",
+      "End date cannot be before start date",
+      (endDateString, context) => {
+        const startDateString = context.parent["abf-startDate"];
+        const startDate = new Date(startDateString);
+        const endDate = new Date(endDateString!);
+        return endDate >= startDate;
+      }
+    ),
```
braxex, line 23: This is the new test for checking startDate vs. endDate. Had to do it this way to play nicely with our new string type.

services/ui-src/src/utils/other/time.ts
```diff
@@ -13,12 +13,14 @@ export const noon: TimeShape = {
...
+): TimeShape => {
   const timeMap: any = {
-    "abf-startDate": midnight,
-    "abf-endDate": oneSecondToMidnight,
+    startDate: midnight,
+    endDate: oneSecondToMidnight,
   };
-  return timeMap[dateType] || noon;
+  return timeMap?.[timeType as keyof typeof timeMap] || noon;
 };
```
braxex, line 24: Updated to be field-name agnostic and instead rely on matching the passed 'timeType'.

services/ui-src/src/utils/other/time.ts
```diff
@@ -57,3 +59,24 @@ export const formatDateUtcToEt = (date: number): string => {
...
   return `${month + 1}/${day}/${year}`;
 };
+
+export const checkDateCompleteness = (date: string) => {
+  const month = parseInt(date.split("/")?.[0]);
+  const day = parseInt(date.split("/")?.[1]);
+  const year = parseInt(date.split("/")?.[2]);
+  const dateIsComplete = month && day && year.toString().length === 4;
+  return dateIsComplete ? { year, month, day } : null;
+};
```
braxex, line 69: New method to check if a date is complete, because it was needed in a couple different places.

services/ui-src/src/utils/other/time.ts
```diff
@@ -57,3 +59,24 @@ export const formatDateUtcToEt = (date: number): string => {
...
+  timeType: string | undefined
+): number | undefined => {
+  const completeDate = checkDateCompleteness(date);
+  let convertedTime;
+  if (completeDate) {
+    const time = calculateTimeByType(timeType);
+    convertedTime = convertDateEtToUtc(completeDate, time);
+  }
+  return convertedTime || undefined;
+};
```
braxex, line 82: Modified to use `checkDateCompleteness`.

ntsummers1 (approved):

gmrabian (approved):

PR #3127, opened 2022-07-19
Draft: Spike/number field

ntsummers1:
## Description
Do you like number fields? Do you especially like _MASKED_ number fields? Oh golly-gee-wilikers do I have a PR for you!

### How to test
- ./dev local
- Go to http://localhost:3000/mcpar/test
- Checkout all the new number inputs!

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/fields/NumberField.tsx
+++ b/services/ui-src/src/components/fields/NumberField.tsx
@@ -1,0 +1,91 @@
+import { useFormContext } from "react-hook-form";
+// components
+import { TextField as CmsdsTextField } from "@cmsgov/design-system";
+import { Box } from "@chakra-ui/react";
+// utils
+import { makeMediaQueryClasses } from "utils";
+import { InputChangeEvent, AnyObject } from "types";
+import React, { useState } from "react";
+import { customMask, CustomMasks, isCustomMask } from "utils/other/mask";
+
+export const NumberField = ({
+  name,
+  label,
+  placeholder,
+  mask,
+  sxOverride,
+  nested,
+  ...props
+}: Props) => {
+  const mqClasses = makeMediaQueryClasses();
+
+  const [value, setValue] = useState(
+    isCustomMask(mask)
+      ? customMask(props?.hydrate || "", mask)
+      : props?.hydrate || ""
+  );
+
+  // get the form context
+  const form = useFormContext();
+
+  const onBlurHandler = (evt: React.ChangeEvent<HTMLInputElement>) => {
+    if (isCustomMask(mask)) {
+      const eventValue = customMask(evt.target.value, mask);
+      setValue(eventValue);
+      form.setValue(name, eventValue, { shouldValidate: true });
+    } else {
+      setValue(evt.target.value);
+      form.setValue(name, evt.target.value, { shouldValidate: true });
+    }
+  };
+
+  // update form data
+  const onChangeHandler = async (event: InputChangeEvent) => {
+    const { name, value } = event.target;
+    setValue(value);
+    form.setValue(name, value, { shouldValidate: true });
+  };
+
+  const errorMessage = form?.formState?.errors?.[name]?.message;
+  const nestedChildClasses = nested ? "nested ds-c-choice__checkedChild" : "";
+
+  return (
+    <Box
+      sx={{ ...sx, ...sxOverride }}
+      className={`${mqClasses} ${nestedChildClasses}`}
+    >
+      <CmsdsTextField
+        id={name}
+        name={name}
+        label={label}
+        placeholder={placeholder}
+        onChange={(e) => onChangeHandler(e)}
+        onBlur={onBlurHandler}
+        errorMessage={errorMessage}
+        inputRef={() => form.register(name)}
+        numeric
+        mask={isCustomMask(mask) ? undefined : mask}
+        value={value}
+        {...props}
+      />
+    </Box>
+  );
+};
+
+interface Props {
+  name: string;
+  label: string;
+  mask?: "currency" | "phone" | "ssn" | "zip" | CustomMasks;
+  placeholder?: string;
+  nested?: boolean;
+  sxOverride?: AnyObject;
+  [key: string]: any;
+}
+
+const sx = {
+  "&.nested": {
+    label: {
+      marginTop: 0,
+    },
+  },
+};
--- a/services/ui-src/src/components/fields/NumberField.tsx
+++ b/services/ui-src/src/components/fields/NumberField.tsx
@@ -49,6 +49,12 @@ export const NumberField = ({
   const errorMessage = form?.formState?.errors?.[name]?.message;
   const nestedChildClasses = nested ? "nested ds-c-choice__checkedChild" : "";
 
+  /*
+   * Check if its not a custom mask and if the mask is defined as "currency" | "phone" | "ssn" | "zip".
+   * If its not, we dont want to use the mask prop so return undefined.
+   */
+  const useMask = mask && !isCustomMask(mask) ? mask : undefined;
+
   return (
     <Box
       sx={{ ...sx, ...sxOverride }}
@@ -63,8 +69,7 @@ export const NumberField = ({
         onBlur={onBlurHandler}
         errorMessage={errorMessage}
         inputRef={() => form.register(name)}
-        numeric
-        mask={isCustomMask(mask) ? undefined : mask}
+        mask={useMask}
         value={value}
         {...props}
       />
--- a/services/ui-src/src/components/fields/NumberField.tsx
+++ b/services/ui-src/src/components/fields/NumberField.tsx
@@ -28,20 +28,17 @@ export const NumberField = ({
   // get the form context
   const form = useFormContext();
 
-  const onBlurHandler = (evt: React.ChangeEvent<HTMLInputElement>) => {
-    if (isCustomMask(mask)) {
-      const eventValue = customMask(evt.target.value, mask);
-      setValue(eventValue);
-      form.setValue(name, eventValue, { shouldValidate: true });
-    } else {
-      setValue(evt.target.value);
-      form.setValue(name, evt.target.value, { shouldValidate: true });
-    }
+  const onBlurHandler = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const eventValue = isCustomMask(mask)
+      ? customMask(e.target.value, mask)
+      : e.target.value;
+    setValue(eventValue);
+    form.setValue(name, eventValue, { shouldValidate: true });
   };
 
   // update form data
-  const onChangeHandler = async (event: InputChangeEvent) => {
-    const { name, value } = event.target;
+  const onChangeHandler = async (e: InputChangeEvent) => {
+    const { name, value } = e.target;
     setValue(value);
     form.setValue(name, value, { shouldValidate: true });
   };
--- a/services/ui-src/src/constants.ts
+++ b/services/ui-src/src/constants.ts
@@ -1,2 +1,6 @@
 // BANNERS
 export const bannerId = "admin-banner-id";
+
+// REGEX
+export const numberFormatRegex =
+  /^(?<!\S)(?=.)(0|([1-9](\d*|\d{0,2}(,\d{3})*)))?(\.\d*[0-9])?(?!\S)$/;
--- a/services/ui-src/src/utils/other/mask.tsx
+++ b/services/ui-src/src/utils/other/mask.tsx
@@ -1,0 +1,57 @@
+/*
+ * Custom Masks Type Guard
+ * Add any future custom masks here!
+ */
+const customMasks = ["comma-seperated"] as const;
+export type CustomMasks = typeof customMasks[number];
+export const isCustomMask = (x: any): x is CustomMasks =>
+  customMasks.includes(x);
+
+/**
+ * Checks if the provided string is contains numbers to mask
+ * @param {String} value
+ * @returns {String}
+ */
+function isCustomValueMaskable(value: string): boolean {
+  if (value && typeof value === "string") {
+    const hasDigits = value.match(/\d/);
+    if (hasDigits) {
+      return true;
+    }
+  }
+  return false;
+}
+
+/**
+ * Converts a number-like string to a comma seperated value
+ * @param {String} value
+ * @returns {String}
+ */
+function toCommaSeperated(value: string): string {
+  // Convert String to a float to begin operation
+  const valueAsFloat = parseFloat(value);
+
+  // Slide any extra decimals down to 2
+  const fixedDecimal = valueAsFloat.toFixed(2);
+
+  // Clean up the float value and add in commas to delineate thousands if needed
+  const cleanedValue = Number(fixedDecimal).toLocaleString("en", {
+    minimumFractionDigits: 2,
+  });
+
+  return `${cleanedValue}`;
+}
+
+/**
+ * Returns the value with additional masking characters, or the same value back if invalid numeric string
+ * @param {String} value
+ * @returns {String}
+ */
+export function customMask(value = "", mask: string): string {
+  if (isCustomValueMaskable(value)) {
+    if (mask === "comma-seperated") {
+      value = toCommaSeperated(value);
+    }
+  }
+  return value;
+}
--- a/services/ui-src/src/components/index.ts
+++ b/services/ui-src/src/components/index.ts
@@ -27,6 +27,7 @@ export { ChoiceField } from "./fields/ChoiceField";
 export { ChoiceListField } from "./fields/ChoiceListField";
 export { DateField } from "./fields/DateField";
 export { DropdownField } from "./fields/DropdownField";
+export { NumberField } from "./fields/NumberField";
 export { RadioField } from "./fields/RadioField";
 export { TextField } from "./fields/TextField";
 export { TextAreaField } from "./fields/TextAreaField";
--- a/services/ui-src/src/forms/mcpar/test.json
+++ b/services/ui-src/src/forms/mcpar/test.json
@@ -123,70 +123,80 @@
         }
       },
       {
-        "id": "test9",
-        "type": "number",
-        "props": {
-          "label": "Custom Unhydrated Comma-seperated Textfield",
-          "mask": "comma-seperated",
-          "hydrate": "",
-          "placeholder": ""
-        }
-      },
-      {
         "id": "test3",
         "type": "number",
         "props": {
-          "label": "Custom Comma-seperated Textfield",
-          "mask": "comma-seperated",
-          "hydrate": "123456789",
-          "placeholder": "2500"
+          "label": "Non-Masked Numberfield",
+          "mask": "",
+          "hydrate": "",
+          "placeholder": ""
         }
       },
       {
         "id": "test4",
         "type": "number",
         "props": {
-          "label": "Currency Textfield",
-          "mask": "currency",
-          "hydrate": "9.99",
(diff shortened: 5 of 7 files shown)


services/ui-src/src/utils/other/mask.tsx
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+  return false;
+}
+
+/**
+ * Converts a number-like string to a comma seperated value
+ * @param {String} value
+ * @returns {String}
+ */
```
braxex, line 29:
```suggestion
// converts string of numbers to comma-separated string
```

services/ui-src/src/utils/other/mask.tsx
```diff
@@ -0,0 +1,57 @@
...
+  // Slide any extra decimals down to 2
+  const fixedDecimal = valueAsFloat.toFixed(2);
+
+  // Clean up the float value and add in commas to delineate thousands if needed
+  const cleanedValue = Number(fixedDecimal).toLocaleString("en", {
+    minimumFractionDigits: 2,
+  });
+
+  return `${cleanedValue}`;
+}
```
braxex, line 43: Typo in function name. We might also need to make sure any commas in the value string are removed prior to this, in case users have entered some, because `parseFloat()` will not work correctly if there's anything but numbers and decimals. Also we only need to call `toFixed()` if there is a decimal in the number to start with I think.

```suggestion
const convertToCommaSeparatedString = (value: string): string => {
  const valueAsFloat = parseFloat(value);
  const fixedDecimal = valueAsFloat.toFixed(2);
  // clean up float value; add commas if needed
  const cleanedValue = Number(fixedDecimal).toLocaleString("en", {
    minimumFractionDigits: 2,
  });
  return cleanedValue.toString();
}
```

services/ui-src/src/utils/other/mask.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Returns the value with additional masking characters, or the same value back if invalid numeric string
+ * @param {String} value
+ * @returns {String}
+ */
```
braxex, line 49:
```suggestion
// returns masked value, or if invalid, unchanged value
```

services/ui-src/src/utils/other/mask.tsx
```diff
@@ -0,0 +1,57 @@
...
+export function customMask(value = "", mask: string): string {
+  if (isCustomValueMaskable(value)) {
+    if (mask === "comma-seperated") {
+      value = toCommaSeperated(value);
+    }
+  }
+  return value;
+}
```
braxex, line 57: Updated function name.

```suggestion
export const maskValue = (value = "", mask: string): string =>  {
  if (isNumberStringMaskable(value) && mask === "comma-separated") {
    return convertToCommaSeparatedString(value)
  }
  else return value
}
```

braxex (changes requested):
Made some suggestions. Check for `seperate` 🚫  vs. `separate` ✅ .

services/ui-src/src/constants.ts
```diff
@@ -1,2 +1,6 @@
 // BANNERS
 export const bannerId = "admin-banner-id";
+
+// REGEX
+export const numberFormatRegex =
+  /^(?<!\S)(?=.)(0|([1-9](\d*|\d{0,2}(,\d{3})*)))?(\.\d*[0-9])?(?!\S)$/;
```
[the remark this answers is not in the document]
ntsummers1: image

services/ui-src/src/utils/other/mask.tsx
ntsummers1: True! Cleaned up and tested 💪

gmrabian (approved):

services/ui-src/src/utils/other/mask.tsx
```diff
@@ -0,0 +1,60 @@
...
+export const isCustomMask = (x: any): x is CustomMasks =>
+  customMasks.includes(x);
+
+/**
+ * Checks if the provided string is contains numbers to mask
+ * @param {String} value
+ * @returns {String}
+ */
+export const isNumberStringMaskable = (value: string): boolean =>
+  value?.match(/\d/) ? true : false;
```
braxex, line 16: I'm confused about what exactly makes a number "maskable". Shouldn't this be the number format regex that allows `0123456789.,` and disallows anything else? 

Also, it seems like this method could just be an inline func. 

Also, the ternary operator is unnecessary here because if it matches, it will return an array and if it does not match, it will return `null`.

Also note the renaming in the suggestion below.

```suggestion
export const isValidNumericalString = (value: string): boolean =>
  value?.match(/\d/)
```

services/ui-src/src/utils/other/mask.tsx
```diff
@@ -0,0 +1,60 @@
+/*
+ * Custom Masks Type Guard
+ * Add any future custom masks here!
+ */
+const customMasks = ["comma-separated"] as const;
+export type CustomMasks = typeof customMasks[number];
+export const isCustomMask = (x: any): x is CustomMasks =>
+  customMasks.includes(x);
```
braxex, line 8: I wonder if instead of resorting to types, this could just use the same kind of pattern as we use for the form factory. Something like: 

``` 
const customMaskMap: any = {
    "commaSeparated": CommaSeparatedMaskGoesHere,
    "futureMask1": FutureMask1GoesHere,
    ...and so on
  };

// then access like this
const maskToUse = customMaskMap[<string key from field props>] || fallbackThing;
```

Would also rename to something like `maskToUse` or `selectedMask` or something, 'cause what we're getting back from it is not just "is it a custom mask" but the actual mask that we want to use.

services/ui-src/src/utils/other/mask.tsx
```diff
@@ -0,0 +1,60 @@
...
+ */
+export const isNumberStringMaskable = (value: string): boolean =>
+  value?.match(/\d/) ? true : false;
+
+/**
+ * Converts a number-like string to a comma seperated value
+ * @param {String} value
+ * @returns {String}
+ */
+export function convertToCommaSeparatedString(value: string): string {
```
braxex, line 23: I think we should split this function in two since it is performing two tasks: 
1. Masking for thousands separator (for this app's purposes, should be run on every number input, including currency)
2. Masking for decimal places (should be run only on currency)

braxex: Maybe even a third function that does the cleanup? Not necessarily.

Every number would need to go through: 
- Cleanup
- Comma-separation

Then optionally a number would additionally go through currency to get the decimals.

services/ui-src/src/utils/other/mask.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  // Slide any extra decimals down to 2
+  const fixedDecimal = valueAsFloat.toFixed(2);
+
+  // Clean up the float value and add in commas to delineate thousands if needed
+  const cleanedValue = Number(fixedDecimal).toLocaleString("en", {
+    minimumFractionDigits: 2,
+  });
+
+  return `${cleanedValue}`;
```
braxex, line 47:
```suggestion
  return cleanedValue.toString()
```

services/ui-src/src/utils/other/mask.test.ts
```diff
@@ -0,0 +1,95 @@
+import {
+  convertToCommaSeparatedString,
+  isCustomMask,
+  isNumberStringMaskable,
+  maskValue,
+} from "utils";
+
+const commaSeparatedMaskAcceptableTestCases = [
```
braxex, line 8: All of these test cases are great, but just calling out again that comma-separated should be distinct from currency/decimal addition. There will be lots of cases where we don't want to have decimals added, for instance: "How many people..."

services/ui-src/src/utils/other/mask.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  { test: "1,234", expected: "1,234.00" },
+  { test: "1234", expected: "1,234.00" },
+  { test: "1,234.00", expected: "1,234.00" },
+  { test: "1,234", expected: "1,234.00" },
+  { test: "100,000,000", expected: "100,000,000.00" },
+  { test: "100000000", expected: "100,000,000.00" },
+  {
+    test: "Technically a wrong input that validation would pick up but would actually still work with 1 number",
+    expected: "1.00",
+  },
```
braxex, line 27: Would remove this one and make it so the number masks only operate on strings composed entirely of 0123456789.,. Even though the number mask can handle it we don't really need to test that since our number field validation won't be allowing any through.
```suggestion
```

services/ui-src/src/utils/other/mask.test.ts
```diff
@@ -0,0 +1,95 @@
...
+});
+
+describe("Test isNumberStringMaskable", () => {
+  test("Check if strings with numbers are maskable", () => {
+    for (let testCase of commaSeparatedMaskAcceptableTestCases) {
+      expect(isNumberStringMaskable(testCase.test)).toEqual(true);
+    }
+  });
+
+  test("Check if strings without numbers are not maskable", () => {
```
braxex, line 46: Suggestion for renaming:
```suggestion
  test("Non-numerical strings are rejected", () => {
```
or similar

services/ui-src/src/utils/other/mask.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    const testCases = ["abc", "harryhadalittlelamb"];
+    for (let testCase of testCases) {
+      expect(isNumberStringMaskable(testCase)).toEqual(false);
+    }
+  });
+});
+
+describe("Test convertToCommaSeparatedString", () => {
+  test("Check if number strings can be appropriately comma-seperated", () => {
```
braxex, line 55:
```suggestion
  test("Valid numerical string is correctly masked with thousands separators", () => {
```

services/ui-src/src/utils/other/mask.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    const testCases = [
+      { test: "No numbers here me lord", expected: "NaN" },
+      { test: "", expected: "NaN" },
+    ];
+    for (let testCase of testCases) {
+      expect(convertToCommaSeparatedString(testCase.test)).toEqual(
+        testCase.expected
+      );
+    }
+  });
```
braxex, line 72: Can be combined with the test on line 46.

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+    setValue(eventValue);
+    form.setValue(name, eventValue, { shouldValidate: true });
+  };
+
+  // update form data
+  const onChangeHandler = async (e: InputChangeEvent) => {
+    const { name, value } = e.target;
+    setValue(value);
+    form.setValue(name, value, { shouldValidate: true });
+  };
```
braxex, line 44: Should also be masking onChange. In fact we probably don't even need the `onBlur` handler because it should be auto-masking on every change event along with the user's typing.

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+  label,
+  placeholder,
+  mask,
+  sxOverride,
+  nested,
+  ...props
+}: Props) => {
+  const mqClasses = makeMediaQueryClasses();
+
+  const [value, setValue] = useState(props?.hydrate || "");
```
braxex, line 26: Suggest changing to `displayValue` for clarity / to match pattern in rest of app. 

```suggestion
  const [displayValue, setDisplayValue] = useState(props?.hydrate || "");
```

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+  // get the form context
+  const form = useFormContext();
+
+  const onBlurHandler = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const eventValue = isCustomMask(mask)
+      ? maskValue(e.target.value, mask)
+      : e.target.value;
+    setValue(eventValue);
+    form.setValue(name, eventValue, { shouldValidate: true });
+  };
```
braxex, line 37: Can remove and just always do on `onChange`.
```suggestion
```

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const errorMessage = form?.formState?.errors?.[name]?.message;
+  const nestedChildClasses = nested ? "nested ds-c-choice__checkedChild" : "";
+
+  /*
+   * Check if its not a custom mask and if the mask is defined as "currency"
+   * If its not, we dont want to use the mask prop so return undefined.
+   */
+  const useMask = mask && !isCustomMask(mask) ? mask : undefined;
```
braxex, line 53: Where is it checking if the mask is defined as "currency"?

Also would avoid the name `useMask` because "use" is usually reserved for hooks as a standard naming convention. If you look in `utils/mask` I made some suggestions that could let us use something like `maskToUse` or `selectedMask` or `applicableMask` here like this that would return the same stuff:

```suggestion
  const maskToUse = selectedMask(mask);
```

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  return (
+    <Box
+      sx={{ ...sx, ...sxOverride }}
+      className={`${mqClasses} ${nestedChildClasses}`}
```
braxex, line 58: As noted above:
```suggestion
      className={mqClasses}
```

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+        label={label}
+        placeholder={placeholder}
+        onChange={onChangeHandler}
+        onBlur={onBlurHandler}
+        errorMessage={errorMessage}
+        inputRef={() => form.register(name)}
+        mask={useMask}
+        value={value}
+        sx={{ ...sx, ...sxOverride }}
+        className={`${mqClasses} ${nestedChildClasses}`}
```
braxex, line 72: Unnecessary.
```suggestion
```

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+  sxOverride?: AnyObject;
+  [key: string]: any;
+}
+
+const sx = {
+  "&.nested": {
+    label: {
+      marginTop: 0,
+    },
+  },
```
braxex, line 94: Unnecessary. Should however set a sensible max width for a number field. Probably something in the designs to match.
```suggestion
  ".ds-c-field": {
    maxWidth: "15rem",
  },
```

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+        {...props}
+      />
+    </Box>
+  );
+};
+
+interface Props {
+  name: string;
+  label: string;
+  mask?: "currency" | CustomMasks;
```
braxex, line 82: Can definitely use types here. If we don't, we can also just let them pass through anything they want and if it's not one of the custom masks it will just fail silently as you're testing for in the test files.

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+    >
+      <TextField
+        id={name}
+        name={name}
+        label={label}
+        placeholder={placeholder}
+        onChange={onChangeHandler}
+        onBlur={onBlurHandler}
+        errorMessage={errorMessage}
+        inputRef={() => form.register(name)}
```
braxex, line 68: Handled for us by `TextField` so unnecessary here.
```suggestion
```

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+        mask={useMask}
+        value={value}
```
braxex, line 70:
```suggestion
        value={displayValue}
```

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,95 @@
...
+      sx={{ ...sx, ...sxOverride }}
+      className={`${mqClasses} ${nestedChildClasses}`}
+    >
+      <TextField
+        id={name}
+        name={name}
+        label={label}
+        placeholder={placeholder}
+        onChange={onChangeHandler}
+        onBlur={onBlurHandler}
```
braxex, line 66: Can delete.
```suggestion
```

braxex (changes requested):

services/ui-src/src/utils/other/mask.tsx
braxex: Screen Shot 2022-07-21 at 12 37 32

services/ui-src/src/utils/other/mask.test.ts
ntsummers1: I think on this one we should keep these are testing 2 different functions though that return two different values

services/ui-src/src/utils/other/mask.tsx
ntsummers1: Kept the same for now since we decided on not writing the custom currency mask. Willing to split this up if you think it still needs to be though.

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  /*
+   * Check if its not a custom mask and if the mask is defined as "currency"
+   * or another CMSDS provided mask
+   * If its not, we dont want to use the mask prop so return undefined.
+   */
+  const isCmsdsProvidedMask = mask && !isCustomMask(mask) ? mask : undefined;
+
+  // const maskToUse = customMaskMap[mask] || mask;
```
braxex, line 47: Nit: Since this is actually returning the mask now, we could rename it. 
```suggestion
  const cmsdsPassedMask = mask && !isCustomMask(mask) ? mask : undefined;
```

services/ui-src/src/components/fields/NumberField.tsx
```diff
@@ -0,0 +1,80 @@
...
+      />
+    </Box>
+  );
+};
+
+interface Props {
+  name: string;
+  label: string;
+  mask?: string;
+  placeholder?: string;
```
braxex, line 70: Nit: Swap these two to match other components.
```suggestion
  placeholder?: string;
  mask?: string;
```

services/ui-src/src/utils/other/mask.tsx
```diff
@@ -0,0 +1,61 @@
+/*
+ * Custom Masks Type Guard
+ * Add any future custom masks here!
+ */
+export type CustomMasks = typeof customMaskMap;
+export const isCustomMask = (x: any): x is CustomMasks =>
```
braxex, line 6: Nit: Maybe `isValidCustomMask`

braxex (changes requested):
Hey this looks great! Left a few nit suggestions but otherwise looking good. 

There is one bug I found where sometimes when you enter a `,` or `.` in the currency or comma-separated masked fields the field will show an error that it needs to be a number. Probably just need to tweak the number regex a bit to allow users to enter their own decimals and commas without getting a validation warning. The error disappears when you blur because the mask is run and the value is reset, but for that moment before they blur we don't want the field to move and have a red error message. 

Getting super close and thanks again!

braxex (approved):
Looks great! Thank you!

gmrabian (approved):

PR #3136, opened 2022-07-20
Add bss entities page; configure dynamic field array functionality

braxex:
## Description
[MDCT-411](https://qmacbis.atlassian.net/browse/MDCT-411) Add BSS entities page. This adds a crucial component which is the ability for users to add and remove fields dynamically. The values from this page will later be used to have the user fill out sections of the form multiple times per answer

### How to test
- Run locally
- Navigate to `/mcpar/program-information/add-bss-entities`
- Test out page
  - Fill values of current box
  - Click to add a row, should only let you add one at a time
  - Click X to remove row - should remove the current row, not the last item added
  - Add multiple, then delete values from multiple. Click X. Should remove all blank fields
  - Should throw an error if you click _Save & continue_ while some fields are empty

## Code author checklist


gmrabian:
Seeing an odd error in the console. When I click on `Add BSS Entities` in the sidebar it throws this error, but the form still works. If I refresh the page immediately the error does not throw again. So seems to be a problem with navigating to this page or something?
```
Warning: Cannot update a component (`Form`) while rendering a different component (`DynamicField`). To locate the bad setState() call inside `DynamicField`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
```

gmrabian:
Tests are doing something really weird. DateField is failing in numerous test files where it's not even used

gmrabian:
> Tests are doing something really weird. DateField is failing in numerous test files where it's not even used

The issue seemed to stem from having the field mapping externalized in forms.ts. This was from a prior iteration on this branch so I moved it back and tests pass now

gmrabian:
> Seeing an odd error in the console. When I click on `Add BSS Entities` in the sidebar it throws this error, but the form still works. If I refresh the page immediately the error does not throw again. So seems to be a problem with navigating to this page or something?
> 
> ```
> Warning: Cannot update a component (`Form`) while rendering a different component (`DynamicField`). To locate the bad setState() call inside `DynamicField`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
> ```

Fixed this by moving initialization append() to useEffect

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -0,0 +1,88 @@
...
+// components
+import { Box, Button, Flex, Image } from "@chakra-ui/react";
+// utils
+import { makeMediaQueryClasses } from "utils";
+// assets
+import cancelIcon from "assets/icons/icon_cancel_x_circle.png";
+import { TextField } from "./TextField";
+
+export const DynamicField = ({ name, label }: Props) => {
+  const mqClasses = makeMediaQueryClasses();
```
braxex, line 12: Don't really need this here -- actually should probably remove it from any components where we aren't directly using .desktop or .tablet or .mobile.
```suggestion
```

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -0,0 +1,88 @@
...
+    name: name,
+    shouldUnregister: true,
+  });
+
+  useEffect(() => {
+    if (fields.length === 0) append("");
+  });
+
+  return (
+    <Box sx={sx} className={mqClasses}>
```
braxex, line 27: to match
```suggestion
```

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -0,0 +1,88 @@
...
+      <Box>
```
braxex, line 28: No need for this `Box` here.

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -0,0 +1,88 @@
...
+              />
+              {index != 0 && (
+                <Button onClick={() => remove(index)} variant="unstyled">
+                  <Image
+                    sx={sx.removeButton}
+                    src={cancelIcon}
+                    alt="Remove item"
+                  />
+                </Button>
+              )}
```
braxex, line 49: There was a bit of styling weirdness with this button where the button had a big wrapper around the icon that made the cursor into a pointer too soon. I started writing out a big thing explaining how to fix it but it was a pain so I just made a PR. I'll put that up soon. 

The first part is not making this a Chakra button and just using a regular HTML one.

Also I'm torn on `index != 0 && ...` -- technically it could just be `index && ...`. But I think it's better this way. Maybe add a comment that explains it's because we never want to have none and that's why it's not on the first one?

```suggestion
              {index != 0 && (
                <button onClick={() => remove(index)}>
                  <Image
                    sx={sx.removeButton}
                    src={cancelIcon}
                    alt="Remove item"
                  />
                </button>
              )}
```

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -0,0 +1,88 @@
...
+  name: string;
+  label: string;
+}
+
+const sx = {
+  removeButton: {
+    boxSize: "1.25rem",
+    marginBottom: "1rem",
+    marginLeft: "0.625rem",
+  },
```
braxex, line 76: There'll be some updated styling for this in the PR.

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -0,0 +1,88 @@
...
+  appendButton: {
+    minWidth: "202px",
+    minHeight: "42px",
+    fontWeight: "bold",
+    fontSize: "1rem",
+    color: "palette.main",
+    bg: "palette.white",
+    border: "1px solid var(--chakra-colors-palette-main)",
+    borderRadius: "3px",
+    marginTop: "2rem",
```
braxex, line 86: Reordering per best practice.
```suggestion
    minWidth: "202px",
    minHeight: "42px",
    marginTop: "2rem",
    border: "1px solid var(--chakra-colors-palette-main)",
    borderRadius: "3px",
    bg: "palette.white",
    fontSize: "1rem",
    fontWeight: "bold",
    color: "palette.main",
```

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  useEffect(() => {
+    if (fields.length === 0) append("");
+  });
+
+  return (
+    <Box sx={sx} className={mqClasses}>
+      <Box>
+        {fields.map((field: any, index: any) => {
```
braxex, line 29: Should tune up the types here.

services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -25,7 +26,9 @@ export const TextField = ({
     form.setValue(name, value, { shouldValidate: true });
   };
 
-  const errorMessage = form?.formState?.errors?.[name]?.message;
+  const errorMessage =
+    form?.formState?.errors?.[name]?.message ||
+    form?.formState?.errors?.[dynamic?.parentName]?.[dynamic?.index]?.message;
```
braxex, line 31: I know I was involved in the creation of this monstrosity, but we should clean it up lol.

With a little work on the types (will have to define dynamic in component Props and guard against undefined) something like this should do the trick.

```suggestion
  const { parentName, index, inputRef } = dynamic;
  const formErrorState = form?.formState?.errors;
  const errorMessage =
    formErrorState?.[name]?.message ||
    formErrorState?.[parentName]?.[index]?.message;
```

services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -54,6 +57,7 @@ interface Props {
   placeholder?: string;
   nested?: boolean;
   sxOverride?: AnyObject;
+  dynamic?: AnyObject;
```
braxex, line 60: Should give a better type.

services/ui-src/src/components/forms/Form.tsx
```diff
@@ -1,5 +1,5 @@
 import { ReactNode } from "react";
-import { useForm, FormProvider } from "react-hook-form";
+import { FormProvider, useForm } from "react-hook-form";
```
braxex, line 2: 🤷

services/ui-src/src/components/fields/DynamicField.test.tsx
```diff
@@ -0,0 +1,50 @@
+import { render, screen } from "@testing-library/react";
+import { axe } from "jest-axe";
+//components
+import { DynamicField } from "components";
+
+const appendButtonText = "Add a row";
```
braxex, line 6: Nit: If you're only using this once, I'd argue it doesn't need to be defined globally -- just use the string in the test.

services/ui-src/src/components/fields/DynamicField.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    register: () => {},
+  }),
+  useFieldArray: jest.fn().mockReturnValue({
+    fields: [
+      {
+        id: "uniqueId",
+        0: {
+          0: "text",
+        },
+      },
```
braxex, line 19: should just be an array of strings: ["value1", "value2"]

services/ui-src/src/components/fields/DynamicField.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    ],
+    append: () => {},
+    remove: () => {},
```
braxex, line 22: will want to make these `jest.fn()` so we can spy on them later in some new tests.

services/ui-src/src/components/fields/DynamicField.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    render(dynamicFieldComponent);
+    const inputBoxLabel = screen.getByText("test-label");
+    expect(inputBoxLabel).toBeVisible();
+  });
+
+  test("DynamicField append button is visible", () => {
+    render(dynamicFieldComponent);
+    const appendButton = screen.getByText(appendButtonText);
+    expect(appendButton).toBeVisible();
+  });
```
braxex, line 41: Need to add a couple more tests here for: 
1. Append button functionality
2. Remove button functionality
3. Remove button only shows on elements after the first.

services/ui-src/src/components/fields/DynamicField.tsx
gmrabian: hmm I tried just `index &&` and it printed the index 0 next to the box. Could be something I missed. I just left it as is. Will add a comment. updated the button.

services/ui-src/src/components/fields/DynamicField.test.tsx
gmrabian: thought it might be easier to update later if the text in dynamic form changes. Can do locally though

services/ui-src/src/components/fields/DynamicField.test.tsx
gmrabian: When I did a printout of fields it looked more like this ^ not sure

braxex:
Once this has tests it will be g2g.

ntsummers1 (approved):

ntsummers1:
This PR be clean bruh. Nice work to both of you!

PR #3176, opened 2022-07-25
modify sidebar placement, styling, functionality

braxex:
## Description

After discussion with design, they would like the sidebar to behave differently. The ask is to make the sidebar behave as a fixed overlay on mobile and tablet, but on desktop have it embedded as part of the actual report page so that expanding it will push the footer down. 

The big change is essentially moving the sidebar inside the report pages so it can push the footer down. This changes the relationship of the sidebar to the page content, so they're now both wrapped in a flex as siblings. Lots of additional complexity from the previous implementation is also removed because it's unnecessary now. Cool. 😎 

### How to test

1. Navigate to `/mcpar/program-information/point-of-contact`
2. On desktop, check that page expands and pushes footer down to the greater of either the content height or the sidebar height. 
3. On tablet and mobile, check that sidebar expands to the bottom of the page (over the footer) when open, and that user can scroll to all elements of the sidebar regardless of how deep the sidebar is open.

### Changed Dependencies

n/a

## Code author checklist


braxex:
@ntsummers1 @gmrabian 

Feel free to make any needed changes and merge while I'm out. 👍

gmrabian (approved):
Works nicely

gmrabian:
Wonder if we still need the sidebarIsOpen state at the app level or if we can move that down too 🤔 but doesn't matter too much at the moment, still works

ntsummers1 (approved):

PR #3259, opened 2022-08-01
MDCT-480: Create Intro Page

ntsummers1:
## Description

Fixes [MDCT-480](https://qmacbis.atlassian.net/browse/MDCT-480)

https://user-images.githubusercontent.com/19439679/182221353-59f21a70-fdfc-44c1-8491-d51be20b1a8b.mov

### How to test
./dev local
Click the link to the MCPAR form
Checkout the new page!

## Code author checklist


services/ui-src/src/types/index.ts
```diff
@@ -124,6 +124,37 @@ export interface TimeShape {
...
+export interface iconlessWidget {
+  type: string;
+  content: {
+    leftBarColor: string;
+    title: string;
+    descriptionList: string[];
+  };
+}
+
+export interface iconWidget {
```
gmrabian, line 137: Maybe iconWidget could extend iconlessWidget, since they share the same properties?

gmrabian (approved):
Looking good! One suggestion on the types

Also the design images don't match the actual form (when sidenav dropdowns are closed the arrow is pointed down). If possible it'd be good that those matched up

ntsummers1:
> Looking good! One suggestion on the types
> 
> Also the design images don't match the actual form (when sidenav dropdowns are closed the arrow is pointed down). If possible it'd be good that those matched up

Gooood catch. I'll hit up design about that

services/ui-src/src/types/index.ts
ntsummers1: If I keep the same format extending looks like this:

```
export interface iconWidget extends iconlessWidget {
  content: {
    leftBarColor: string;
    icon: string;
    iconDescription: string;
    title: string;
    descriptionList: string[];
    additionalInfo?: string;
  };
}
```

just by way of the content object needing to be defined.. Could change up these types though so the content part is actually what the interface is and was something I was debating for quite awhile before settling on combining them. Thoughts on that?

ntsummers1:
Updated Images!

services/ui-src/src/types/index.ts
gmrabian: Hmm I see what you mean. It looks incomplete this way. Ok I'm cool with keeping them separate! Maybe Brax will have an opinion here.

gmrabian (approved):
Good stuff! thanks for fixing the images

services/ui-src/src/components/layout/Section.tsx
```diff
@@ -0,0 +1,71 @@
+import React from "react";
+// components
+import { Box, Flex, Heading, Text } from "@chakra-ui/react";
+import { WidgetContainer } from "components/widgets/WidgetContainer";
```
braxex, line 4: Suggestion: Could just reference this from components directly because of the way we're exporting it.
```suggestion
import { WidgetContainer } from "components";
```

services/ui-src/src/components/layout/Section.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export const Section = ({ index, content, ...props }: Props) => {
```
braxex, line 6: This is probably good for now. I would maybe change the name to `IntroSection` or something so it's clear that this is a very specific type of section. Alternatively, I might make the `Section` component be just the outside styling, something like: 
```js
<Flex sx={sx.sectionContainer} {...props}>
  {children}
</Flex>
```
and then make a separate component `IntroSection` which would be the child stuff that goes in the section. This is the kind of pattern we use for `Card`. Then call them together like:
```
<Section>
  <IntroSection />
</Section>
```
Idk, just a thought.

services/ui-src/src/components/layout/Section.tsx
```diff
@@ -0,0 +1,71 @@
...
+    </Flex>
+  );
+};
+
+interface Props {
+  index: number;
+  content: {
+    header: string;
+    body: string;
+    widgets: any[];
```
braxex, line 32: Can probably tighten up the `widgets` type.

services/ui-src/src/components/layout/Section.tsx
```diff
@@ -0,0 +1,71 @@
...
+  },
+  numberContainer: {
+    display: "flex",
+    width: "1.5rem",
+    height: "1.5rem",
+    marginTop: ".2rem",
+    marginRight: "1rem",
+    background: "#333",
+    borderRadius: "100%",
+    color: "white",
```
braxex, line 51: Suggest switching to theme colors (the gray seems close enough to me).
```suggestion
    background: "palette.gray_dark",
    borderRadius: "100%",
    color: "palette.white",
```

services/ui-src/src/components/layout/Section.tsx
```diff
@@ -0,0 +1,71 @@
...
+    fontWeight: "bold",
+  },
+  header: {
+    marginBottom: "1rem",
+    color: "palette.gray",
+    fontSize: "2xl",
+  },
+  widgetsContainer: {
+    marginTop: "1rem",
+    gridGap: "2rem",
```
braxex, line 65: Suggestion: We haven't really used grid stuff anywhere else, so this might be something we want to swap out for margins instead. 🤷

services/ui-src/src/types/index.ts
```diff
@@ -131,6 +131,37 @@ export interface TimeShape {
   second: number;
 }
 
+// Widgets
+export interface iconlessWidget {
```
braxex, line 135: Type interfaces are usually capitalized to keep them separate from regular variables.

services/ui-src/src/types/index.ts
braxex: Could make a `WidgetContent` interface that includes most of that. Then extend it to an `IconWidgetContent` and `IconlessWidgetContent` (as needed) that could be extended in those other ones. Seems to me like you're only really ever accessing the content anyway, and even then it could probably just be declared as the `Props` in the actual widget component files. Happy to pair on this.

services/ui-src/src/routes/Intro/Intro.tsx
```diff
@@ -1,29 +1,83 @@
 // components
-import { Box, Heading, Text } from "@chakra-ui/react";
-import { ReportPage } from "components";
+import { Box, Button, Image, Heading } from "@chakra-ui/react";
+import { BasicPage, Section } from "components";
+import { useNavigate } from "react-router-dom";
```
braxex, line 4: Nit: Not a component, so should be on line 1 above components to match other files.

services/ui-src/src/verbiage/pages/mcpar/mcpar-intro.ts
```diff
@@ -1,6 +1,73 @@
...
+            type: "iconWidget",
+            content: {
+              leftBarColor: "#2E8540",
+              icon: greenSpreadsheetIcon,
+              iconDescription: "Excel Workbook Icon",
+              title: "Find in the Excel Workbook",
+              descriptionList: ["A_Program_Info"],
+              additionalInfo:
+                "Use these guides to understand which sections match specific tabs in the Excel workbook.",
+            },
```
braxex, line 50: I'm conflicted about putting colors here since we're trying to stick to theming so much (I know this value probably would need to be added to the theme anyway though). 

I wonder if it would make sense to take the common things from here that all widgets of a certain type could share, for instance, we will be using this spreadsheet widget in like a million places with just slightly different wording. So we could have the type be 'spreadsheetWidget' or something and when it's that type of widget, it gets the spreadsheet icon and green color bar. Just trying to think for the future where we won't be able to pass in icon or what to manage colors in the form.json where the spreadsheet icons will be used a lot. Especially since having all those details here in json puts the burden of getting it right on the json writer in 3 areas (type, color, icon) instead of just one: which type of widget do you want.

braxex (changes requested):
I like what you've done here, but I'm also concerned about implementing the concept of widgets when really the vast majority of these (other than the spreadsheet widget) are only going to be used in this one location. It adds a lot of components and additional complexity where we really just need a static page with some images and text.

An alternative pattern would be to just make the sections (see comment about making a `Section` component that then is filled with children) and then fill each of them with the JSX they need to display the text/images instead of running them through `WidgetContainer` and each individual Widget component. Doing it this way would decrease complexity in components and the json you're using to manage the page, and makes it easier to maintain in the future. Something to think about. I'm happy to talk it through with you whenever. 👍

services/ui-src/src/components/widgets/WidgetContainer.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <Flex sx={sx.widgetsContainer} className={mqClasses} {...props}>
+      {widgets.map((widget, index) => widgetSelector(widget, index))}
+    </Flex>
+  );
+};
+
+interface Props {
+  widgets: Array<iconWidget | iconlessWidget | imageWidget>;
```
BearHanded, line 49: Do these have shared parentage?

services/ui-src/src/components/widgets/WidgetContainer.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Flex } from "@chakra-ui/react";
+import { IconlessWidget, IconWidget, ImageWidget } from "components";
+// utils
+import { iconlessWidget, iconWidget, imageWidget } from "types";
+import { makeMediaQueryClasses } from "utils";
+
+export const widgetSelector = (widget: any, index: number) => {
+  const key = `widget-${index}`;
+  switch (widget.type) {
+    case "iconWidget":
```
BearHanded, line 12: Could these cases be mapped as an enum somewhere rather than just matching by string? It could give you somewhere better to look for all the supported widget types, as opposed to digging through this switch block as it expands.

services/ui-src/src/types/index.ts
```diff
@@ -131,6 +131,37 @@ export interface TimeShape {
...
+    leftBarColor: string;
+    icon: string;
+    iconDescription: string;
+    title: string;
+    descriptionList: string[];
+    additionalInfo?: string;
+  };
+}
+
+export interface imageWidget {
```
BearHanded, line 156: capitalize interface

services/ui-src/src/verbiage/pages/mcpar/mcpar-intro.ts
BearHanded: Agreeing with Brax here in regards to the theming. I think it makes sense to just have the form include something like `leftBarColor: 'green'` and have those mapped against colors from known themes internally. It's a lot less of a mental burden when spinning up a new section than having to go back and find the code for green you used elsewhere.

BearHanded (dismissed):
I think this needs some cleanup, but is generally on the right track.

ntsummers1:
Updated PR to include the Slimfast Treatment (Removed about 10 files and simplified the code)

services/ui-src/src/routes/Intro/Intro.tsx
```diff
@@ -1,29 +1,138 @@
...
-    </ReportPage>
+      <div>
+        <IntroSection index={1} content={section1} key={`section-1`}>
+          <Flex sx={sx.widgetsContainer} className={mqClasses}>
+            <Box sx={sx.imagelessWidgetContainer}>
+              <Text sx={sx.imagelessWidgetTitle}>{section1.widget?.title}</Text>
+              <Box>
+                {section1.widget?.descriptionList.map((description, index) => (
+                  <Text key={index}>{description}</Text>
+                ))}
```
braxex, line 33: Typically want to avoid using `index` as key, and I think react will usually yell at you for it. I'd suggest just using the `description` as the key.

services/ui-src/src/routes/Intro/Intro.tsx
```diff
@@ -1,29 +1,138 @@
...
+    <BasicPage sx={sx.layout} data-testid="intro-view">
       <Box sx={sx.leadTextBox}>
         <Heading as="h1" sx={sx.headerText}>
           {intro.header}
         </Heading>
-        <Text>{intro.body}</Text>
       </Box>
-    </ReportPage>
+      <div>
+        <IntroSection index={1} content={section1} key={`section-1`}>
```
braxex, line 26: Since you're already passing in all the section content, I'd suggest including the `index` here (name should be changed as noted above) in the content that's stored in the json as like `sectionNumber`. Also I don't think the `key` prop is needed anymore.
```suggestion
        <IntroSection content={section1}>
```

services/ui-src/src/routes/Intro/Intro.tsx
```diff
@@ -1,29 +1,138 @@
...
+  },
+  imagelessWidgetContainer: {
+    marginTop: "1rem",
+    paddingLeft: "1rem",
+    borderLeft: ".3rem solid",
+    borderColor: "palette.main",
+  },
+  imagelessWidgetTitle: {
+    fontWeight: "bold",
+  },
```
braxex, line 108: Since these aren't really widgets anymore, we should change these style names to something else. Maybe: `sectionContentContainer`, `sectionContentContainerTitle`, etc.

services/ui-src/src/routes/Intro/Intro.tsx
```diff
@@ -1,29 +1,138 @@
...
+  image: {
+    maxWidth: "20rem",
+  },
+  additionalInfo: {
+    marginTop: "1rem",
+    fontSize: "14",
```
braxex, line 114: Use font size from theme where possible.

services/ui-src/src/verbiage/pages/mcpar/mcpar-intro.ts
```diff
@@ -1,6 +1,51 @@
+import NavigationSectionsImage from "../../../assets/images/NavigationSections_2x.png";
+import NavigationSectionsSubmissionImage from "../../../assets/images/NavigationSectionsSubmission_2x.png";
+
```
braxex, line 3: Eventually all of these will be stored as JSON files, so rather than storing the images here, let's move them to be imported inside the intro page and called directly in the image sections 2 and 3. That way we're not referencing dynamic things in what should just be a static file.

services/ui-src/src/verbiage/pages/mcpar/mcpar-intro.ts
```diff
@@ -1,6 +1,51 @@
...
+        header: "Start by adding all Medicaid managed care programs",
+        body: "You will need to submit one MCPAR for every managed care program in your state. For this MCPAR report, a managed care program is defined by distinct benefits and eligibility criteria articulated in a contract between states and the state’s managed care plans.",
+        widget: {
+          title: "Examples of managed care programs:",
+          descriptionList: [
+            "Health and Recovery Plans (Comprehensive MCO)",
+            "Dental Managed Care",
+          ],
+        },
+      },
```
braxex, line 20: As noted above, add a key/value pair of `sectionNumber: 1,` instead of passing the number in directly in the code. Since the numbers are really text instead of "number" numbers it makes more sense to keep them here in this static text file.

braxex (changes requested):
Looking really good! Just a couple minor tweaks and then this will be good to go!

services/ui-src/src/routes/Intro/Intro.tsx
ntsummers1: using description would be a really long key.. maybe `section-1-desc-${index}`

services/ui-src/src/routes/Intro/Intro.tsx
ntsummers1: Can rename widgetsContainer to sectionContent but i'm not sure about the imagelessWidgetContainer->sectionContentContainer. If we're keeping the Spreadsheet Widget name and this small part looks exactly like it without the image and a change in bold title/description then I'm not sure what else to call it

services/ui-src/src/routes/Intro/Intro.tsx
braxex: Yeah I mean even `index` is fine as a key in this instance because the children are not dynamically changing their order or having items added or deleted, which is really why you don't use index, because on re-renders it can cause bugs. In this case we can just use `index` though. 

In your suggestion the issue above could still happen because the only unique part is the index. As for description being real long, that's true, but also 🤷 it's in the DOM and computers don't care.

So for now, just key={index} is fine. 👍

services/ui-src/src/components/layout/IntroSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+    fontSize: "2xl",
+  },
+  widgetsContainer: {
+    marginTop: "1rem",
+    gridGap: "2rem",
+    flexDirection: "column",
+    "&.desktop": {
+      flexDirection: "row",
+    },
+  },
```
braxex, line 67: Looks like this isn't used anymore.
```suggestion
```

services/ui-src/src/routes/Intro/Intro.tsx
braxex: Ah gotcha. How about `spreadsheetWidgetContainer` and `spreadsheetWidgetTitle`?

braxex (approved):

services/ui-src/src/routes/Intro/Intro.tsx
ntsummers1: I think I see the confusion here. So theres 2 "widgets", for lack of a better term, on the page. We have the top one which was using the imagelessWidgetContainer and imagelessWidgetTitle class: 

Screen Shot 2022-08-03 at 6 08 54 PM

and we have the bottom one that uses its own reusable component with its own classes:
Screen Shot 2022-08-03 at 6 09 53 PM

I wouldn't want to rename the imagelessWidgetContainer or imagelessWidgetTitle css class to spreadsheetWidgetContainer or sectionContentContainer since that seems either not what it is or too vague.

I've settled, for now, on reducing the class names to widgetContainer and widgetTitle to reflect this but happy to change it to something that could be more fitting.

PR #3388, opened 2022-08-09
Create ChakraUI Modal Component

ntsummers1:
## Description

Modal using Chakra UI but stylized to look like a CMSDS one

Screen Shot 2022-08-10 at 4 48 53 PM

### How to test
./dev local
Go to /mcpar/review-and-submit
Click the button!

## Code author checklist


services/ui-src/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,172 @@
...
+  Text,
+} from "@chakra-ui/react";
+import { CloseIcon } from "@cmsgov/design-system";
+import { makeMediaQueryClasses } from "utils";
+
+export const Modal = ({ actionFunction, content, modalState }: Props) => {
+  const mqClasses = makeMediaQueryClasses();
+  return (
+    <ChakraModal
+      data-testid="modal"
```
gmrabian, line 23: Does it need a test id? Looks like we don't use it in the test

services/ui-src/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,172 @@
...
+    >
+      <ModalOverlay />
+      <ModalContent sx={sx.modalContent}>
+        <ModalHeader sx={sx.modalHeader}>
+          <Heading as="h1" sx={sx.modalHeaderText}>
+            {content.heading}
+          </Heading>
+        </ModalHeader>
+        <Box sx={sx.modalCloseContainer}>
+          <Button sx={sx.modalClose} leftIcon={<CloseIcon />} variant="link">
```
gmrabian, line 35: Close button doesn't do anything right now. Do we need it if we also have cancel? If so, we should add the close function here too

ntsummers1: Forgot my final merge to push before I went and made some dinner. Good catch 👍

services/ui-src/src/components/modal/Modal.tsx
ntsummers1: Removed!

gmrabian (approved):
Lovely

services/ui-src/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+  ModalOverlay,
+  Text,
+} from "@chakra-ui/react";
+import { CloseIcon } from "@cmsgov/design-system";
+import { makeMediaQueryClasses } from "utils";
+
+export const Modal = ({ actionFunction, content, modalState }: Props) => {
+  const mqClasses = makeMediaQueryClasses();
+  return (
+    <ChakraModal isOpen={modalState.isOpen} onClose={modalState.onClose}>
```
braxex, line 22: I would maybe pass the `isOpen` and `onClose` props directly rather than wrapped in `modalState` but this is fine too.

services/ui-src/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+      <ModalOverlay />
+      <ModalContent sx={sx.modalContent}>
+        <ModalHeader sx={sx.modalHeader}>
+          <Heading as="h1" sx={sx.modalHeaderText}>
+            {content.heading}
+          </Heading>
+        </ModalHeader>
+        <Box sx={sx.modalCloseContainer}>
```
braxex, line 30: Since this is styled with `display: flex` you could just use the `<Flex />` component here instead of `<Box />`.

services/ui-src/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+    right: "2rem",
+  },
+  modalClose: {
+    color: "palette.primary",
+    fontWeight: "bold",
+    textDecoration: "underline",
+    _hover: {
+      textDecoration: "none",
+      color: "palette.primary_darker",
+    },
```
braxex, line 113: Shouldn't need any of this since you're using the new button `variant="link"` which has been updated and will style this right.

services/ui-src/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+    padding: "0",
+    paddingTop: "2rem",
+  },
+  action: {
+    justifyContent: "start",
+    marginTop: "1rem",
+    borderRadius: "0.25rem",
+    background: "palette.primary",
+    fontWeight: "bold",
+    color: "palette.white",
```
braxex, line 138: Shouldn't need now that we have default Button styling as the 'primary' variant.

services/ui-src/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+    marginRight: "2rem",
+    span: {
+      marginLeft: "0.5rem",
+      marginRight: "-0.25rem",
+    },
+    _hover: {
+      background: "palette.primary_darker",
+    },
```
braxex, line 146: Same note as above.

services/ui-src/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+    "&.mobile": {
+      fontSize: "sm",
+    },
+  },
+  close: {
+    justifyContent: "start",
+    paddingY: ".5rem",
+    paddingLeft: "1rem",
+    paddingRight: "1rem",
```
braxex, line 155: Can set all at once:
```suggestion
    padding: ".5rem 1rem",
```

services/ui-src/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+    marginTop: "1rem",
+    borderRadius: "0.25rem",
+    fontWeight: "bold",
+    color: "palette.primary",
+    textDecoration: "underline",
```
braxex, line 160: Shouldn't need thanks to new `variant="link"` styling.

services/ui-src/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+    span: {
+      marginLeft: "0rem",
+      marginRight: "0.5rem",
+    },
+    _hover: {
+      textDecoration: "none",
+    },
```
braxex, line 167: Same note as above.

braxex (approved):
Looks really good! Just a couple things to tune up and you're good to go!

PR #3398, opened 2022-08-10
Number field is now req by default with optional call

ntsummers1:
## Description
Realized in the last PR that it took away the numberfield being required by default. Added that back in with a separate call if you don't want it to be required.

### How to test
./dev local
Check out any numberfields (Test page has a few)

## Code author checklist

### DIFF
--- a/services/ui-src/src/forms/mcpar/apoc/apoc.schema.ts
+++ b/services/ui-src/src/forms/mcpar/apoc/apoc.schema.ts
@@ -1,11 +1,11 @@
-import { object, string } from "yup";
-import { email, text } from "utils/forms/schemas";
+import { object } from "yup";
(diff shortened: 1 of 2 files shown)


gmrabian (approved):

braxex (commented):
Nick! This PR got me thinking and then I couldn't stop myself from trying it out. 

Instead of basing required fields on optional fields, I've added a `.notRequired()` call to the end of optional fields and created them for each type. That way we have them when we need them. 

@ntsummers1 @gmrabian thoughts?

gmrabian (approved):
cool either way

ntsummers1:
That is so weird that it allows for .required.notRequired().. Works though!

PR #3409, opened 2022-08-10
fix text field undefined label

braxex:
Hotfix for text field issue.


gmrabian:
Isn't label required? When does this happen?

juliuscaesar (approved):

karla-vm (approved):
❤️ +1|-1

braxex:
> Isn't label required? When does this happen?

@gmrabian It's changed to optional because some of the nested text fields for like "Other, specify" options do not have a label or a hint. But an error was being thrown from within the CMSDS component at not getting at least something for a label, so here we are.

PR #3414, opened 2022-08-11
MDCT-469: Review and Submit

ntsummers1:
## Description
Bringing a success page to a form near you

https://user-images.githubusercontent.com/19439679/184305941-d2173163-fa9d-4660-80b1-1b7ab5d0ee35.mov

## Code author checklist


services/ui-src/src/components/reports/ReportProvider.tsx
```diff
@@ -16,7 +17,7 @@ import {
 import { reportErrors } from "verbiage/errors";
 
 export const ReportContext = createContext<ReportContextShape>({
-  reportStatus: "",
+  reportStatus: {} as AnyObject,
```
braxex, line 20: As discussed, this may be changing soon but it's fine for right now. We'll have to look into exactly what we want the data structure of this to be.

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+  Heading,
+  Text,
+  useDisclosure,
+} from "@chakra-ui/react";
+import { Icon, Modal, ReportContext, ReportPage, Sidebar } from "components";
+// types
+import { ReportStatus, UserRoles } from "types";
+// utils
+import { useUser } from "utils";
+// form data
```
braxex, line 17:
```suggestion
// verbiage
```

services/ui-src/src/types/index.ts
```diff
@@ -41,6 +41,12 @@ export interface ReportPath {
   children?: ReportPath[];
 }
 
+export enum ReportStatus {
+  INITIAL = "",
+  IN_PROGRESS = "In Progress",
+  COMPLETED = "Completed",
```
braxex, line 47: We should double-check with David what these statuses should be. For now, let's go with the following and make any necessary changes to the other files to align.
```suggestion
  CREATED = "Created",
  IN_PROGRESS = "In Progress",
  SUBMITTED = "Submitted",
```

services/ui-src/src/verbiage/pages/mcpar/mcpar-successful-submit.ts
```diff
@@ -0,0 +1,9 @@
+export default {
+  intro: {
+    header: "Successfully Submitted",
+    infoHeader: "Thank you",
+    additionalInfoHeader: "What happens now?",
+    additionalInfo:
+      "No further action is needed at this point. CMS will reach out if in the case they have any questions.",
+  },
+};
```
braxex, line 9: Let's combine this into the other `review-and-submit` verbiage object so we just have the one object for the page.

services/ui-src/src/verbiage/pages/mcpar/mcpar-review-and-submit.ts
```diff
@@ -1,6 +1,16 @@
...
-    body: "Placeholder page",
+    header: "Review & Submit",
+    infoHeader: "Ready to Submit?",
+    info: "Double check that everything in your MCPAR Report is accurate. You will be able to make edits after submitting, and resubmit. Once you’ve reviewed your report, certify that it’s in compliance with 42 CFR § 438.66(e).",
+  },
+  modal: {
+    action: "Submit MCPAR",
+    body: "You will be able to make edits to this MCPAR after submitting, and resubmit.",
+    close: "Cancel",
+    heading: "Are you sure you want to submit MCPAR?",
```
braxex, line 11: Nit: Order these so that they flow like the modal and maybe update the names for clarity, like:
```suggestion
    heading: "Are you sure you want to submit MCPAR?",
    body: "You will be able to make edits to this MCPAR after submitting, and resubmit.",
    actionButtonText: "Submit MCPAR",
    closeButtonText: "Cancel",
    
```

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+  const { isOpen, onOpen, onClose } = useDisclosure();
+
+  // get user's state
+  const { user } = useUser();
+  const { state, userRole } = user ?? {};
+  const reportYear = "2022";
+  const reportKey = `${state}${reportYear}`;
+
+  // TODO: get real program name per report
+  const programName = "tempName";
```
braxex, line 35: This will be changing and we'll get it from the ReportContext but this is fine for now.

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+  useEffect(() => {
+    fetchReportStatus(reportDetails);
+  }, []);
+
+  const submitForm = () => {
+    // TODO: Add check to make sure user filled out the form
+    if (userRole === UserRoles.STATE_USER || userRole === UserRoles.STATE_REP) {
+      updateReportStatus(reportDetails, ReportStatus.COMPLETED);
+    }
+    // TODO: Handle Error for other users coming on to the page
```
braxex, line 51: Other users will be able to see the state of the form but not click the submit button, so this will likely just be a `disabled` button state instead of a check here. Can remove TODO.

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+const ReadyToSubmit = ({
+  submitForm,
+  isOpen,
+  onOpen,
+  onClose,
+}: ReadyToSubmitProps) => {
+  const { intro, modal, pageLink } = reviewVerbiage;
+
+  return (
+    <Flex sx={sx.contentContainer} data-testid="ready-view">
```
braxex, line 88: Nit: Wherever possible, it's best practice to only place `data-testid` in the test files, where it is usually spread onto the element via props, instead of placing the ids in the component code directly.

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+          <Text>{intro.info}</Text>
+        </Box>
       </Box>
-    </ReportPage>
+      <Flex sx={sx.submitContainer}>
+        <Button
+          type="submit"
+          colorScheme="colorSchemes.primary"
+          rightIcon={<Icon icon="arrowRight" />}
+          onClick={() => onOpen()}
```
braxex, line 103: Because you aren't passing any props to `onOpen` you can simplify this. 
```suggestion
          onClick={onOpen}
```

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+        <Box sx={sx.infoTextBox}>
+          <Text sx={sx.infoHeading}>{intro.infoHeader}</Text>
+          <Text>{intro.info}</Text>
+        </Box>
       </Box>
-    </ReportPage>
+      <Flex sx={sx.submitContainer}>
+        <Button
+          type="submit"
+          colorScheme="colorSchemes.primary"
```
braxex, line 101: You actually don't have to use a color scheme here because of the recent updates with buttons. Since the default variant is `primary`, you can just remove this line and it will be the same. 👍 
```suggestion
```

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+          rightIcon={<Icon icon="arrowRight" />}
+          onClick={() => onOpen()}
+          data-testid="submit-button"
```
braxex, line 104: `data-testid` is considered the method of last resort for testing and ideally we're using any other more functional way to get this button during testing. For instance, we can get the element by `type="submit"` because it is the only submission button in the DOM at this point.

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+        >
+          {pageLink.text}
+        </Button>
+      </Flex>
+      <Modal
+        actionFunction={() => submitForm()}
```
braxex, line 110: Same simplification possible here.
```suggestion
        actionFunction={submitForm}
```

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+export const SuccessMessage = ({
+  programName,
+  date,
+  givenName,
+  familyName,
+}: SuccessMessageProps) => {
+  const { intro } = successVerbiage;
+  const utcDateToReadableDate = new Intl.DateTimeFormat("en-US", {
+    dateStyle: "full",
+  }).format(date);
```
braxex, line 137: Would probably move this into time utilities 'cause we'll want to use it later.

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+  const submittedDate = `was submitted on ${utcDateToReadableDate}`;
+  const submittersName = ` by ${givenName} ${familyName}`;
+  return (
+    <Flex sx={sx.contentContainer} data-testid="submitted-view">
```
braxex, line 141: For the test ids like this, we can probably just check for actual text in the dom, like "Successfully submitted" or whatever it says.

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -1,29 +1,215 @@
...
+    fontWeight: "bold",
+    marginBottom: ".5rem",
+  },
+  submitContainer: {
+    justifyContent: "flex-end",
+  },
+  headerImage: {
+    display: "inline-block",
+    marginRight: "1rem",
+    height: "1.72rem",
```
braxex, line 205: Nit: I usually avoid using `rem` in increments smaller than quarters because they're harder to reason about (i.e. off the top of your head, how big is `1.72rem`?). In this case, the difference is less than half a pixel, so we could just use `1.75rem`. Usually you can round to the nearest quarter of a rem but if it has to be precise (for instance maybe it _has to be 27 pixels_ or something) sometimes_ I'll fall back to just directly using pixels because `27px` is obviously not a rem-divisible number and everyone will recognize it has to be a precision thing.

braxex (changes requested):
Looks and feels awesome. Great work.

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
ntsummers1: I thought so too but that pops up with an error:
 `Type 'Function' is not assignable to type 'MouseEventHandler<HTMLButtonElement>'.`

braxex: Best way to solve this would be something like this then:
`onClick={onOpen as MouseEventHandler}`

braxex (approved):

services/ui-src/src/routes/ReviewSubmit/ReviewSubmit.tsx
ntsummers1: So easy I should have seen that 🙃

PR #3552, opened 2022-08-18
add report deletion handler/method

braxex:
## Description

We need to be able to delete reports, and with them, report data. Both of these operations are limited to only the Business Owner Rep (ADMIN) role. Handler has been created but is not currently wired up beyond being made available in the report context (waiting on dashboard completion).  

### How to test

1. Confirm all tests pass. 

### Changed Dependencies

n/a

## Code author checklist


services/app-api/utils/types/types.ts
```diff
@@ -12,11 +12,14 @@ export interface DynamoWrite {
...
+  state: string;
+  reportId: string;
+}
+
 export interface DynamoDelete {
   TableName: string;
-  Key: {
-    key: string;
-  };
+  Key: { key: string } | CompositeKey;
```
BearHanded, line 22: this is going to bite you whenever you have your next Delete. Key should just be any number of prop: string pairs in an object. Imo enforcing that is enough, rather than enforcing every single optional key name as a project grows.

(Really it's max 2 in dynamo, but just enforcing the prop types is enough)

services/app-api/serverless.yml
```diff
@@ -159,6 +159,19 @@ functions:
               paths:
                 state: true
                 reportId: true
+  deleteReportData:
+    handler: handlers/banners/delete.deleteReportData
+    events:
+      - http:
+          path: reportData/{state}/{reportId}
```
BearHanded, line 166: Are both these functions being kicked off by the same path? If they should always run together, why not have them as a single function?

services/app-api/serverless.yml
```diff
@@ -197,6 +210,19 @@ functions:
               paths:
                 state: true
                 reportId: true
+  deleteReport:
+    handler: handlers/banners/delete.deleteReport
+    events:
+      - http:
+          path: reports/{state}/{reportId}
```
BearHanded, line 217: See above comment

services/ui-src/src/utils/api/requestMethods/reportData.ts
```diff
@@ -34,4 +34,18 @@ async function writeReportData(
...
-export { getReportData, writeReportData };
+async function deleteReportData(reportDetails: ReportDetails) {
+  const requestHeaders = await getRequestHeaders();
+  const request = {
+    headers: { ...requestHeaders },
+  };
+  const { state, reportId } = reportDetails;
+  const response = await API.del(
+    "reportData",
+    `/reportData/${state}/${reportId}`,
```
BearHanded, line 45: same endpoint?

services/ui-src/src/utils/api/requestMethods/report.ts
```diff
@@ -40,4 +40,18 @@ async function writeReport(reportDetails: ReportDetails, reportStatus: string) {
...
-export { getReport, getReportsByState, writeReport };
+async function deleteReport(reportDetails: ReportDetails) {
+  const requestHeaders = await getRequestHeaders();
+  const request = {
+    headers: { ...requestHeaders },
+  };
+  const { state, reportId } = reportDetails;
+  const response = await API.del(
+    "reports",
+    `/reports/${state}/${reportId}`,
```
BearHanded, line 51: same endpoint?

services/app-api/handlers/reports/delete.ts
```diff
@@ -0,0 +1,32 @@
...
+    throw new Error(NO_KEY_ERROR_MESSAGE);
+  } else {
+    const params = {
+      TableName: process.env.REPORT_TABLE_NAME!,
+      Key: {
+        state: event?.pathParameters?.state!,
+        reportId: event?.pathParameters?.reportId!,
+      },
+    };
+    await dynamoDb.delete(params);
```
BearHanded, line 29: Does this need to clean up the related data? Shouldn't it make a second dynamoDb.delete() for report data?

BearHanded (changes requested):
Unless I am misunderstanding something, your two endpoints should probably be combined to delete a report and its associated data on request.

services/app-api/utils/types/types.ts
braxex: yeah good call. hard to say where a project will go in the future. will change. 👍

services/app-api/handlers/reports/delete.ts
braxex: I went back and forth on this one, but I think you're right that it makes more sense to keep that logic tied together in the backend. Changed. 👍

services/app-api/serverless.yml
```diff
@@ -159,6 +159,19 @@ functions:
               paths:
                 state: true
                 reportId: true
+  deleteReportData:
+    handler: handlers/banners/delete.deleteReportData
```
BearHanded, line 163: you removed this endpoint right?

services/app-api/serverless.yml
```diff
@@ -197,6 +210,19 @@ functions:
               paths:
                 state: true
                 reportId: true
+  deleteReport:
+    handler: handlers/banners/delete.deleteReport
```
BearHanded, line 214: I think this path got mixed up, should be `handlers/reports/delete.deleteReport`, right?

services/ui-src/src/utils/api/requestMethods/reportData.ts
BearHanded, line 45: this endpoint no longer exists

services/ui-src/src/utils/api/requestMethods/reportData.test.ts
```diff
@@ -7,7 +7,11 @@ describe("Test report methods", () => {
     expect(getReportData(mockReportDetails)).toBeTruthy();
   });
 
-  test("postReport", () => {
+  test("writeReport", () => {
     expect(writeReportData(mockReportDetails, mockReportData)).toBeTruthy();
   });
+
+  test("deleteReport", () => {
+    expect(deleteReportData(mockReportDetails)).toBeTruthy();
```
BearHanded, line 15: reportData enpoint removed

BearHanded (changes requested):
Just a little cleanup remaining

BearHanded (approved):
lgtm

PR #3726, opened 2022-09-01
mobile, tablet and ultra-wide styles

scwambach:
## Description

Added new styles to help the UI in multiple breakpoints. Added a new section that is basically a mirror to the table, but works on smaller screen sizes.

### How to test

create a program in the dashboard and adjust your browser window
[Figma doc for reference](https://www.figma.com/file/GBktI8ZIJM9YYcdNva0ltg/MDCT---MCR?node-id=4113%3A211504)

## Code author checklist
- [ ] I have performed a self-review of my code
- [ ] I have added [thorough](https://bit.ly/3zPrxuZ) tests
- [ ] I have added analytics, if necessary
- [ ] I have updated the documentation, if necessary

## Reviewer checklist (two different people)
- [ ] I have done the deep review and verified the items in the above checklist are g2g
- [ ] I have done the lgtm review

PR #3774, opened 2022-09-06
only admins can delete reports

britt-mo:
## Description

Change permissions allowing only admin users to delete programs.

### How to test

Have a program, log in as an Admin and see that the delete button is still there. Log in as an approver or help desk user and see that the delete button is not there.

## Code author checklist

### DIFF
--- a/services/ui-src/src/routes/Dashboard/Dashboard.test.tsx
+++ b/services/ui-src/src/routes/Dashboard/Dashboard.test.tsx
@@ -11,6 +11,8 @@ import {
   mockHelpDeskUser,
   mockNoUser,
   mockReportStatus,
+  mockStateApprover,
+  mockStateRep,
   mockStateUser,
   RouterWrappedComponent,
 } from "utils/testing/setupJest";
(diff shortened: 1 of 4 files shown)


gmrabian (approved):
Looks good!

services/ui-src/src/routes/Dashboard/Dashboard.test.tsx
```diff
@@ -129,6 +130,16 @@ describe("Test dashboard with admin user", () => {
...
 });
 
+describe("Test dashboard with help desk user", () => {
+  test("Help desk user cannot delete reports", async () => {
+    mockedUseUser.mockReturnValue(mockHelpDeskUser);
+    await act(async () => {
+      await render(dashboardViewWithReports);
+    });
+    expect(screen.queryByAltText("Delete Program")).toBeNull();
+  });
```
braxex, line 140: Might want to add similar tests to ensure state user, state rep, and approver roles all can't delete it either. And depending on the order in which this PR is merged with respect to [this one](https://github.com/CMSgov/mdct-mcr/pull/3726), we'll want to add similar tests for mobile view since the view of that button is repeated there. Might make sense to extract that logic up higher in the component or something so we're not repeating ourselves too much. @scwambach tagging you for visibility on this since you're working on the related dashboard PR.

services/ui-src/src/utils/testing/setupJest.tsx
```diff
@@ -57,6 +57,20 @@ export const mockStateUser: UserContextI = {
   loginWithIDM: () => {},
 };
 
+export const mockHelpDeskUser: UserContextI = {
+  user: {
+    userRole: UserRoles.HELP_DESK,
+    email: "helpdeskuser@test.com",
+    given_name: "Clippy",
+    family_name: "Helperson",
+    full_name: "Clippy Helperson",
```
braxex, line 66: 😆

braxex (approved):
Looks great! I left a note about potentially adding some more tests for peace of mind but it's not necessarily blocking. Let me know your thoughts.

Also, while reviewing I noticed a weird thing that was happening with the Modal styling. It's a one-line fix that I added as a [separate PR branched off yours](https://github.com/CMSgov/mdct-mcr/pull/3780), if you want to take a look/merge it in.

services/ui-src/src/routes/Dashboard/Dashboard.test.tsx
britt-mo: Added the tests for the other users

services/ui-src/src/utils/testing/setupJest.tsx
britt-mo: Just using the names that were already in the file!! Debated making them all Futurama themed in the setupJest file.

PR #3783, opened 2022-09-07
MDCT-1910: Ratio Input

ntsummers1:
## Description
Fixes [MDCT-1920](https://qmacbis.atlassian.net/browse/MDCT-1910)

https://user-images.githubusercontent.com/19439679/188916823-3cab3ee4-3127-4ef8-bbd3-376b89166a1a.mov

### How to test
./dev local
Create a program
Checkout the temporary ratio mask input field on the first program page

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/fields/NumberField.test.tsx
+++ b/services/ui-src/src/components/fields/NumberField.test.tsx
@@ -42,6 +42,10 @@ const percentageMaskedNumberFieldComponent = (
   <NumberField name="testNumberField" label="test-label" mask="percentage" />
 );
 
+const ratioMaskedNumberFieldComponent = (
+  <NumberField name="testNumberField" label="test-label" mask="ratio" />
+);
+
 describe("Test Maskless NumberField", () => {
   test("NumberField is visible", () => {
     const result = render(numberFieldComponent);
@@ -122,6 +126,30 @@ describe("Test Percentage Masked NumberField", () => {
     await userEvent.type(numberFieldInput, "12055.99");
     await userEvent.tab();
     expect(numberFieldInput.value).toEqual("12,055.99");
+  });
+});
+
+describe("Test Ratio Masked NumberField", () => {
+  test("onChangeHandler updates ratio field value", async () => {
+    const result = render(ratioMaskedNumberFieldComponent);
+    const numberFieldInput: HTMLInputElement = result.container.querySelector(
+      "[name='testNumberField']"
+    )!;
+    await userEvent.type(numberFieldInput, "123:123");
+    expect(numberFieldInput.value).toEqual("123:123");
+    await userEvent.tab();
+    expect(numberFieldInput.value).toEqual("123:123");
+    await userEvent.clear(numberFieldInput);
+    await userEvent.type(
+      numberFieldInput,
+      "123,,,4567.1234567.1234:12,3456,7.1"
+    );
+    await userEvent.tab();
+    expect(numberFieldInput.value).toEqual("1,234,567.12:1,234,567.1");
+    await userEvent.clear(numberFieldInput);
+    await userEvent.type(numberFieldInput, ":");
+    await userEvent.tab();
+    expect(numberFieldInput.value).toEqual(":");
   });
 });
 
--- a/services/ui-src/src/forms/mcpar/mcpar.json
+++ b/services/ui-src/src/forms/mcpar/mcpar.json
@@ -21,15 +21,6 @@
             "id": "apoc",
             "fields": [
               {
-                "id": "test1",
-                "type": "number",
-                "props": {
-                  "label": "Ratio Label",
-                  "hint": "Ratio Hint.",
-                  "mask": "ratio"
-                }
-              },
-              {
                 "id": "apoc-a1",
                 "type": "text",
                 "prefill": "stateName",
--- a/services/ui-src/src/forms/mcpar/mcpar.json
+++ b/services/ui-src/src/forms/mcpar/mcpar.json
@@ -21,6 +21,15 @@
             "id": "apoc",
             "fields": [
               {
+                "id": "test1",
+                "type": "number",
+                "props": {
+                  "label": "Ratio Label",
+                  "hint": "Ratio Hint.",
+                  "mask": "ratio"
+                }
+              },
+              {
                 "id": "apoc-a1",
                 "type": "text",
                 "prefill": "stateName",
--- a/services/ui-src/src/forms/mcpar/mcpar.schema.ts
+++ b/services/ui-src/src/forms/mcpar/mcpar.schema.ts
@@ -9,7 +9,6 @@ import {
   nested,
   number,
   radio,
-  ratio,
   text,
   textOptional,
   url,
@@ -18,7 +17,6 @@ import {
 export default {
   apoc: {
     // TODO: change to required fields after prefilled fields are functional
-    test1: ratio(),
     "apoc-a1": textOptional(),
     "apoc-a2a": text(),
     "apoc-a2b": email(),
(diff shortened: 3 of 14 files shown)


services/ui-src/src/utils/forms/schemas.ts
```diff
@@ -9,16 +9,45 @@ export const textOptional = () => text().notRequired();
...
+          return firstTest && secondTest;
+        }
+        return (
+          numberSchema()
+            .transform((_value, originalValue) => {
+              return Number(
+                originalValue.toString().replace(replaceCharsRegex, "")
+              );
+            })
+            .isValidSync(val) || validNAValues.includes(val)
```
ntsummers1, line 48: Would love thoughts on how to make this more readable

ntsummers1:
Would love thoughts on the schemas and mask file about how readable that code is. I've added comments and tried to reduce the code down but it still might be a little complex

services/ui-src/src/components/fields/NumberField.test.tsx
```diff
@@ -125,6 +129,30 @@ describe("Test Percentage Masked NumberField", () => {
...
+    await userEvent.type(
+      numberFieldInput,
+      "123,,,4567.1234567.1234:12,3456,7.1"
+    );
+    await userEvent.tab();
+    expect(numberFieldInput.value).toEqual("1,234,567.12:1,234,567.1");
+    await userEvent.clear(numberFieldInput);
+    await userEvent.type(numberFieldInput, ":");
+    await userEvent.tab();
+    expect(numberFieldInput.value).toEqual(":");
```
gmrabian, line 152: Do we want this to be the case? A single colon is listed in the "bad test cases" below

gmrabian: I guess maybe it's ok to leave it, it'll just fail validation, right?

services/ui-src/src/forms/mcpar/mcpar.json
```diff
@@ -20,6 +20,15 @@
...
             "fields": [
+              {
+                "id": "test1",
+                "type": "number",
+                "props": {
+                  "label": "Ratio Label",
+                  "hint": "Ratio Hint.",
+                  "mask": "ratio"
+                }
+              },
```
gmrabian, line 31: Just making this note so we remember to remove this before merge

services/ui-src/src/forms/mcpar/mcpar.schema.ts
```diff
@@ -17,6 +17,7 @@ import {
 export default {
   apoc: {
     // TODO: change to required fields after prefilled fields are functional
+    test1: number(true),
```
gmrabian, line 20: same as above - remove before merge

services/ui-src/src/utils/forms/schemas.test.tsx
```diff
@@ -14,15 +14,54 @@ describe("Schemas", () => {
...
-    testCase(badNumberTestCases, false);
+  const testNumberSchemaWithRatioPassed = (
+    testCases: Array<string>,
+    expectedReturn: boolean
+  ) => {
+    for (let testCase of testCases) {
+      let test = number(true).isValidSync(testCase);
+      expect(test).toEqual(expectedReturn);
+    }
+  };
```
gmrabian, line 56: This is the same as the `testNumberSchema` method above. Do we need both, and both tests? Could we not just do one?

services/ui-src/src/utils/forms/schemas.ts
```diff
@@ -9,16 +9,55 @@ export const textOptional = () => text().notRequired();
...
 const validNAValues = ["N/A", "Data not available"];
 
-export const number = () =>
+const ignoreCharsForSchema = (value: string, charsToReplace: RegExp) => {
+  return numberSchema().transform((_value) => {
+    return Number(value.replace(charsToReplace, ""));
+  });
+};
+
+export const number = (isRatio = false) =>
```
gmrabian, line 18: Is it possible to make a separate ratio field that validates against number : number ? I feel like that'd be pretty smooth

services/ui-src/src/components/fields/NumberField.test.tsx
ntsummers1: Right, that'll fail validation so it wouldn't get through on the form. Basically this test is to make sure the masking gets called and that mask returns the right value back when you tab away.

services/ui-src/src/utils/forms/schemas.test.tsx
ntsummers1: They were slightly different, however I reduced it down to the same function now as I agree that they are too similar.

services/ui-src/src/utils/forms/schemas.ts
ntsummers1: Separated out!

gmrabian: I'm glad this is separated, though my idea was wondering if it's possible to do like
`ratio: mixed(number(), ":", number())` if that makes sense. Like having the schema validate there are two numbers with a colon. Maybe there's no easy way to do that except for regex. Just curious since that would simplify it a lot!

ntsummers1: Not to my knowledge, no :( .. And if i try to do something like that then I'd come back to the complexity before I separated it out.

services/ui-src/src/forms/mcpar/mcpar.json
braxex: +1

services/ui-src/src/forms/mcpar/mcpar.schema.ts
```diff
@@ -17,6 +18,7 @@ import {
 export default {
   apoc: {
     // TODO: change to required fields after prefilled fields are functional
+    test1: ratio(),
```
braxex, line 21: remove before merge

services/ui-src/src/utils/forms/schemas.ts
```diff
@@ -27,6 +35,46 @@ export const number = () =>
...
+
+        // If both sides are valid numbers, return true!
+        return firstTest && secondTest;
+      },
+    })
+    .test({
+      message: error.REQUIRED_GENERIC,
+      test: (val) => val != "",
+    })
+    .required(error.REQUIRED_GENERIC);
```
braxex, line 76: Required message should fire when the field is empty instead of the ratio message. Just need to move these two above the first test. Might need to make same change for Number.

services/ui-src/src/utils/other/mask.test.ts
```diff
@@ -71,6 +83,29 @@ describe("Test convertToCommaSeparatedString", () => {
...
+  test("Valid ratio is split and masked on each side individually", () => {
+    for (let testCase of ratioMaskAcceptableTestCases) {
+      expect(convertToCommaSeparatedRatioString(testCase.test)).toEqual(
+        testCase.expected
+      );
+    }
+  });
+  test("Check if non number strings can fail gracefully", () => {
+    const testCases = [
+      { test: "No numbers here me lord", expected: ":" },
```
braxex, line 96: If we're gonna go here, might as well go all in. 😉 
```suggestion
      { test: "No numbers here, m'lord", expected: ":" },
```

braxex (approved):
One change needed in the order of tests for ratio validation, but otherwise approved once that's done. Remember to take out the test cases before merging.

gmrabian (approved):
works nicely locally

services/ui-src/src/utils/other/mask.test.ts
ntsummers1: Fixed :)

services/ui-src/src/utils/forms/schemas.ts
ntsummers1: Worked like a charm. Fixed!

PR #3807, opened 2022-09-07
MDCT-1901: Form Refresh -> User NOT Kicked Out to Dashboard

karla-vm:
## Description

Closes [MDCT-1901](https://qmacbis.atlassian.net/browse/MDCT-1901).

Right now, a user can refresh at any point in a report form and is kicked out to the report dashboard 💔 
With these changes, that shouldn't happen because we're storing the `reportId` in local storage!

P.S. Also, added a condition to ensure that `selectedState` is set in local storage only for **ADMIN, APPROVER, HELP_DESK** users.

### How to test

- `./dev local`
- Log in as any user (e.g. `stateuser2@test.com` or `adminuser@test.com`)
- Navigate to a report section
- Refresh the page ... and stay there ✨ 

### Changed Dependencies

N/A

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/reports/ReportPage.tsx
+++ b/services/ui-src/src/components/reports/ReportPage.tsx
@@ -24,7 +24,8 @@ import {
 
 export const ReportPage = ({ reportJson, route }: Props) => {
   // get report, form, and page related-data
-  const { report, updateReportData, updateReport } = useContext(ReportContext);
+  const { report, setReport, updateReportData, updateReport } =
+    useContext(ReportContext);
   const reportId = report?.reportId || localStorage.getItem("selectedReport");
   const { basePath, routes } = reportJson;
   const { form, page } = route;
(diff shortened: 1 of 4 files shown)


gmrabian:
I know this is still WIP so ignore me if you're already on this - I was just playing around with this and noticed when you refresh it does not rehydrate the current page. So where the fields were once filled out, they are now blank (it's possible the values are still there in the form context and/or reportData/hydration, though they are removed from the DOM and validation makes it seem like they aren't there)

What if we started checking for reportId from local storage instead of from context? That way refresh probably wouldn't have the useState clearing issue.

Just some ideas!

karla-vm:
> I know this is still WIP so ignore me if you're already on this - I was just playing around with this and noticed when you refresh it does not rehydrate the current page. So where the fields were once filled out, they are now blank (it's possible the values are still there in the form context and/or reportData/hydration, though they are removed from the DOM and validation makes it seem like they aren't there)
> 
> What if we started checking for reportId from local storage instead of from context? That way refresh probably wouldn't have the useState clearing issue.
> 
> Just some ideas!
> 
> Also since Admin puts selected state in storage, I'm wondering if state users should just permanently have their state in storage?

I like that idea of having state users' corresponding states persisting in storage! That would just make our lives easier if we ever need to pull that data in. 

re: checking for the `reportId`, where were you thinking we should switch out pulling from context to pulling from storage? We're doing something like that in `ReportPage.tsx`, but that might not be the best place for it! What do you think?

gmrabian (approved):

PR #3804, opened 2022-09-07
MDCT-1857 pt. 1 -- add form table and api

gmrabian:
## Description
Partial [MDCT-1857](https://qmacbis.atlassian.net/browse/MDCT-1857) Adds the backend table and API by which we will fetch exact forms or push forms.

### How to test

Ask me for help if this is confusing

There is no frontend component yet so you can test against the API directly
- `./dev local`
- View the Network tab on your browser. Find an API request (like banner or dashboard) and get the `x-api-key` security header value. Put that key: value in your preferred API testing tool (I use Postman)
- `POST localhost:3030/local/forms` to create a form (needs `formId` and `formTemplate` in the body)
- `GET localhost:3030/local/forms/{formId}` to test getting a specific form, using the formId you created

### Changed Dependencies
**Add one line to your root .env file (it's in the README for reference)**

## Code author checklist
- [x] I have performed a self-review of my code
- [x] I have added [thorough](https://bit.ly/3zPrxuZ) tests
- [x] I have added analytics, if necessary
- [x] I have updated the documentation, if necessary

## Reviewer checklist (two different people)
- [ ] I have done the deep review and verified the items in the above checklist are g2g
- [ ] I have done the lgtm review

PR #3823, opened 2022-09-08
MDCT-1857 pt. 2 --add frontend components for form template api

gmrabian:
## Description
Partial [MDCT-1857](https://qmacbis.atlassian.net/browse/MDCT-1857) Adds frontend api components and writes form to db when adding a program

Follows [PR 3804](https://github.com/CMSgov/mdct-mcr/pull/3804)

### How to test
- `./dev local`
- State user -> dashboard -> create program
- DynamoDb admin -> see report and form template, ids should match
- When you edit a report's details from the dashboard the form template id does not change

### Changed Dependencies
Need root .env addition if you haven't already
`FORM_TEMPLATE_TABLE_NAME=local-mcpar-form-templates`

## Code author checklist


braxex (approved):
This all looks good to me.

services/ui-src/src/forms/mcpar/mcpar.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "MCPAR",
   "basePath": "/mcpar",
-  "version": "0.0.0",
+  "version": "MCPAR_Sep-8-2022",
```
braxex, line 4: Nit: My personal preference is for a numeric-only date and I don't think design will ever see this, so I think we should go for that. I'm not the only person with opinions though so if everyone else wants to go with something like this instead that's cool.

gmrabian: 🤷 sure. Doesn't matter to me

gmrabian, line 4:
```suggestion
  "version": "MCPAR_2022-09-08", // yyyy-mm-dd
```

PR #3835, opened 2022-09-08
Disable forms for admins

britt-mo:
## Description

Disable report fields for admin users

### How to test

log in as an admin and see that the report fields are disabled (including nested fields), and report fields for state users are not disabled.

## Code author checklist


services/ui-src/src/forms/mcpar/index.ts
```diff
@@ -3,15 +3,19 @@ import reportSchema from "./mcpar.schema";
 // utils
 import {
   addValidationToReportJson,
+  copyAdminDisabledStatusToForms,
   flattenReportRoutesArray,
 } from "utils/reports/reports";
 import { ReportJson } from "types";
 
+const stateOrAdminVersion: ReportJson =
+  copyAdminDisabledStatusToForms(rawReportJson);
```
braxex, line 12: Nit: Would maybe name this something like `reportJsonWithDisabledStatus` instead.
```suggestion
const reportJsonWithDisabledStatus: ReportJson =
  copyAdminDisabledStatusToForms(rawReportJson);
```

services/ui-src/src/utils/reports/reports.test.ts
```diff
@@ -112,3 +118,39 @@ describe("Sort Reports", () => {
     expect(sortReportsOldestToNewest(unsortedReports)).toEqual(sortedReports);
   });
 });
+
+describe("Test copyAdminDisabledStatusToForms", () => {
+  const newReportJson: ReportJson = {
```
braxex, line 123: We actually have some mockReportJson available [here in setupJest.tsx](https://github.com/CMSgov/mdct-mcr/blob/ddb762731824513e0bd8a78e977c475709a4d036/services/ui-src/src/utils/testing/setupJest.tsx#L230-L235) you could use.

braxex (changes requested):
Looks good! There are a couple things yet to do but we're getting close:

1. Dynamic fields (see form `aap` or `absse`) aren't disabled. This is because the props aren't getting fully passed down into the TextField components that are rendered. Adding `{...props}` to line 40 of `DynamicField` should do the trick.
2. Validation is still running on the forms when admin users click 'Save & continue'. We should stop that from happening. Probably the easiest way is when constructing the form in Form.tsx, we could conditionally tell it if the form needs validation or not by passing the resolver something like this:
```
  const form = useForm({
    resolver: !shouldDisableAllFields ? yupResolver(formSchema) : undefined,
```
3. The 'Save and continue' button probably shouldn't say that anyway. Maybe just 'Continue' if the form is disabled. Would need to modify this in ReportPage.

services/ui-src/src/components/forms/Form.tsx
```diff
@@ -6,15 +6,27 @@ import { yupResolver } from "@hookform/resolvers/yup";
...
   const formSchema = yupSchema(formJson.validation || {});
 
+  // determine if fields should be disabled (based on admin roles )
+  const { userRole } = useUser().user ?? {};
+  const isAdminUser =
+    userRole === UserRoles.ADMIN ||
+    userRole === UserRoles.APPROVER ||
+    userRole === UserRoles.HELP_DESK;
+  const shouldDisableAllFields = isAdminUser && formJson.adminDisabled;
+
```
braxex, line 29: Nit: Would maybe rename this something like `formInputDisabled` or `fieldInputDisabled` or something so when we use it below to determine if a resolver is set it reads a bit better.

ntsummers1 (approved):
Nice work! I wasn't able to break anything on this and the codes looking good.

gmrabian (changes requested):
Looks like the dynamic field delete button still showed up for admin, and I was able to delete fields. It doesn't actually remove them from the database, and they show up the next time you enter the form. But we should not allow this to be here even so

Screen Shot 2022-09-12 at 8 44 52 AM



Screen Shot 2022-09-12 at 8 44 56 AM

ntsummers1 (approved):

ntsummers1:
Garretts callout seems fixed. Looks great on my end

gmrabian (approved):
looks good!

gmrabian (approved):

PR #3854, opened 2022-09-09
MDCT 1917/1899: Hydrate when necessary and pre-fill AddEditModal if editing

ntsummers1:
## Description

Fixes [1917](https://qmacbis.atlassian.net/browse/MDCT-1917) and [1899](https://qmacbis.atlassian.net/browse/MDCT-1899)

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/forms/Form.tsx
+++ b/services/ui-src/src/components/forms/Form.tsx
@@ -57,7 +57,7 @@ export const Form = ({
   // hydrate and create form fields using formFieldFactory
   const renderFormFields = (fields: FormField[]) => {
     const fieldsToRender = hydrateFormFields(fields, formData);
-    return formFieldFactory(fieldsToRender, fieldInputDisabled);
+    return formFieldFactory(fieldsToRender, fieldInputDisabled || false);
   };
 
   return (
--- a/services/ui-src/src/components/forms/Form.tsx
+++ b/services/ui-src/src/components/forms/Form.tsx
@@ -12,7 +12,7 @@ export const Form = ({
   id,
   formJson,
   onSubmit,
-  formData,
+  formData = {},
   children,
   ...props
 }: Props) => {
--- a/services/ui-src/src/components/forms/Form.tsx
+++ b/services/ui-src/src/components/forms/Form.tsx
@@ -6,7 +6,7 @@ import { yupResolver } from "@hookform/resolvers/yup";
 import { Box } from "@chakra-ui/react";
 // utils
 import { formFieldFactory, hydrateFormFields, sortFormErrors } from "utils";
-import { FormJson, FormField } from "types";
+import { FormJson, FormField, AnyObject } from "types";
 
 export const Form = ({
   id,
@@ -64,7 +64,7 @@ interface Props {
   id: string;
   formJson: FormJson;
   onSubmit: Function;
-  formData?: any;
+  formData?: AnyObject;
   children?: ReactNode;
   [key: string]: any;
 }
--- a/services/ui-src/src/components/forms/Form.tsx
+++ b/services/ui-src/src/components/forms/Form.tsx
@@ -1,17 +1,22 @@
-import { ReactNode, useContext } from "react";
+import { ReactNode } from "react";
 import { FormProvider, useForm } from "react-hook-form";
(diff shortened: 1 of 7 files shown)


gmrabian:
seeing this bug: form field shows old hydration value in the wrong form. (Also noticed this bug on [PR3859](https://github.com/CMSgov/mdct-mcr/pull/3859). It's not an issue on main.)
https://user-images.githubusercontent.com/57802560/189680303-3bcb6360-28df-4fc8-bf7e-d05957adac83.mov

also posted in [a comment on referenced PR](https://github.com/CMSgov/mdct-mcr/pull/3859#issuecomment-1243796922)

gmrabian (approved):
Tried it out locally. Looks great!

services/ui-src/src/components/forms/Form.tsx
```diff
@@ -15,9 +14,16 @@ import {
 } from "utils";
 import { AnyObject, FormJson, FormField, UserRoles } from "types";
 
-export const Form = ({ id, formJson, onSubmit, children, ...props }: Props) => {
+export const Form = ({
+  id,
+  formJson,
+  onSubmit,
+  formData = {},
```
braxex, line 21: I think you should be able to leave this undefined since it's optional. Haven't tested it though so I might be missing something big-picture.

services/ui-src/src/components/forms/Form.tsx
```diff
@@ -50,8 +56,8 @@ export const Form = ({ id, formJson, onSubmit, children, ...props }: Props) => {
 
   // hydrate and create form fields using formFieldFactory
   const renderFormFields = (fields: FormField[]) => {
-    const fieldsToRender = hydrateFormFields(fields, reportData);
-    return formFieldFactory(fieldsToRender, fieldInputDisabled!);
+    const fieldsToRender = hydrateFormFields(fields, formData);
+    return formFieldFactory(fieldsToRender, fieldInputDisabled || false);
```
braxex, line 60: Would !! work here?
```suggestion
    return formFieldFactory(fieldsToRender, !!fieldInputDisabled);
```

services/ui-src/src/components/modals/AddEditProgramModal.tsx
```diff
@@ -18,6 +18,7 @@ import { mcparReportJsonNested } from "forms/mcpar";
 export const AddEditProgramModal = ({
   activeState,
   selectedReportId,
+  selectedReportData,
```
braxex, line 21: Because we already have the concept of `reportData`, maybe we could name all the `reportData` stuff in this PR to `reportMetadata` or something? What does everyone think?

services/ui-src/src/routes/Dashboard/Dashboard.tsx
```diff
@@ -52,6 +52,9 @@ export const Dashboard = () => {
   const [selectedReportId, setSelectedReportId] = useState<string | undefined>(
     undefined
   );
+  const [selectedReportFormData, setSelectedReportFormData] = useState<
+    AnyObject | undefined
+  >(undefined);
```
braxex, line 57: Rather than setting and passing it to the modal separately, what if we changed the `setReportId` to not just take the ID but the entire report metadata object, so this would become something like below, then we'd just pass the whole thing in to the modal? 
```suggestion
  const [selectedReportMetadata, setSelectedReportMetadata] = useState<
    AnyObject | undefined
  >(undefined);
```

services/ui-src/src/routes/Dashboard/Dashboard.tsx
```diff
@@ -89,6 +92,30 @@ export const Dashboard = () => {
   const openAddEditProgramModal = (reportId?: string) => {
     // if reportId provided, set as selected program
     setSelectedReportId(reportId);
+
+    const selectedReport = reportsByState?.find(
+      (o: { reportId: string }) => o.reportId === reportId
+    );
```
braxex, line 98: If we pass in the whole report metadata object to this function instead of passing just the reportId, we can save ourselves this `.find` and make the checks below more clear too.

braxex (approved):
Looking really good. Left a few notes that might help us out with naming down the road. Let me know your thoughts.

PR #3860, opened 2022-09-11
implement server-side validation

braxex:
## Description

Well, we needed server-side validation -- all data needs to be validated before it hits the database. 

Changes:
- Added server-side validation in general via the `validateData` utility in _/utils/validation/validation_.
- Added server-side validation to `reportData` write call. This required:
    - Added `fetchReport` call to get the corresponding `formTemplateId`. _(/handlers/reportData/write.ts)_
- Added server-side validation (with custom server-side validation schema) to the following `write` calls:
    - Write `report` metadata on report creation or edit via `AddEditProgramModal` _(/handlers/reports/write.ts)_
    - Write `banner` data from `AddAdminBanner` _(/handlers/banners/write.ts)_
- Fixed related tests and added some more

### How to test

1. Test the full lifecycle of creating a new report, entering it, filling out and saving information. See that everything works as expected.
2. Try to break the validation by sending invalid data. (Note: The easiest way to do this is to subtly alter the server-side validation schemas, say from `string` to `number` and submit the same data. That way you don't have to disable the client-side inline field validation.)

### Changed Dependencies

Added `yup` to `package.json` in app-api. You will have to run `yarn` in there to get it up and running the first time.

## Code author checklist


ntsummers1:
One issue I'm seeing is trying to click Save & Continue after filling out the first part of the program. Doing so returns:
```
VALIDATION ERRORS ValidationError: combinedData is a required field
      api|     at createError (/Users/nicholassummers/Desktop/MDCT/mdct-mcr/services/app-api/node_modules/yup/lib/util/createValidation.js:54:21)
      api|     at /Users/nicholassummers/Desktop/MDCT/mdct-mcr/services/app-api/node_modules/yup/lib/util/createValidation.js:72:107
      api|     at processTicksAndRejections (internal/process/task_queues.js:95:5) {
      api|   value: { status: 'In progress', lastAlteredBy: 'Carol California' },
      api|   path: 'combinedData',
      api|   type: 'required',
      api|   errors: [ 'combinedData is a required field' ],
      api|   params: {
      api|     value: undefined,
      api|     originalValue: undefined,
      api|     label: undefined,
      api|     path: 'combinedData'
      api|   },
      api|   inner: []
      api| }
      api| 
      api| 2022-09-12T17:23:01.778Z API event {
      api|   body: '{"status":"In progress","lastAlteredBy":"Carol California"}',
      api|   pathParameters: { state: 'CA', reportId: '7a84a8-a8e5-e285-3d8c-a4e6f2f7d81' },
      api|   queryStringParameters: null
      api| }
```

Looks to me like its required to grab the combinedData checkbox thats included on the AddEditProgramModal, but the program page wouldn't have that checkbox on there. Maybe then combinedData is not required and optionally attached?

ntsummers1:
Following the note above - When an error happens and the report can't write to the DB and validate, it still progresses seemingly fine on the front-end. There'll be an error in the console but the user won't know until they've gotten to the review and submit where the page simply won't update after the submission fails so it looks like nothing happened. For clarification of what i'm talking about, heres a simple video showing this:

https://user-images.githubusercontent.com/19439679/189721127-7c7e6c07-f720-4de4-be8f-3885092208f3.mov

I think it would be good to toss an error banner on the SimpleFormPage as well as the ReviewSubmit page so that we can let the user know something went wrong here

services/app-api/handlers/reports/write.test.ts
```diff
@@ -63,13 +77,26 @@ describe("Test writeReport API method", () => {
...
+      headers: {
+        "Access-Control-Allow-Origin": "string",
+        "Access-Control-Allow-Credentials": true,
+      },
+      body: "{}",
+    });
+    const res = await writeReport(creationEventWithInvalidData, null);
+    expect(res.statusCode).toBe(StatusCodes.SERVER_ERROR);
+  });
+
```
ntsummers1, line 99: This is just one test of many you wrote, but great job on writing all these tests out. I'm learning so much from reading through them

services/app-api/utils/validation/validation.ts
```diff
@@ -0,0 +1,85 @@
...
+  options?: AnyObject
+) => {
+  try {
+    // returns valid data to be passed through API
+    return await validationSchema.validate(data, {
+      stripUnknown: true,
+      ...options,
+    });
+  } catch (e: any) {
+    console.log("VALIDATION ERRORS", e); // eslint-disable-line no-console
```
ntsummers1, line 19: Not sure if we want to keep these kinds of console logs going forward, so dropping a note here

services/ui-src/src/components/modals/AddEditProgramModal.tsx
```diff
@@ -31,6 +31,10 @@ export const AddEditProgramModal = ({
     // prepare payload
     const programName = formData["aep-programName"];
     const dueDate = calculateDueDate(formData["aep-endDate"]);
+    const combinedDataArray = formData["aep-combinedData"];
+    const combinedData =
+      combinedDataArray?.[0] ||
+      "No, the report does not contain items for which the state is unable to remove Separate CHIP information";
```
ntsummers1, line 37: Could abstract the "No, the report does not contain items.." out to the global constants file and use there

karla-vm (commented):
This is awesome. I tried submitting without responding to required fields and it won't allow me to submit, but there's no feedback to the user on the frontend, like @ntsummers1 mentioned. I'm thinking we could throw a generic (or specific, either way) message to users letting them know _why_ they can't submit just yet. So, maybe a banner with "Please respond to all required fields" or something.

services/app-api/utils/validation/validation.ts
braxex: Removed. I had initially thought they might be good to keep in the console but they do make the tests noisier for kind of no reason, so 🤷

braxex:
@ntsummers1 @karla-vm Great callout on that error ([fixed here](https://github.com/CMSgov/mdct-mcr/pull/3860/commits/6b59e475041b4d6461444b202e6bd22de70fb7db)). I shouldn't have had those fields listed as required for the `writeReport` call that's chained to the `writeReportData` call. When we update report data by entering data on any given page, we also send a separate update to the report metadata that tells it who last edited it and when. That call was failing server-side validation because it didn't have most of the fields that were previously marked required but shouldn't have been in the first place. 

As for how we are alerting users to failures, which you both brought up:
We might want to add more user messaging in the future, but if we do, we'll probably just use the client-side `onError` method defined in `Form.tsx`, since it won't need to be asynchronous and is performed before any database calls are sent. There's an outside chance we'll want to do something with server-side validation messaging in the future, but I don't think it's something we need right now, since client-side validation will catch any issues. For now, we have the inline client error messaging and validation errors. 👍

braxex:
Note to reviewers: There are CodeClimate warnings because we now have more code that is duplicated in frontend/backend. For now we can disregard that, but I'm open to ideas as to how we could possibly leverage Serverless to expose those files both front and back instead of writing and maintaining them in two locations.

britt-mo (commented):
Do we want to have validation that a user has entered any words? 

Screen Shot 2022-09-13 at 3 37 58 PM

services/app-api/handlers/banners/write.ts
```diff
@@ -16,23 +18,39 @@ export const writeBanner = handler(async (event, _context) => {
...
+        Item: {
+          key: event.pathParameters.bannerId,
+          createdAt: Date.now(),
+          lastAltered: Date.now(),
+          lastAlteredBy: event?.headers["cognito-identity-id"],
+          title: unvalidatedPayload.title,
+          description: unvalidatedPayload.description,
+          link: unvalidatedPayload.link,
+          startDate: unvalidatedPayload.startDate,
+          endDate: unvalidatedPayload.endDate,
```
gmrabian, line 49: Should these come from the `validatedPayload` object instead? Values should be the same and I think it reads better if we're storing the validated values

services/app-api/handlers/formTemplates/write.ts
```diff
@@ -15,24 +17,46 @@ export const writeFormTemplate = handler(async (event, _context) => {
...
+
+    if (validatedPayload) {
+      const params = {
+        TableName: process.env.FORM_TEMPLATE_TABLE_NAME!,
+        Item: {
+          formTemplateId: unvalidatedPayload.formTemplateId,
+          createdAt: Date.now(),
+          lastAltered: Date.now(),
+          lastAlteredBy: event?.headers["cognito-identity-id"],
+          formTemplate: unvalidatedPayload.formTemplate,
```
gmrabian, line 54: same note as above: using validatedPayload to read instead

services/app-api/handlers/formTemplates/write.ts
```diff
@@ -15,24 +17,46 @@ export const writeFormTemplate = handler(async (event, _context) => {
...
-        createdAt: Date.now(),
-        lastAltered: Date.now(),
-        lastAlteredBy: event?.headers["cognito-identity-id"],
-        formTemplate: body.formTemplate,
-        ...body,
-      },
-    };
-    await dynamoDb.put(params);
-    return { status: StatusCodes.SUCCESS, body: params };
+    const validationSchema = yup.object().shape({
```
gmrabian, line 26: Wondering if we want to turn these schemas into "objects/models" of sorts (to use the Java term). Meaning we'd store them elsewhere. Either in a types-esque file or in separate files per schema

services/app-api/handlers/banners/write.ts
braxex: Yes. To do.

services/app-api/handlers/formTemplates/write.ts
braxex: Not a bad idea.

services/app-api/handlers/formTemplates/write.ts
braxex: 👍

services/ui-src/src/components/modals/AddEditProgramModal.tsx
braxex: Good idea

gmrabian (changes requested):
The backend isn't saving properly. I filled out a bunch of data and it's looking like this. First guess is it might be an issue with naming (reportData vs. formData?)
```
{
  "reportId": "334b13-08e5-c7cc-ccb-5c40a01d234",
  "fieldData": {},
  "state": "AL"
}
```

gmrabian (approved):
lgtm

services/ui-src/src/forms/mcpar/mcpar.json
```diff
@@ -113,7 +113,7 @@
                 "id": "arp-a5b",
                 "type": "date",
                 "hydrate": "reportingPeriodEndDate",
-                "validation": "dateOptional",
+                "validation": "textOptional",
```
ntsummers1, line 116: Why the swap here?

braxex: Uh... I can't remember. I'm thinking they should be date. I will change that.

services/app-api/handlers/banners/write.ts
```diff
@@ -16,23 +18,39 @@ export const writeBanner = handler(async (event, _context) => {
...
+          lastAltered: Date.now(),
+          lastAlteredBy: event?.headers["cognito-identity-id"],
+          title: validatedPayload.title,
+          description: validatedPayload.description,
+          link: validatedPayload.link,
+          startDate: validatedPayload.startDate,
+          endDate: validatedPayload.endDate,
+        },
+      };
+      await dynamoDb.put(params);
```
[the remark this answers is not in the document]
braxex: Yes, and I believe if it were to fail the dynamo lib would return an error for us.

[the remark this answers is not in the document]
gmrabian: Any errors thrown in a custom handler route up to the global handler-lib.ts file where they'll return a 500

gmrabian (approved):
Tested it out. Saw no issues locally

PR #3930, opened 2022-09-18
MDCT-1891: Add dynamic entities to DrawerSections

ntsummers1:
## Description

Fixes [MDCT-1891](https://qmacbis.atlassian.net/browse/MDCT-1891)

A picture is worth 1000 words, so heres my million word essay

https://user-images.githubusercontent.com/19439679/190922859-ff43d8c1-3936-4b8a-8fd7-5949d1af41f9.mov

### How to test
./dev local
Create a program
Go to Section D and see that because theres no plans created you'll get a "No Plans Available" message.
Go to back to Section A and add as many plans as you want
Go back to Section D and see all your new plans in their dynamic glory.

## Code author checklist


services/ui-src/src/types/index.ts
```diff
@@ -113,12 +113,19 @@ export interface ReportContextShape extends ReportContextMethods {
 
 // FORM & FIELD STRUCTURE
 
+export declare type PageEntityType =
+  | "plan"
+  | "bssEntity"
+  | "qualityMeasures"
+  | "someOtherEntity";
```
ntsummers1, line 120: Did we have specific values for this? Is it just the first 3 for now?

services/ui-src/src/components/reports/ReportPage.tsx
```diff
@@ -76,6 +78,20 @@ export const ReportPage = ({ route }: Props) => {
     }
   };
 
+  const getEntities = (formEntityType: PageEntityType) => {
+    const entityTypeMap = {
+      plan: "aap-1", // reportData fieldId of the entity array
+      bssEntity: "absse-a8",
+      qualityMeasures: "apoc-3", // To Be Added
+      someOtherEntity: "apoc-4", // To Be Added
```
ntsummers1, line 86: See callout further down

services/ui-src/src/components/reports/sections/StaticDrawerSection.tsx
```diff
@@ -16,27 +28,28 @@ export const StaticDrawerSection = ({ form, drawer, onSubmit }: Props) => {
     onOpen();
   };
 
-  const tempEntityMap = {
-    plans: ["Plan A", "Plan B", "Plan C"],
-  };
+  const emptyEntitiesMessage = (
+    <Text sx={sx.emptyEntityMessage}>Please enter a plan</Text>
```
ntsummers1, line 32: Update pending designs feedback on what this should actually say

ntsummers1:
Did we need to also do the Dynamic Drawer updates in this PR as well? @braxex

services/ui-src/src/types/index.ts
braxex: Yep it's actually just the first two for now. We're actually going to be changing the names of the fields too, so the field name for plans will not be `aap-1`, it will be `plans`, likewise with `arp-1` → `bssEntities`.

services/ui-src/src/components/reports/ReportPage.tsx
```diff
@@ -76,6 +78,20 @@ export const ReportPage = ({ route }: Props) => {
     }
   };
 
+  const getEntities = (formEntityType: PageEntityType) => {
+    const entityTypeMap = {
+      plan: "aap-1", // reportData fieldId of the entity array
+      bssEntity: "absse-a8",
```
braxex, line 84: This is exactly what I was envisioning. We'll need to modify because the field names will be changing to `plans` and `bssEntities`, so I wonder if the form JSON prop maybe becomes like `mappedEntityField` or something.

services/ui-src/src/components/reports/ReportPage.tsx
```diff
@@ -76,6 +78,20 @@ export const ReportPage = ({ route }: Props) => {
...
+      qualityMeasures: "apoc-3", // To Be Added
+      someOtherEntity: "apoc-4", // To Be Added
+    };
+    const pageEntityType: keyof typeof entityTypeMap = formEntityType;
+    const entitiesToFetch = entityTypeMap[pageEntityType];
+    const entities: string[] | undefined =
+      reportData?.fieldData[entitiesToFetch];
+    return entities;
+  };
```
braxex, line 93: Because we're just passing all this stuff to the Entity section anyway, we should just pull in `reportData` to the entity section and do this all there rather than pass it through from `ReportPage`.

braxex (dismissed):
Looking good!

services/ui-src/src/components/reports/ReportPage.tsx
ntsummers1: I thought this too, but then I wasn't sure if we also wanted to use it for the `DynamicDrawerSection` so I moved it up to the `ReportPage`. I can stick it in the `StaticDrawerSection` then and do there if we aren't going to use it in both places

services/ui-src/src/components/reports/ReportPage.tsx
```diff
@@ -76,6 +78,18 @@ export const ReportPage = ({ route }: Props) => {
     }
   };
 
+  const getEntities = (formEntityType: MappedEntityType) => {
+    const entityTypeMap = {
+      plan: "plans", // reportData fieldId of the entity array
+      bssEntity: "bssEntities",
+    };
```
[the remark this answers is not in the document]
ntsummers1: Removed!

braxex (approved):

PR #3992, opened 2022-09-22
hide deleteReport UI from val & prod

braxex:
## Description

So I had a thought: we're removing deleteReport functionality from the API, but archiveReport functionality won't be in place by the time the pen test starts, which would leave the deleteReport UI visible to testers without it being connected to anything, which could result in confusion. 

So I removed the deleteReport button from the UI in val & prod environments until archiveReport functionality can be wired up, which should be done midway through the pen test.

### How to test

1. After deployment, check that deleteReport UI is visible in main but not in val & prod.

### Changed Dependencies

n/a

## Code author checklist


services/ui-src/src/routes/Dashboard/Dashboard.test.tsx
```diff
@@ -7,11 +7,13 @@ import { Dashboard } from "routes";
 import { ReportContext } from "components";
 // utils
 import {
-  mockAdminUser,
-  mockHelpDeskUser,
+  /*
```
braxex, line 10: I commented out the code in this file so @TheDanMiller has it for reference when implementing archive functionality. Just thought it would be more helpful than throwing it away since he'll be doing this so soon.

karla-vm (approved):
Lookin' good!

ntsummers1 (approved):

PR #3994, opened 2022-09-22
upgrade user role utility

braxex:
## Description

In the olden days if you wanted to see if a user was an admin-type user, you had to do this:
```
import { useUser } from "utils";
import { UserRoles } from "types";
const { userRole } = useUser().user ?? {};
const isStateTypeUser = userRole === UserRoles.STATE_USER || userRole === UserRoles.STATE_REP;
```
Now you can just do this:

```
import { useUser } from "utils";
const { userIsStateUser, userIsStateRep } = useUser().user ?? {};
const isStateTypeUser = userIsStateUser || userIsStateRep;Ï
```

It's a small change, but one that hopefully makes your life easier.

### How to test

 1. All tests pass
 2. Admin functions still show up for admin users (dropdown on homepage,  delete button on dashboard, disabled fields on report entry, /admin route, banner admin button on /profile).
 3. State functions still show up for state users (normal homepage, add/edit buttons on dashboard, normal form fill experience, no access to /admin route, no banner admin button on /profile).

### Changed Dependencies

n/a

## Code author checklist


gmrabian (approved):
Looks good and works locally.

I wonder if in the future we should consider also setting distinct types since many of these roles often share the same use cases. But for now let's keep it at this.

karla-vm (approved):
👩🏽‍🍳 💋

ntsummers1 (approved):

braxex:
> I wonder if in the future we should consider also setting distinct types since many of these roles often share the same use cases. But for now let's keep it at this.

I wondered the same thing, but struggled when it came to coming up with good naming conventions. Does ADMIN, HELP_DESK, APPROVER become like `isSuperAdmin` and STATE_REP and STATE_USER become `isStateUser`? `isNonAdmin`? I don't like any of the solutions I've come up with so far but it's worth thinking about because yeah, a lot of these scenarios are repeated.

britt-mo (approved):

gmrabian (approved):

PR #4004, opened 2022-09-22
[MDCT-1956] Dynamic Dropdown Field Functionality

juliuscaesar:
## Description
Adds functionality for dynamic values in `DropdownField`. Closes [MDCT-1956](https://qmacbis.atlassian.net/browse/MDCT-1956)

### How to test
Verify Plans populate in D3.VIII.3 Plan name under D: Sanctions.

### Changed Dependencies
N/A

## Code author checklist


britt-mo (approved):
Tested and it looked good

gmrabian (changes requested):
Maybe this is temporary/intended for now. Just checking. It errors if I don't have any plans filled out.

https://user-images.githubusercontent.com/57802560/191802702-768cb6ff-bc95-4739-9236-2c03ca92ed4e.mov

braxex (changes requested):
Looks good overall. There are some failing tests from the merge and an error to fix, but you're pretty close. 👍

gmrabian (approved):
Looks good! Works locally. Blank when no plans, so I see no problems!

services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -24,6 +25,26 @@ export const DropdownField = ({
...
+  const { report } = useContext(ReportContext);
+
+  let dropdownOptions = [];
+  if (typeof options === "string") {
+    const dynamicOptionValues = report?.fieldData[options];
+    if (dynamicOptionValues) {
+      const fieldOptions = dynamicOptionValues.map((value: string) => ({
+        label: value,
+        value: value,
+      }));
```
braxex, line 37: When we change this to loop through objects instead I think it should look like this, using the id as the value because it is immutable, and using the mutable name as the label.
```suggestion
      const fieldOptions = dynamicOptionValues.map((option: EntityShape) => ({
        label: option.name,
        value: option.id,
      }));
```

services/ui-src/src/utils/forms/forms.ts
```diff
@@ -101,8 +102,11 @@ export const initializeDropdownFields = (fields: FormField[]) => {
     (field: FormField) => field.type === "dropdown"
   );
   dropdownFields.forEach((field: FormField) => {
-    // if first option is not already a blank default value
-    if (field?.props?.options[0].value !== "") {
+    // if first provided option is not already a blank default value
+    if (
+      typeof field?.props?.options !== "string" &&
+      field?.props?.options[0].value !== ""
+    ) {
```
braxex, line 109: Rather than do this in two different places, we should pull this into the dropdown field component instead. Makes it less likely that we'll introduce some bug if we change it in the future.

braxex (dismissed):
A couple things to look at.

gmrabian (approved):
updates look good

services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -47,6 +54,29 @@ export const DropdownField = ({
...
+          label: option.name,
+          value: option.id,
+        }));
+
+        dropdownOptions = fieldOptions;
+      }
+      dropdownOptions.splice(0, 0, {
+        label: dropdownDefaultOptionText,
+        value: "",
+      });
```
braxex, line 73: Blocking: Deleting the method in forms.ts was the right call, but on lines 70-72 we're only applying the initial "Select an option" option to dynamic dropdowns. This introduces a bug of sorts, so non-dynamic dropdowns like the admin dash state selector don't have one anymore. 

Screen Shot 2022-09-28 at 17 57 56

Suggestion: To fix, let's either A) duplicate the code for non-string dropdown options, or B) remove the current implementation on 70-72 and instead add a conditional to check once if it's necessary for either prior to returning. I have a slight preference for option B but am also open to other ideas.

services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -47,6 +54,29 @@ export const DropdownField = ({
...
+    } else {
+      dropdownOptions = options;
```
braxex, line 75: Option A:
```suggestion
    } else {
      if (options[0].value !== "") {
        options.splice(0, 0, {
          label: dropdownDefaultOptionText,
          value: "",
        });
      }
      dropdownOptions = options;
```

services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -47,6 +54,29 @@ export const DropdownField = ({
...
+    }
+    return dropdownOptions;
+  };
```
braxex, line 78: Option B:
```suggestion
    } else {
      dropdownOptions = options;
    }
    if (dropdownOptions[0].value !== "") {
      dropdownOptions.splice(0, 0, {
        label: dropdownDefaultOptionText,
        value: "",
      });
    }
    return dropdownOptions;
  };
```

braxex (changes requested):
Looking really good! There's one bug that crept in but easily solvable. Let me know what you think about the suggestions I included. After this is solved we'll be good to go.

braxex (approved):
Thanks for doing that. You can remove lines 70-73 now because that functionality is now handled on 77-82, but other than that small change request, approved!

PR #4039, opened 2022-09-23
Mdct 1878

scwambach:
## Description

Added a loading spinner to all section that require and have the functionally to facilitate a loading state.

### How to test

http://localhost:3000/mcpar

- [ ] 	load the page
- [ ] 	create a program
- [ ] 	refresh the page
- [ ] 	click edit icon, click save

http://localhost:3000/mcpar/program-information/point-of-contact

- [ ] 	fill fields and click “Save & Continue ->”

http://localhost:3000/admin

- [ ] 	fill out fields and click “Replace Current Banner”
- [ ] 	click “Delete Current Banner”

## Code author checklist


services/ui-src/src/components/forms/AdminBannerForm.tsx
```diff
@@ -33,9 +37,11 @@ export const AdminBannerForm = ({ writeAdminBanner, ...props }: Props) => {
     };
     try {
       await writeAdminBanner(newBannerData);
+      setLoading(false);
       window.scrollTo(0, 0);
     } catch (error: any) {
       setError(bannerErrors.REPLACE_BANNER_FAILED);
+      setLoading(false);
```
gmrabian, line 44: if we want to set this to false in the try and catch block, we could just put it at the end of the method

services/ui-src/src/routes/Admin/Admin.tsx
```diff
@@ -26,10 +37,13 @@ export const Admin = () => {
   }, [errorMessage]);
 
   const deleteBanner = async () => {
+    setLoading(true);
     try {
       await deleteAdminBanner();
+      setLoading(false);
     } catch (error: any) {
       setError(bannerErrors.DELETE_BANNER_FAILED);
+      setLoading(false);
```
gmrabian, line 46: Same note as above

gmrabian (commented):
This is slick as HECK. I love how it looks and it makes the usability so much better. Excellent!!!

Only notes are efficiency on setting the state for try/catch blocks (maybe I'm way off on those too, let me know).

Also for everything outside of the Modal the size of the button changes to match the spinner. I think it'd look better if it were static size, no?

services/ui-src/src/routes/Admin/Admin.tsx
```diff
@@ -134,6 +151,11 @@ const sx = {
   currentBannerFlex: {
     flexDirection: "column",
   },
+  buttonLoading: {
+    width: "13.3rem",
```
gmrabian, line 155: I'm cool if this is the best way to do this, but couldn't the width just go on the `deleteBannerButton` properties permanently?

gmrabian (approved):
Looks great. Just one question on if styling could be simplified

services/ui-src/src/routes/Admin/Admin.tsx
scwambach: absolutely, i only did this because I always try to isolate what I'm working on as much as possible from anyone else code, as to not create new probs

karla-vm (approved):
This is so cool, Scott!

services/ui-src/src/components/reports/ReportPageFooter.tsx
```diff
@@ -32,17 +33,18 @@ export const ReportPageFooter = ({
           {!form?.id || formIsDisabled ? (
             <Button
               onClick={() => navigate(nextRoute)}
-              rightIcon={<Icon icon="arrowRight" />}
+              rightIcon={loading ? <></> : <Icon icon="arrowRight" />}
```
braxex, line 36: Nit: I'd probably go with `undefined` here instead of forcing it to a fragment. So like:

```suggestion
              rightIcon={loading ? undefined : <Icon icon="arrowRight" />}
```
or
```suggestion
              rightIcon={!loading ? <Icon icon="arrowRight" /> : undefined}
```

services/ui-src/src/routes/Dashboard/Dashboard.tsx
```diff
@@ -159,6 +160,11 @@ export const Dashboard = () => {
             />
           )
         ) : (
+          <Flex alignItems="center" w="full" justifyContent="center" p="10">
```
braxex, line 163: Nit: Any more than 1 or 2 inline styles and it's usually a good sign you want to make a class or `sx` entry for readability.

braxex (commented):
hey i know this is already in but had some thoughts to share. 

when we show this to design, let's get their thoughts on `speed` and `thickness` (https://chakra-ui.com/docs/components/spinner/props). if they want something consistent across all of them, that's an area where maybe we could create some cool custom variants like we have for `Button`. 

we could even maybe get the spinner to inherit variant from the button parent if there is one, and just key off that each time. 🤔

PR #4040, opened 2022-09-23
script destroys hanging log

gmrabian:
## Description
Sometimes when destroying branches the script leaves behind an api-gateway log group (this gets automatically created when we create the gateway and is not managed by our stack).

This checks if there is a hanging group and deletes it. If there isn't one, it does nothing. Even if it tries to run the delete, it'll just error if there is no matching log group.

Note: this should only ever matter if the branch was deleted improperly and we are trying to cleanup after it. On the whole I don't see this being used much

### How to test
uhh ping me if you want to test this

## Code author checklist

### DIFF
--- a/scripts/destroy.sh
+++ b/scripts/destroy.sh
@@ -26,7 +26,7 @@ if [[ $stage =~ $protected_stage_regex ]] ; then
     """
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


karla-vm (approved):

braxex (approved):
nice change 👍

braxex (approved):

karla-vm (approved):
LGTM ✨

PR #4085, opened 2022-09-27
MDCT-1994 - Change entity shape

gmrabian:
## Description
Completes subtask [MDCT-1994](https://qmacbis.atlassian.net/browse/MDCT-1994)

Needed to change dynamicfields (specifically the shape of the entities they create) to be an array of objects instead of an array of strings so that we can better manage the data in the future for drawer sections

Note: validation for the dynamicfields now only happens when you click `Save & continue`. This can be investigated further at a later time if desired

### How to test
Dynamic field works as expected

## Code author checklist

### DIFF
--- a/services/app-api/utils/validation/schemaMap.ts
+++ b/services/app-api/utils/validation/schemaMap.ts
@@ -139,7 +139,15 @@ export const radioOptional = () => radio().notRequired();
 
 // DYNAMIC
 export const dynamic = () =>
-  array().min(1).of(mixed()).required(error.REQUIRED_GENERIC);
+  array()
+    .min(1)
+    .of(
+      object().shape({
+        id: text(),
+        name: text(),
+      })
+    )
+    .required(error.REQUIRED_GENERIC);
 export const dynamicOptional = () => dynamic().notRequired();
 
 // NESTED
--- a/services/app-api/utils/validation/schemaMap.ts
+++ b/services/app-api/utils/validation/schemaMap.ts
@@ -139,7 +139,7 @@ export const radioOptional = () => radio().notRequired();
 
 // DYNAMIC
 export const dynamic = () =>
-  array().min(1).of(text()).required(error.REQUIRED_GENERIC);
+  array().min(1).of(mixed()).required(error.REQUIRED_GENERIC);
 export const dynamicOptional = () => dynamic().notRequired();
 
 // NESTED
--- a/services/ui-src/package.json
+++ b/services/ui-src/package.json
@@ -66,6 +66,7 @@
     "pa11y-runner-htmlcs": "^2.0.1",
     "patch-package": "^6.4.7",
     "postinstall-postinstall": "^2.1.0",
+    "react-uuid": "^1.0.3",
     "serverless-cloudfront-invalidate": "^1.5.0",
     "serverless-s3-sync": "^1.14.4",
     "typescript": "^4.6.3"
--- a/services/ui-src/package.json
+++ b/services/ui-src/package.json
@@ -37,6 +37,7 @@
     "react-query": "^3.34.5",
     "react-router-dom": "6.0.2",
     "react-scripts": "^5.0.0",
+    "react-uuid": "^1.0.3",
     "sass": "^1.37.5",
     "yup": "^0.32.11"
   },
@@ -66,7 +67,6 @@
     "pa11y-runner-htmlcs": "^2.0.1",
     "patch-package": "^6.4.7",
     "postinstall-postinstall": "^2.1.0",
-    "react-uuid": "^1.0.3",
     "serverless-cloudfront-invalidate": "^1.5.0",
     "serverless-s3-sync": "^1.14.4",
     "typescript": "^4.6.3"
--- a/services/ui-src/src/components/fields/DynamicField.tsx
+++ b/services/ui-src/src/components/fields/DynamicField.tsx
@@ -35,7 +35,8 @@ export const DynamicField = ({ name, label, ...props }: Props) => {
   const appendNewRecord = () => {
     const newEntity = { id: uuid(), name: "" };
     append(newEntity);
-    setDisplayValues([...displayValues, newEntity]);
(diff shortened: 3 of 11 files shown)


services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -1,76 +1,103 @@
...
+    const { id, value } = event.target;
+    const currentEntity = displayValues.find((entity) => entity.id === id);
+    const currentEntityIndex = displayValues.indexOf(currentEntity!);
+    const newDisplayValues = [...displayValues];
+    newDisplayValues[currentEntityIndex].name = value;
+    setDisplayValues(newDisplayValues);
+  };
+
+  const appendNewRecord = () => {
+    const newEntity = { id: Date.now().toString(), name: "" };
```
braxex, line 35: Blocking: Needs to be changed to a real UUID.

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -1,76 +1,103 @@
...
+  };
+
+  const removeRecord = (index: number) => {
+    remove(index);
+    const newDisplayValues = [...displayValues];
+    newDisplayValues.splice(index, 1);
+    setDisplayValues(newDisplayValues);
+  };
+
+  // set initial display value to form state field value or hydration value
```
braxex, line 47:
```suggestion
  // set initial value to form field value or hydration value
```

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -1,76 +1,103 @@
...
-    if (fields.length === 0) {
-      append(props?.hydrate || "");
+    if (hydrationValue) {
+      setDisplayValues(hydrationValue);
+      append(hydrationValue);
+    } else {
+      appendNewRecord();
     }
-  }, []);
+  }, [props?.hydrate]); // only runs on hydrationValue fetch/update
```
braxex, line 56: Nit: Use `hydrationValue` here too.
```suggestion
  }, [hydrationValue]); 
```

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -1,76 +1,103 @@
...
+      {displayValues.map((field: EntityShape, index: number) => {
         return (
-          <Flex key={field.id} alignItems="flex-end">
-            <TextField
+          <Flex key={field.id} sx={sx.textField}>
+            <CmsdsTextField
+              id={field.id}
               name={`${name}[${index}]`}
-              label={label}
+              label={label || ""}
```
braxex, line 73: Don't necessarily need the "" here because it's a required prop.
```suggestion
              label={label}
```

services/ui-src/src/utils/validation/schemas.ts
```diff
@@ -124,7 +124,7 @@ export const radioOptional = () => radio().notRequired();
 
 // DYNAMIC
 export const dynamic = () =>
-  array().min(1).of(text()).required(error.REQUIRED_GENERIC);
+  array().min(1).of(mixed()).required(error.REQUIRED_GENERIC);
```
braxex, line 127: Blocking: We should actually update this now so the validation works. Something like this should do the trick:
```suggestion
  array().min(1).of(object().shape({
    id: text(),
    name: text(),
  })).required(error.REQUIRED_GENERIC);
```

services/app-api/utils/validation/schemaMap.ts
```diff
@@ -139,7 +139,7 @@ export const radioOptional = () => radio().notRequired();
 
 // DYNAMIC
 export const dynamic = () =>
-  array().min(1).of(text()).required(error.REQUIRED_GENERIC);
+  array().min(1).of(mixed()).required(error.REQUIRED_GENERIC);
```
braxex, line 142: Not sure how this will end up, but the backend one might need to stay like this -- depends how we decide to modify the onSubmit handler.

services/ui-src/src/components/reports/sections/StaticDrawerSection.tsx
```diff
@@ -51,13 +57,13 @@ export const StaticDrawerSection = ({ form, page, setLoading }: Props) => {
...
+  const entityRows = (entities: EntityShape[]) =>
     entities.map((entity) => (
-      <Flex key={entity} sx={sx.entityRow}>
-        <Heading as="h5">{entity}</Heading>
+      <Flex key={entity.id} sx={sx.entityRow}>
+        <Heading as="h5">{entity.name}</Heading>
         <Button
           sx={sx.enterButton}
-          onClick={() => openRowDrawer(entity)}
+          onClick={() => openRowDrawer(entity.name)}
```
braxex, line 66: Question: I would think we'd want to pass through the entity id and name here so we can tell it which one to update later in the drawer. Right?

braxex (dismissed):

braxex:
@gmrabian 

Do we still need to add in the delete modal?

Are we planning on doing a different onSubmit handler for this data? I'm still thinking that will be necessary, though it could be a part of other work.

braxex:
Next person to comment on this ticket loses.

gmrabian:
@braxex 

We do not need a different onSubmit handler - verified in stacked branch https://github.com/CMSgov/mdct-mcr/pull/4096

DeleteModal: Follow-up

services/ui-src/package.json
```diff
@@ -66,6 +66,7 @@
     "pa11y-runner-htmlcs": "^2.0.1",
     "patch-package": "^6.4.7",
     "postinstall-postinstall": "^2.1.0",
+    "react-uuid": "^1.0.3",
```
braxex, line 69: I think this should be in dependencies instead of dev dependencies.

braxex (dismissed):

gmrabian:
don't review yet - fixing one issue with error displays

braxex (approved):

juliuscaesar (approved):

britt-mo (approved):
Looked good in browser & db

PR #4125, opened 2022-09-29
[MDCT-1959] Clear nested fields after they are hidden from view

juliuscaesar:
## Description
Closes [MDCT-1959: Nested fields shouldn't persist data in form state or save to database](https://qmacbis.atlassian.net/browse/MDCT-1959)
Also fixes what I assume were some incorrect validation types and removes a comma-separated mask on a textarea.
Note: this doesn't really work for fields that are already persisted to the database. Don't worry, it's fine. We'll address that when we do autosave 😬 

### How to test
* After filling out a nested field, unselect (for a checkbox) or change the parent value (radio) to hide the nested field. Then reselect the original parent value and ensure the nested field does not display what you entered previously.
* For textarea nested under checkbox: Section B, Encounter Data Report
* For textarea nested under radio: Section B, Program Integrity
* For radio nested under checkbox: Section B, Encounter Data Report
* For number nested under radio: Section D, Program Integrity
* For date nested under radio: Section D, Quality Measures

### Changed Dependencies
N/A

## Code author checklist


gmrabian (commented):
Clears things out and I don't see them in the database! However I'm also seeing an issue with saving to the database in general from the program integrity page. I see it on main too, so likely unrelated, but I can't be sure this isn't saving to the db until I know what's causing that

ntsummers1:
Screen Shot 2022-09-29 at 12 51 24 PM

Looks nice! Can confirm it is working.

One thing though, and maybe this is just me, but should instead of just setting the value to an empty string, should we delete the field altogether? So for instance, should the field I highlighted in the attached picture be instead removed from the form? That would keep consistency with how other fields are handled across the app if I'm not mistaken

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -59,6 +59,16 @@ export const ChoiceListField = ({
 
     // update DOM choices checked status
     choices.forEach((choice: FieldChoice) => {
+      // if a choice is not selected and there are children, clear out any saved data
+      if (!choice.checked && choice.children) {
```
braxex, line 63: Note: This works great as long as we only have checkedChildren. If we eventually have uncheckedChildren, we'll have to modify. Just a note for future us.

braxex (approved):
All looks good to me. 

@ntsummers1 I agree it's a little odd that the fields remaining in the database, but we don't currently have a method for deleting individual fields from the database. Once they're set, they're set. They don't hurt anything by being there so I figure we're good.

gmrabian (changes requested):
Seeing an issue where nested fields aren't rehydrating

https://user-images.githubusercontent.com/57802560/193131088-b79ba916-31b8-4d8b-a205-5f58360c79ee.mov

gmrabian (changes requested):
I typed in a nested text field, changed my outermost radio to no (so all nested fields didn't show) but the nested text value still save and shows up if I open that one again
saved like this: 
Screen Shot 2022-09-29 at 2 20 04 PM

come back to page and click yes, yes
Screen Shot 2022-09-29 at 2 20 08 PM

db has value as we
Screen Shot 2022-09-29 at 2 21 08 PM
ll

gmrabian (approved):
😌

karla-vm (approved):

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -83,6 +80,28 @@ export const ChoiceListField = ({
     });
   };
 
+  const clearNestedValues = (choices: FieldChoice[]) => {
+    return choices.forEach((choice: FieldChoice) => {
```
braxex, line 84: Suggestion: Because we aren't using the output value anywhere, there's no real need to return anything here.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -83,6 +80,28 @@ export const ChoiceListField = ({
...
+                choice.checked = false;
+              });
+              clearNestedValues(child.props.choices);
+            }
+          } else {
+            form.setValue(child.id, "", { shouldValidate: true });
+          }
+        });
+      }
+      setCheckedOrUnchecked(choice);
```
braxex, line 101: Thought: Since this action isn't really related to the function, it could live outside the function to keep things easier to reason about and ensure that the function has only one purpose. If it needs to be invoked on the nested child fields, though, that would be another thing. I don't think it needs to, though because they will get it run on them when they are rendered in their own ChoiceListField instance.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -83,6 +80,28 @@ export const ChoiceListField = ({
     });
   };
 
+  const clearNestedValues = (choices: FieldChoice[]) => {
+    return choices.forEach((choice: FieldChoice) => {
+      // if a choice is not selected and there are children, clear out any saved data
+      if (!choice.checked && choice.children) {
+        choice.children.map((child) => {
```
braxex, line 87: Nit: `.forEach` is maybe a better choice here because we don't need the output array that it is creating. `.forEach` is a bit cleaner in that way and this may be contributing to CC's "cognitive complexity score" or whatever.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -83,6 +80,28 @@ export const ChoiceListField = ({
...
+            if (child.props?.choices) {
+              child.props.choices.forEach((choice: FieldChoice) => {
+                choice.checked = false;
+              });
+              clearNestedValues(child.props.choices);
+            }
+          } else {
+            form.setValue(child.id, "", { shouldValidate: true });
+          }
+        });
```
braxex, line 99: This might be a good use case for a switch statement so it reads a bit easier. We also want to support more nested field types in the future (dropdown, single checkbox -- Jira ticket here: https://qmacbis.atlassian.net/browse/MDCT-2005) so this might be a step in that direction.

braxex (commented):
Haven't had a chance to check out the functionality myself, but leaving some thoughts on the method because of the CodeClimate cognitive complexity warning.

braxex (approved):

gmrabian (approved):

PR #4128, opened 2022-09-29
Fix radio buttons not validating correctly in main

ntsummers1:
## Description
It was reported that sometimes the form would no longer be able to submit and wouldn't save a users answers. Turns out it was the validation for radio buttons getting reset back to the old ways. This should fix that!

### How to test
./dev local
Create a program
Fill out a form. You should be able to fill out each page individually. Of particular note are the radio buttons so please make sure that those submit correctly as well.

## Code author checklist


braxex (approved):
Looks great. Thank you @ntsummers1 @gmrabian for the quick resolution.

PR #4173, opened 2022-10-01
alternate banner deletion display bug solution

braxex:
heyo - 

so i was reviewing your PR and couldn't figure out why it worked at first, so i was playing around with it trying to break things, etc. and kinda stumbled across this other potential solution. 

so like you, i assumed the reason for the flicker was a mismatch between the form state and the button submission/deletion state. and it kinda is. but even with that fix, i was still seeing the button show up (albeit with the spinner correctly showing) while it disappeared. i was puzzling over that when it hit me -- the real problem behind the flicker isn't necessarily the submission/deletion state, it's that the banner gets deleted before the button disappears. that's because the button hide is inside a collapse that has a .3s delay on it (for ✨ animation ✨ ). so the banner gets deleted, then the spinner state is set back to false, and at the same time the collapse animation starts. but before the animation can end, the flicker happens because even though the button shouldn't be shown, it still is being shown and it's set back to default state.

i figured there must be a simpler way to convince it to not show even though it's inside that collapse, so i figured i'd play around with it a bit and here's what i came up with. the fix for the whole thing ended up being 2 lines (well, 3 i guess).

1. set loading state to false at end of deletion call (`AdminPage line 37`)
2. wrap the current banner display and button in the same display logic as the collapse (`AdminPage line 69 (and 80 i guess)`)

thoughts?

### DIFF
--- a/services/ui-src/src/components/forms/AdminBannerForm.tsx
+++ b/services/ui-src/src/components/forms/AdminBannerForm.tsx
@@ -11,11 +11,7 @@ import { FormJson } from "types";
 // data
 import formJson from "forms/addAdminBanner/addAdminBanner.json";
 
-export const AdminBannerForm = ({
-  writeAdminBanner,
-  reset,
-  ...props
-}: Props) => {
+export const AdminBannerForm = ({ writeAdminBanner, ...props }: Props) => {
   const [error, setError] = useState<string>();
   const [submitting, setSubmitting] = useState<boolean>(false);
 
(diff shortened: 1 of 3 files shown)


braxex:
edit: deleted an overkill banner test that was failing. could have just fixed it, but it was truly overkill. 🤷

scwambach (approved):

ntsummers1 (approved):

PR #4188, opened 2022-10-03
MDCT-1995: Add Delete Modal to DynamicFields

ntsummers1:
## Description
Fixes [MDCT-1995](https://qmacbis.atlassian.net/browse/MDCT-1995)

Also of note, this makes it so dynamic fields can't be deleted so there are no more fields on the screen. The last field will simply be cleaned of input and be ready for new input again.

https://user-images.githubusercontent.com/19439679/193655898-97de3978-26be-4e29-9b6f-28df852476b4.mov

### How to test
./dev local
Create a program
Go to the Add a Plan page or the BSS Entities page
Add Plans/BSS Entities
Delete them
Save the page and comeback to make sure they are all still there
Delete another/add another
Save
Come back and confirm changes

## Code author checklist


karla-vm (approved):

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -2,19 +2,34 @@ import { useEffect, useState } from "react";
...
-import { Box, Button, Flex, Image } from "@chakra-ui/react";
+import { Box, Button, Flex, Image, useDisclosure } from "@chakra-ui/react";
 import { TextField as CmsdsTextField } from "@cmsgov/design-system";
 import { svgFilters } from "styles/theme";
 // utils
-import { EntityShape, InputChangeEvent } from "types";
+import { EntityShape, EntityType, InputChangeEvent } from "types";
 // assets
 import cancelIcon from "assets/icons/icon_cancel_x_circle.png";
+import { DeleteDynamicFieldModal } from "components";
```
gmrabian, line 12: nit: move this to under the components comment

services/ui-src/src/components/modals/DeleteDynamicFieldModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+    setDeleting(false);
+    modalDisclosure.onClose();
+  };
+
+  return (
+    <Modal
+      onConfirmHandler={deleteProgramHandler}
+      modalDisclosure={modalDisclosure}
+      content={{
+        heading: `Delete ${convertToReadableEntityName(entityType)}`,
```
gmrabian, line 37: Wondering if this should display the entity name instead of it's general class

services/ui-src/src/components/modals/DeleteDynamicFieldModal.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+  });
+
+  test("DeleteDynamicFieldModal bottom cancel button can be clicked", () => {
+    fireEvent.click(screen.getByText("Cancel"));
+    expect(mockCloseHandler).toHaveBeenCalledTimes(1);
+  });
+
+  test("DeleteDynamicFieldModal delete plan button can be clicked", async () => {
+    fireEvent.click(screen.getByText("Yes, delete Plan"));
+    await expect(mockCloseHandler).toHaveBeenCalledTimes(1);
```
gmrabian, line 54: could add a check here for mock remove record 🤷

gmrabian (approved):
Looking good! I don't mind if it goes in as is. Just some suggestions below:
Most importantly: I wonder if we want to indicate which entity you are deleting in the modal, just to be clear. Does design have any input on this?

services/ui-src/src/components/modals/DeleteDynamicFieldModal.tsx
ntsummers1: I gave that a thought too but stayed with the requirements that design and the ticket gave which is why its Plan or BSS Entity instead of the specific name

services/ui-src/src/components/modals/DeleteDynamicFieldModal.test.tsx
ntsummers1: Updated!

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -80,7 +104,8 @@ export const DynamicField = ({ name, label, ...props }: Props) => {
             />
             <Box sx={sx.removeBox}>
               <button
-                onClick={() => removeRecord(index)}
+                type="button"
+                onClick={() => openDeleteProgramModal(index)}
```
braxex, line 108: Question: Could we pass the entire field object through here and just set it as the selected one directly?

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -104,12 +129,21 @@ export const DynamicField = ({ name, label, ...props }: Props) => {
           Add a row
         </Button>
       )}
+      <DeleteDynamicFieldModal
```
braxex, line 132: Nit: The name of this component makes it sound like it deletes the entire DynamicField instead of just one record of the DynamicField. Not sure how I'd change it though, so nbd.

services/ui-src/src/components/modals/DeleteDynamicFieldModal.tsx
```diff
@@ -0,0 +1,64 @@
+// components
+import { Text } from "@chakra-ui/react";
+import { Spinner } from "@cmsgov/design-system";
+import { Modal } from "components";
+import { useState } from "react";
```
braxex, line 5: Info: Would usually import this first since it's a top-level react thing, but it's also nbd.

services/ui-src/src/components/modals/DeleteDynamicFieldModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+  modalDisclosure,
+}: Props) => {
+  const [deleting, setDeleting] = useState<boolean>(false);
+  const convertToReadableEntityName = (name: any) => {
+    const fieldTypeMap = {
+      plans: "Plan",
+      bssEntities: "BSS Entity",
+    };
+    const fieldType: keyof typeof fieldTypeMap = name;
+    return fieldTypeMap[fieldType];
```
braxex, line 22: Info: If you specify that `fieldTypeMap` has type `:any`, then you can save yourself the keyof typeof hassle.

```suggestion
    const fieldTypeMap: any = {
      plans: "Plan",
      bssEntities: "BSS Entity",
    };
    return fieldTypeMap[name];
```

braxex: Thought: In fact, you could simplify things a bit in this file if you just declare the `fieldTypeMap` outside on like line 8, delete this function, then just reference the fieldTypeMap wherever you need it, like: `${fieldTypeMap[name]}`. More on this below.

services/ui-src/src/components/modals/DeleteDynamicFieldModal.tsx
braxex: Thought: Following on my comment above about simplifying. If All of these lines that have some variant of: `${convertToReadableEntityName(entityType)}` could just reference `${fieldTypeMap[name]}` instead of that function.

braxex (approved):

braxex:
@ntsummers1 Looks great! I left a few non-blocking comments/suggestions that might be helpful to incorporate before merging, at your discretion.

services/ui-src/src/components/fields/DynamicField.tsx
ntsummers1: We could, though we really just need the index in this case. If we needed the name of the program then definitely we should pass it through

services/ui-src/src/components/fields/DynamicField.tsx
ntsummers1: Hows DeleteDynamicFieldRecordModal? God its wordy but it tells it like it is

services/ui-src/src/components/modals/DeleteDynamicFieldModal.tsx
ntsummers1: Trimmed down!

services/ui-src/src/components/fields/DynamicField.tsx
braxex: lolol that it does. works for me 🤷

gmrabian (approved):
I [simplified the mapping](https://github.com/CMSgov/mdct-mcr/pull/4188/commits/052347ac4a97eaa6062eeeee3315fdb9f0574b73). Let me know what you think! Otherwise I think it's ready!

PR #4221, opened 2022-10-04
MDCT-1980: Edit Measure Details

karla-vm:
## Description

Closes [MDCT-1980](https://qmacbis.atlassian.net/browse/MDCT-1980).

Adding functionality for the "Edit measure" button in **Section C: Access Measures**

### How to test

Run locally!

* Navigate to `/mcpar/program-level-indicators/availability-and-accessibility/access-measures`
* **Add access measure**
* **Edit measure**

Title should say _Edit access measure_, previously-entered data should be hydrating correctly, and when you edit the details, they should be reflecting on the original card.

### Changed Dependencies

N/A

## Code author checklist


services/ui-src/src/components/modals/DeleteEntityModal.tsx
```diff
@@ -57,7 +57,7 @@ export const DeleteEntityModal = ({
 
 interface Props {
   entityType: string;
-  selectedEntity: AnyObject;
+  selectedEntity: EntityShape | undefined;
```
gmrabian, line 60: would this be better if it were optional rather than allowed to be undefined? Maybe it all comes out the same

gmrabian (approved):

services/ui-src/src/components/modals/DeleteEntityModal.tsx
karla-vm: yup! this is a good catch, I missed it on the merge from Nick's changes.

services/ui-src/src/components/modals/AddEditEntityModal.test.tsx
```diff
@@ -100,12 +96,14 @@ describe("Test AddEditEntityModal", () => {
 
 describe("Test AddEditEntityModal functionality", () => {
   afterEach(() => {
+    mockUpdateCallPayload.fieldData.accessMeasures.splice(-1);
```
braxex, line 99: Suggestion: What if we did something like this, so it's more clear what we're up to.
```suggestion
    // reset payload to baseline with only mockEntity
    mockUpdateCallPayload.fieldData.accessMeasures = [mockEntity];
```

braxex (approved):
A few comments/suggestions but nothing blocking. Great work! 👍

PR #4264, opened 2022-10-06
redline section A

britt-mo:
## Description
Updated per redline spec for section A

### How to test
Go to 'Point of Contact' and 'Add BSS Entities' sections, see that they both have the spreadsheet widget w/ new text and formatting.

## Code author checklist

### DIFF
--- a/services/ui-src/src/forms/mcpar/mcpar.json
+++ b/services/ui-src/src/forms/mcpar/mcpar.json
@@ -36,7 +36,7 @@
                 "validation": "text",
                 "props": {
                   "label": "A.2a Contact name",
-                  "hint": "First and last name of the contact person. <br/> States that do not wish to list a specific individual on the report are encouraged to use a department or program-wide email address that will allow anyone with questions to quickly reach someone who can provide answers"
+                  "hint": "First and last name of the contact person. <br/> States that do not wish to list a specific individual on the report are encouraged to use a department or program-wide email address that will allow anyone with questions to quickly reach someone who can provide answers."
                 }
               },
               {
--- a/services/ui-src/src/forms/mcpar/mcpar.json
+++ b/services/ui-src/src/forms/mcpar/mcpar.json
@@ -87,7 +87,8 @@
           "pageType": "standard",
           "intro": {
             "section": "Section A: Program Information",
-            "subsection": "Reporting Period"
+            "subsection": "Reporting Period",
+            "spreadsheet": "A_Program_Info"
           },
           "form": {
             "id": "arp",
@@ -132,7 +133,8 @@
           "intro": {
             "section": "Section A: Program Information",
             "subsection": "Add plans (A.7)",
-            "info": "Enter the name of each plan that participates in the program for which the state is reporting data."
+            "info": "Enter the name of each plan that participates in the program for which the state is reporting data.",
+            "spreadsheet": "A_Program_Info"
           },
           "form": {
             "id": "aap",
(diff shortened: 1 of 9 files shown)


karla-vm (approved):

scwambach (approved):

services/ui-src/src/forms/mcpar/mcpar.json
```diff
@@ -155,10 +155,25 @@
...
             "subsection": "Add BSS entities (A.8)",
+            "spreadsheet": "A_Program_Info",
             "info": [
               {
-                "type": "text",
-                "content": "Enter the names of Beneficiary Support System (BSS) entities that support enrollees in the program for which the state is reporting data."
+                "type": "html",
+                "as": "span",
+                "content": "Enter the names of Beneficiary Support System (BSS) entities that support enrollees in the program for which the state is reporting data. Learn more about BSS entities at "
+              },
```
braxex, line 164: Info: Because `info:` is an array, all of this will be parsed by the [custom HTML parser](https://github.com/CMSgov/mdct-mcr/blob/3671491dfa78d091e1e32883c2a5fdbd49d9aa69/services/ui-src/src/utils/other/parsing.ts#L10-L46) we built.

There are two primary ways to interact with that parser:
1. You can specify `type: "html"` to tell it to check `content:` and parse whatever HTML is in there. But you cannot specify `as:` if you do this because the `as` will be disregarded in favor of whatever HTML you've written in the content. 
2. You can specify something else as `type:`, including any valid html tag or one of our custom types (heading, text, externalLink, internalLink) and they will be handled accordingly.

Because you're specifying both `type: "html"` and `as:` here, the `as` is being disregarded and instead it's just parsing the string you pass in `content:`, which means it renders as text loose inside a div, which can cause accessibility problems.

Screen Shot 2022-10-06 at 17 16 40

`as` can only really effectively be used with `heading` to specify what kind of heading (h1, h2, etc.) you want.

In this case, you want to do either: 

```suggestion
                "type": "html",
                "content": "<span>Enter the names of Beneficiary Support System (BSS) entities that support enrollees in the program for which the state is reporting data. Learn more about BSS entities at </span>"
              },
```

OR

```suggestion
                "type": "span",
                "content": "Enter the names of Beneficiary Support System (BSS) entities that support enrollees in the program for which the state is reporting data. Learn more about BSS entities at "
              },
```

services/ui-src/src/forms/mcpar/mcpar.json
```diff
@@ -155,10 +155,25 @@
...
+                "props": {
+                  "href": "https://www.ecfr.gov/current/title-42/chapter-IV/subchapter-C/part-438/subpart-B/section-438.71",
+                  "target": "_blank"
+                }
+              },
+              {
+                "type": "html",
+                "as": "span",
+                "content": " See Glossary in Excel Workbook for the definition of BSS entities.<br/><br/>Examples of BSS entity types include a: State or Local Government Entity, Ombudsman Program, State Health Insurance Program (SHIP), Aging and Disability Resource Network (ADRN), Center for Indepedent Living (CIL), Legal Assistance Organization, Community-based Organization, Subcontractor, Enrollment Broker, Consultant, or Academic/Research Organization."
               }
```
braxex, line 177: Nit: You'll want a period at the beginning of this, so there's a period after the link. Also they'll want these to be paragraphs. Also the same note about `type` and `as` as shown above.
```suggestion
              {
                "type": "html",
                "content": "<span>. See Glossary in Excel Workbook for the definition of BSS entities.</span><p>Examples of BSS entity types include a: State or Local Government Entity, Ombudsman Program, State Health Insurance Program (SHIP), Aging and Disability Resource Network (ADRN), Center for Indepedent Living (CIL), Legal Assistance Organization, Community-based Organization, Subcontractor, Enrollment Broker, Consultant, or Academic/Research Organization.</p>"
              }
```

braxex (changes requested):
Just a note or two to check out. Let me know if you have any questions.

services/ui-src/src/forms/mcpar/mcpar.json
britt-mo: I tried both of those options but the text turned gray in both scenarios. I just got rid of the "as" key, is that also viable?

braxex (approved):

braxex (approved):

PR #4268, opened 2022-10-07
fix nested field validation bug

braxex:
## Description

Well this was a fun one. 2 weeks ago we inadvertently introduced a bug to the codebase that broke nested field validation. Nested field validation depends on some code in `schemas.ts` that checks if the parent option is checked, and if so, creates validation schema for the nested field. This code depended on the field value matching an expected passed value that we set in `mcpar.json`. When we changed ChoiceList components (radio & checkbox) to store data in a `{name: "", value: ""}` object, it broke that method because the value of a given field was no longer a string, but an object. The fix for this was technically pretty easy -- just change that method to look deeper. But because it was now looking in an array of objects, a bit of tweaking was needed.

As I was fixing it, I realized that we actually don't even need to rely on the `visibleOptionValue` (e.g. "Other, specify") but could instead rely on something else: the `parentOptionId` (e.g. "option5"). A couple tweaks to a couple other methods later... where we used to have to specify the option value separately in `mcpar.json` like this...

```
{
    "name": "option7",
    "label": "Other, specify",
    "children": [
        {
            "id": "bssEntity_entityRole-otherText",
            "type": "textarea",
            "validation": {
                "type": "text",
                "nested": true,
                "parentFieldName": "bssEntity_entityRole",
                "visibleOptionValue": "Other, specify"
            }
        }
    ]
}
```

... we can now omit the `visibleOptionValue` (as shown below), which is now `parentOptionId` and is programmatically set behind the scenes, making `mcpar.json` management just a bit easier.

```
{
    "name": "option7",
    "label": "Other, specify",
    "children": [
        {
            "id": "bssEntity_entityRole-otherText",
            "type": "textarea",
            "validation": {
                "type": "text",
                "nested": true,
                "parentFieldName": "bssEntity_entityRole"
            }
        }
    ]
}
```

Resolves: https://qmacbis.atlassian.net/browse/MDCT-2019
Bug introduced in: https://github.com/CMSgov/mdct-mcr/pull/3959

### How to test

1. Create a new report.
2. Navigate to pages with radios and/or checkboxes and find nested fields. Ensure that they have functioning field-level validation.

### Changed Dependencies

n/a

## Code author checklist

### DIFF
--- a/services/app-api/utils/validation/schemaMap.ts
+++ b/services/app-api/utils/validation/schemaMap.ts
@@ -150,7 +150,7 @@ export const dynamicOptional = () => dynamic().notRequired();
 export const nested = (
   fieldSchema: Function | any,
   parentFieldName: string,
-  parentOptionValue: any
+  parentOptionName: any
 ) => {
   const fieldTypeMap = {
     array: array(),
@@ -162,7 +162,9 @@ export const nested = (
   const baseSchema: any = fieldTypeMap[fieldType];
 
   return baseSchema.when(parentFieldName, {
-    is: (value: any) => value && value.indexOf(parentOptionValue) != -1,
+    is: (value: any[]) =>
+      // look for parentOptionName in checked choices
+      value?.find((option: any) => option.key === parentOptionName),
     then: () => fieldSchema(),
   });
 };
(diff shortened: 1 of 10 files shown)


braxex:
Note: I'm not concerned about the level of test coverage.

ntsummers1 (approved):

PR #4336, opened 2022-10-11
Customize entity card per type

gmrabian:
## Description
Abstract the logic for entity cards so it's a little easier to create a display that matches each type

### How to test
Access measures entity cards work as normal

## Code author checklist


services/ui-src/src/types/index.ts
```diff
@@ -169,6 +169,11 @@ export declare type EntityType =
   | "qualityMeasures"
   | "sanctions";
 
+export enum ModalDrawerEntityTypes {
+  ACCESS_MEASURES = "accessMeasures",
+  QUALITY_MEASURES = "qualityMeasures",
+  SANCTIONS = "sanctions",
+}
```
braxex, line 176: Suggestion: Can probably combine these with the `EntityType` type above or something. Like maybe we make a DrawerEntityType that's just `plans` and `bssEntities` and then EntityType is the union of Drawer and ModalDrawer.

services/ui-src/src/components/index.ts
```diff
@@ -19,7 +19,9 @@ export { Banner } from "./banners/Banner";
 export { PreviewBanner } from "./banners/PreviewBanner";
 // cards
 export { Card } from "./cards/Card";
-export { EntityCard } from "./cards/EntityCard";
+export { EntityCard } from "./cards/EntityCard/EntityCard";
+export { EntityCardTopText } from "./cards/EntityCard/EntityCardTopText";
+export { EntityCardBottomText } from "./cards/EntityCard/EntityCardBottomText";
```
braxex, line 24: Suggestion: Maybe `Section` instead of `Text`? Idk you know I'm bad at naming things.

services/ui-src/src/components/cards/EntityCard/EntityCardBottomText.tsx
```diff
@@ -0,0 +1,96 @@
...
+              <Text sx={sx.subtext}>{formattedEntityData.methodFrequency}</Text>
+            </>
+          ) : (
+            <Text sx={sx.unfinishedMessage}>
+              Complete the remaining indicators for this access measure by
+              entering details.
+            </Text>
+          )}
+        </>
+      );
```
braxex, line 45: Nit: Don't need the fragment wrapper here.

services/ui-src/src/components/cards/EntityCard/EntityCardBottomText.tsx
```diff
@@ -0,0 +1,96 @@
...
+export const EntityCardBottomText = ({
+  entityType,
+  entityCompleted,
+  formattedEntityData,
+}: Props) => {
+  switch (entityType) {
+    case ModalDrawerEntityTypes.ACCESS_MEASURES:
+      return (
+        <>
+          {entityCompleted ? (
```
braxex, line 15: Thought: There's an argument to be made that this component should be dumb like `EntityCardTopText`. But that would require us to keep the unfinished message outside of it. But since every card has that and we could just map it through the json, that might be the way to go here.

services/ui-src/src/components/cards/EntityCard/EntityCardBottomText.tsx
gmrabian: @braxex I tried not having them and it doesn't like it. Which ones could I remove?

gmrabian: Note: I changed this to make it dumb like you suggested below so maybe this doesn't apply anymore

services/ui-src/src/components/cards/EntityCard/EntityCardBottomText.tsx
gmrabian: Good call - moved out

services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <>
+          <Text sx={sx.description}>Sanctions TODO</Text>
+        </>
+      );
+    case ModalDrawerEntityTypes.QUALITY_MEASURES:
+      return (
+        <>
+          <Text sx={sx.description}>Quality Measures TODO</Text>
+        </>
+      );
```
ntsummers1, line 32: Nit: Maybe I'm wrong, but since these are both one liners can they return just the Text component?

`return <Text sx={sx.description}>Sanctions TODO</Text>`

ntsummers1 (approved):

services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.tsx
gmrabian: good callout! Fixed!

services/ui-src/src/types/index.ts
braxex: 👍

services/ui-src/src/components/cards/EntityCard/EntityCardBottomText.tsx
braxex: Yeah don't sweat it.

braxex:
@gmrabian Looks great. 🚢 it

PR #4416, opened 2022-10-13
fix cross-form invalid data bug

braxex:
## Description

Resolves a nasty bug that we knew about, but which was made much, much worse by getting server-side validation in place. 

[Familiarize yourself with the bug](https://qmacbis.atlassian.net/browse/MDCT-1902), then let me tell you a story.

Basically the bug was that you could enter invalid data (say an incomplete email address in the `apoc` form), then go to another page via the sidebar and enter data there, the invalid data from the previous form page would be saved along with the valid data from the current form page due to the nature of the react-hook-form form state. 

But all that changed when we implemented server-side validation.

The good news: 
As you'd imagine, with server-side validation in place, it doesn’t save the invalid data!

The bad news: It doesn’t save any data, so if you have a field that doesn’t pass validation anywhere in any form, users can’t actually save any data to the database. And there is no user feedback that the server-side validation is failing and blocking all data from being saved in the database. And the form remembers the state of the fields, so it will continue happening until the user finds the field that’s invalid, and that’s assuming that they know something is wrong. So a user could conceivably just fill out fields straight-through the app, thinking they’re saving data and it would look like they were, but on next login, they'd discover that potentially hours of work hadn't been saved.

The great news: 
It's not a terribly involved fix and we found it pre-deploy. 🎉 

### How to test

Let's be safe and say that you should check all current submission.
1. Enter something invalid and navigate away.
2. Enter valid fill in other forms (StandardReportPage, DrawerReportPage, ModalDrawerReportPage (both AddEditEntityModal) and drawer)), and see that the values correctly make it to the database and appropriately hydrate on return.

### Changed Dependencies

n/a

## Code author checklist
- [] I have added [thorough](https://bit.ly/3zPrxuZ) tests

### DIFF
--- a/services/ui-src/src/components/modals/AddEditEntityModal.tsx
+++ b/services/ui-src/src/components/modals/AddEditEntityModal.tsx
@@ -6,7 +6,7 @@ import { Text } from "@chakra-ui/react";
 import { Spinner } from "@cmsgov/design-system";
 // utils
 import { AnyObject, EntityShape, FormJson, ReportStatus } from "types";
-import { useUser } from "utils";
+import { filterFormData, useUser } from "utils";
 
 export const AddEditEntityModal = ({
   entityType,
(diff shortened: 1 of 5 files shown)


scwambach (approved):
lgtm, works as expected

britt-mo (approved):
Looked good locally

karla-vm (approved):

ntsummers1 (approved):

PR #4447, opened 2022-10-17
new branch without yarn lock change

britt-mo:
### Description

Integration test for state user creating, editing, and entering a program

### How to test
n/a

### Changed Dependencies
n/a

## Code author checklist
- [x ] I have performed a self-review of my code
- [x ] I have added [thorough](https://bit.ly/3zPrxuZ) tests

### DIFF
--- a/tests/cypress/tests/e2e/state-user-dashboard.spec.js
+++ b/tests/cypress/tests/e2e/state-user-dashboard.spec.js
@@ -1,22 +1,8 @@
-// element selectors
-const enterMcparOnlineHomepageButton = '[data-testid="enter-mcpar-online"]';
-const enterMcparOnlineGetStartedButton =
-  '[data-testid="second-enter-mcpar-button"]';
-const editProgramButtom = "[data-testid='edit-button']";
-const submitButton = "[data-testid='modal-submit-button']";
-const programNameListItem = "[data-testid='program-name']";
-
-// selectors for all the required fields
-const titleInput = "[name='programName']";
-const startDateInput = "[name='reportingPeriodStartDate']";
-const endDateInput = "[name='reportingPeriodEndDate']";
-const checkbox = "[name='combinedData']";
-
 beforeEach(() => {
   cy.visit("/");
   cy.authenticate("stateUser");
-  cy.findByRole("button", {name: "Enter MCPAR online"}).click();
-  cy.findAllByRole("button", {name: "Enter MCPAR online"}).click();
+  cy.findByRole("button", { name: "Enter MCPAR online" }).click();
+  cy.findAllByRole("button", { name: "Enter MCPAR online" }).click();
 });
 
 describe("state user creates a program", () => {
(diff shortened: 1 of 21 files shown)


BearHanded (approved):
👍

tests/cypress/tests/e2e/state-user-dashboard.spec.js
```diff
@@ -0,0 +1,63 @@
...
+const titleInput = "[name='programName']";
+const startDateInput = "[name='reportingPeriodStartDate']";
+const endDateInput = "[name='reportingPeriodEndDate']";
+const checkbox = "[name='combinedData']";
+
+beforeEach(() => {
+  cy.visit("/");
+  cy.authenticate("stateUser");
+  cy.get(enterMcparOnline).click();
+  cy.get(secondButton).click();
```
braxex, line 22: Nit: Especially with end-to-end tests, I try to use really descriptive element names so it's more readable if (when) something goes wrong later. For instance,

```suggestion
  cy.get(enterMcparOnlineHomepageButton).click();
  cy.get(enterMcparOnlineGetStartedButton).click();
```

tests/cypress/tests/e2e/state-user-dashboard.spec.js
```diff
@@ -0,0 +1,63 @@
+// element selectors
+const enterMcparOnline = '[data-testid="enter-mcpar-online"]';
+const secondButton = '[data-testid="second-enter-mcpar-button"]';
+const addProgram = '[data-testid="add-program"]';
+const enterProgram = '[data-testid="enter-program"]';
+const editProgramButtom = "[data-testid='edit-button']";
+const submitButton = "[data-testid='modal-submit-button']";
+const closeButton = "[data-testid='close-modal']";
+const programNameListItem = "[data-testid='program-name']";
+const leaveFormButton = "[data-testid='leave-form-button']";
```
braxex, line 10: Info: `data-testid` is generally considered the last resort for selecting elements in e2e testing. Wherever possible, it's preferred to use things like the inner text of the button, or a `label` or `role` attribute. These tests could be made more resilient by switching to one of those instead of data-testids. A good rule of thumb is that writing e2e tests should usually not require modifying the tested code.

Read more here:
- [byTestId TestingLibrary docs](https://testing-library.com/docs/queries/bytestid/)
- [Why You Should Avoid Testing React Components With Test IDs](https://betterprogramming.pub/why-you-should-avoid-testing-react-components-with-test-ids-ee50d20d37d2)
- [Making your UI tests resilient to change](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change)

I'd be happy to work through these with you and switch 'em over if you're interested. 👍

tests/cypress/tests/e2e/state-user-dashboard.spec.js
```diff
@@ -0,0 +1,63 @@
...
+    cy.get(endDateInput).should("have.value", "07/14/2026");
+    cy.get(checkbox).should("be.checked");
+
+    cy.get(closeButton).click();
+  });
+
+  it("enters the program and returns to dashboard", () => {
+    cy.get(enterProgram).last().click();
+    cy.location("pathname").should("match", /point-of-contact/);
+    cy.get(leaveFormButton).first().click();
```
braxex, line 52: Question: I think there should only be one leave form button, so we wouldn't need this `.first()`? I might be missing something though.

braxex (approved):
A few comments but nothing blocking.

tests/cypress/tests/e2e/state-user-dashboard.spec.js
britt-mo: This is just because of the nature of the test--the first ever time it runs, it needs to look at the first program name and change it and there's no problem. But every subsequent time it needs to look at a different program whose name hasn't been modified yet.

tests/cypress/tests/e2e/state-user-dashboard.spec.js
britt-mo: Definitely interested

karla-vm (approved):

tests/cypress/tests/e2e/state-user-dashboard.spec.js
britt-mo: Good catch , will remove

braxex (approved):
Looks great! Thank you!

PR #4472, opened 2022-10-19
fix filter bug by adding recursion

braxex:
## Description

A [while back](https://qmacbis.atlassian.net/browse/MDCT-1958) we added a utility `filterFormFields` to our `onSubmit` handlers to filter out any data not related to the current form. This was necessary because if a user entered invalid data on a page, then navigated to another page, the form would still remember the data from that previous page and send it along with the form payload without client-side validation blocking it. This was bad because then server-side validation would catch it and fail the entire payload silently. So we added the `filterFormFields` utility to fix that. And it worked great. Except it didn't because when it compiled the fields from the current form to check the user-entered data against, it didn't check nested fields for child fields, and so child field data was always getting stripped out, which actually broke all nested field data saving.

The problem is solved by adding recursion to the utility (and some better testing).

Special thank you to @ntsummers1 for identifying this bug.

### How to test

1. Confirm that you can have invalid data on a given form page, then navigate to another page, and are still able to submit the second page's payload to the database successfully. 
2. Confirm that you can save nested field data to the database successfully.
3. Combine the two tests above by making the second form one with nested data. Confirm that all the valid data shows up in the database as expected and that the invalid data does not.

### Changed Dependencies

n/a

## Code author checklist

### DIFF
--- a/services/ui-src/src/utils/forms/forms.test.ts
+++ b/services/ui-src/src/utils/forms/forms.test.ts
@@ -1,4 +1,15 @@
-import { formFieldFactory, hydrateFormFields, sortFormErrors } from "./forms";
+import {
+  filterFormData,
+  flattenFormFields,
+  formFieldFactory,
+  hydrateFormFields,
+  sortFormErrors,
+} from "./forms";
+import {
+  mockDrawerFormField,
+  mockFormField,
+  mockNestedFormField,
+} from "utils/testing/setupJest";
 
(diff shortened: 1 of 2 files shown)


karla-vm (approved):
Confirmed locally! LGTM 💯

gmrabian (approved):
saw the bug on main, saw it fixed here

PR #4476, opened 2022-10-19
[MDCT-1965, 1975-1978] - Quality measures

ntsummers1:
## Description
Fixes everything related to Quality Measures

### How to test
./dev local
Create a program
Go to section D
Create a quality Measure
See that it creates a quality measure for you to fill out
Double check that the modal hydrates with your answers
Open the quality measure drawer
See that theres associated data at the top and no plans at the bottom 
Close the drawer
Back track and add as many plans as you want in Section A
Go back to Quality Measures
Open the drawer
Add in your responses to the new plans at the bottom
See the status update in the entity card

Now delete a plan, see that its still ok
Add a new plan, see that its not ok
Create new measures and make sure all the possible other fields can be entered, hydrated, and reflected
Delete the new measures if you want
navigate away and come back and see things are still deleted

## Code author checklist


scwambach (commented):
if I create a program and go straight to `Quality measures` it errors out on me

https://user-images.githubusercontent.com/8299237/197052820-4fa7d589-d261-47d4-ad9c-080954ff634f.mov

braxex:
@ntsummers1 @scwambach 

The no plans error is being thrown because when the page loads it attempts to get the `formattedEntityData`, which tries to get the number of plans by referencing `plans.length` in `entities.ts` [line 127](https://github.com/CMSgov/mdct-mcr/blob/e3b08dce00c2530cbc236149a2ceb5e73484667f/services/ui-src/src/utils/reports/entities.ts#L127), and it fails because of course there are no plans yet. We can get around this by changing `plans.length` to `plans?.length`, but I've got another idea I'll leave in my review comments. 👍

services/ui-src/src/components/drawers/ReportDrawerDetails.tsx
```diff
@@ -42,7 +42,45 @@ export const ReportDrawerDetails = ({ entityType, drawerDetails }: Props) => {
...
+            D2.VII.9a and D2.VII.9b
+          </Heading>
+          <Heading as="h3" sx={sx.questionHeader}>
+            Add plan-level measure results
+          </Heading>
+          <Text sx={sx.questionDescription}>
+            Add quality & performance measure results for specific plans. Not
+            seeing a plan? Be sure to add all plans in A: Program Information -
+            Add Plans before entering measure results.
+          </Text>
```
braxex, line 81: Note: I'd like to talk to R&D about this design. It feels a bit odd to me because we're basically repeating the title instruction at the top of the header. It feels like this ought to be a header above the other information, and some of it is repetitive or could be omitted entirely. On most monitors we're gonna be pressed for vertical real estate, so I think it would be good to try to slim it down. See screenshot below with some notes/ideas.

qm-drawer-design-ideas

services/ui-src/src/components/drawers/ReportDrawerDetails.tsx
```diff
@@ -42,7 +42,45 @@ export const ReportDrawerDetails = ({ entityType, drawerDetails }: Props) => {
         </Box>
       );
     case ModalDrawerEntityTypes.QUALITY_MEASURES:
-      return <Text sx={sx.detailDescription}>Quality Measures TODO</Text>;
+      return (
+        <Box sx={sx.detailBox}>
+          <Text sx={sx.detailSubtitle}>Measure Domain</Text>
+          <Text sx={sx.detailSubtext}>{drawerDetails.domain}</Text>
+          <Grid gap="normal 6" templateColumns="33% 66%">
```
braxex, line 49: Nit: We don't use grid elsewhere for this type of thing so I'm tempted to say we should keep it the same for the sake of consistency. I'm not necessarily opposed to it, though.

services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.tsx
```diff
@@ -46,7 +46,37 @@ export const EntityCardTopSection = ({
...
     case ModalDrawerEntityTypes.QUALITY_MEASURES:
-      return <Text sx={sx.description}>Quality Measures TODO</Text>;
+      return (
+        <>
+          <Heading as="h4" sx={sx.heading}>
+            {formattedEntityData.name}
+          </Heading>
+          <Text sx={sx.subtitle}>Measure Domain</Text>
+          <Text sx={sx.subtext}>{formattedEntityData.domain}</Text>
+          <Grid gap="normal 6" templateColumns="33% 66%">
```
braxex, line 56: Request: We need a bit more space between the first and second columns, as shown in the attached screenshot. I had a similar issue in sanctions if you want some prior art for reference: [JSX](https://github.com/CMSgov/mdct-mcr/blob/e3b08dce00c2530cbc236149a2ceb5e73484667f/services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.tsx#L30-L41) | [JSS.](https://github.com/CMSgov/mdct-mcr/blob/e3b08dce00c2530cbc236149a2ceb5e73484667f/services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.tsx#L98-L100)

Screen Shot 2022-10-20 at 18 50 15

services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
```diff
@@ -33,6 +39,18 @@ export const ModalDrawerReportPage = ({ route }: Props) => {
...
+  const drawerForm = { ...drawerFormJson };
+  const formContainsFieldsToRepeat = drawerFormJson.fields.find(
+    (field: FormField) => field.repeat
+  );
+  if (formContainsFieldsToRepeat) {
+    drawerForm.fields = createRepeatedFields(
+      drawerFormJson.fields,
+      report?.fieldData
+    );
+  }
```
braxex, line 52: Note: Someday I think we'll want to move this elsewhere (a piece of FormProvider Context state? 🤔 ) but that will require some additional refactoring that is for sure out of scope right now. 👍

services/ui-src/src/forms/mcpar/mcpar.json
```diff
@@ -2651,6 +2651,7 @@
                 "id": "qualityMeasure_plan_measureResults",
                 "type": "text",
                 "validation": "text",
+                "repeat": "plans",
```
braxex, line 2654: Request: Not related to this field but I have nowhere else to put it. At design's request, Scott has changed the validation of the field with id `qualityMeasure_nqfNumber` from `number` to `text`. Might as well change it in this PR too so no one is confused during review. 🤷

services/ui-src/src/types/index.ts
```diff
@@ -331,6 +332,11 @@ export interface InputChangeEvent extends React.ChangeEvent<HTMLInputElement> {}
 
 export type { IconType } from "react-icons";
 
+export interface AccessMeasureAssociatedPlan {
+  name: string;
+  value: string;
+}
+
```
braxex, line 339: Note: We probably won't need this after the (likely) design changes to the card bottom.

services/ui-src/src/utils/reports/entities.ts
```diff
@@ -15,6 +20,56 @@ const getCheckboxValues = (entity: EntityShape | undefined, label: string) => {
...
+    ? "Cross-program rate: " +
+        entity?.qualityMeasure_crossProgramReportingRateProgramList
+    : entity?.qualityMeasure_reportingRateType?.[0]?.value;
+};
+
+const getReportingPeriod = (entity: EntityShape | undefined) => {
+  return entity?.qualityMeasure_reportingPeriod?.[0]?.value === "No"
+    ? `${entity?.qualityMeasure_reportingPeriodStartDate} - ${entity?.qualityMeasure_reportingPeriodEndDate}`
+    : entity?.qualityMeasure_reportingPeriod?.[0]?.value;
+};
```
braxex, line 35: Note: I looked long and hard at these two trying to think of a way to reliably abstract away the need for them, and I have an idea or two, but for right now it doesn't make sense to mess with it. Something to think on in the future, though. 👍

services/ui-src/src/utils/reports/entities.ts
```diff
@@ -15,6 +20,56 @@ const getCheckboxValues = (entity: EntityShape | undefined, label: string) => {
...
+    if (planName && accessMeasureValue) {
+      const formattedAccessMeasure: AccessMeasureAssociatedPlan = {
+        name: planName,
+        value: accessMeasureValue,
+      };
+      convertedPlans.push(formattedAccessMeasure);
+    }
+  }
+  return convertedPlans;
+};
```
braxex, line 71: Note: I don't think we're going to end up needing this anymore with the (likely) design changes to the card bottom. We'll render out a list of data sections, one for each plan, and display the plan name, then if there's associated data, display it. Something like this pseudocode:

```js
// for each plan in report data
reportDataPlans.forEach((plan: {id: string, name: string}) => {
    // prepare entered data here
    const enteredMeasureData = formattedEntityData.planMeasures;
    const planMeasureData = enteredMeasureData[`baselinestuff_plan_measure_${plan.id}`];
    // render a data section in the card bottom with plan name and data (if it exists)
    return (
        <DataSection>
            <Text>{plan.name}</Text>
            {planMeasureData && <Text>{planMeasureData}</Text>}
        </DataSection>
    )
})
```

services/ui-src/src/utils/reports/entities.ts
```diff
@@ -61,6 +116,17 @@ export const getFormattedEntityData = (
...
     case ModalDrawerEntityTypes.QUALITY_MEASURES:
+      entityData = {
+        domain: getRadioValue(entity, "qualityMeasure_domain"),
+        name: entity?.qualityMeasure_name,
+        nqfNumber: entity?.qualityMeasure_nqfNumber,
+        reportingRateType: getReportingRateType(entity),
+        set: getRadioValue(entity, "qualityMeasure_set"),
+        reportingPeriod: getReportingPeriod(entity),
+        description: entity?.qualityMeasure_description,
+        numberOfPlans: reportFieldData?.plans.length,
```
braxex, line 127: Note: Probably won't need this anymore.

services/ui-src/src/utils/reports/entities.ts
```diff
@@ -61,6 +116,17 @@ export const getFormattedEntityData = (
...
+        plans: entity ? getPlans(entity, reportFieldData) : undefined,
```
braxex, line 128: Note: With the aforementioned changes we'll probably simplify this to just be something like (pseudocode warning):

```js
Object.entries(entity) // to array
.filter(**down to just the ones that start with whatever_something_measure_**) // filter time
.fromEntries() || undefined // back to obj or something idk i'm tired
```

services/ui-src/src/utils/testing/setupJest.tsx
```diff
@@ -427,6 +427,116 @@ export const mockCompletedAccessMeasuresFormattedEntityData = {
   methodFrequency: "mock-oversight-method-frequency",
 };
 
+export const mockQualityMeasuresEntity = {
```
braxex, line 430: Love the new mocks in the setup file.

services/ui-src/src/utils/reports/entities.ts
braxex: Also this is where the no-plans error is.
```suggestion
        numberOfPlans: reportFieldData?.plans?.length,
```

braxex (dismissed):
I'm throwing a block on this, not because of any of the code, but because we need to talk to design about some of this stuff and the half-finished state that can be achieved by later addition of new plans.

PR #4484, opened 2022-10-20
MDCT-1960 pt. 1 - add archive function

gmrabian:
## Description
Completes the backend portion of [MDCT-1960](https://qmacbis.atlassian.net/browse/MDCT-1960)

Changes:
- The line diff looks confusing - all that I did was move the last chunk of code in the `updateReport` method to be inside an `else{}` block and then created an `archiveReport` method that fires from update if the user is an ADMIN

Reasoning:
We can rethink this later if we want, but right now the only updating an Admin (BO Rep) is allowed to do is archive/unarchive. I considered creating a new endpoint and lambda altogether, though it really should live at the same endpoint, and just have the permissions restrict what it's allowed to do. Turned out that was easier than I thought! Rearranging the order of checks in `updateReport` and adding the archive path was quite simple.

### How to test
- Create a new report on the dashboard
- Use postman to hit that new report with some archive/unarchive stuff. I've updated the [postman collection](https://qmacbis.atlassian.net/l/cp/r8FNquPT)
- View the changes in dynamodb-admin!

## Code author checklist


ntsummers1 (approved):
Works like a charm!

scwambach (approved):
works for me 👍

PR #4504, opened 2022-10-21
fix nested field validation

braxex:
## Description

Nested field validation was broken because the concatenation of parentFieldId was happening over and over again on every re-render of the form (which with react-hook-form is a lot).

Screen Shot 2022-10-20 at 20 10 22

I'm not convinced this is the best way to do this or solve it. It seems like an inappropriately naive guard to add, and I wonder if we should instead be adding the parent field id manually ourselves, or doing it at some other time, say during initial template ingestion, so it is not repeated. What are your thoughts fellow codewrites?

### How to test

Well first check out how nested field validation is broken in main. Then see how it's not broken in this branch.

### Changed Dependencies

n/a

## Code author checklist


gmrabian (changes requested):
Got this console error on pages with nested fields. Page threw error state
Screen Shot 2022-10-21 at 9 18 16 AM

gmrabian:
> I wonder if we should instead be adding the parent field id manually ourselves, or doing it at some other time, say during initial template ingestion, so it is not repeated. What are your thoughts fellow codewrites?

Curious how long it'd take to prove this out? I don't have it quite in my head what you mean so I'd need to discuss or see an example

braxex:
@gmrabian Ah, I thought that error was fishy, since I obviously test on pages with nested fields and never saw that error. It turns out there were a handful of nested field options whose identifier keys were never changed from `name` to `id`. Found and fixed.

gmrabian (approved):
Looks good. This one radio had an undesirable error message
Screen Shot 2022-10-21 at 3 56 12 PM

braxex:
@gmrabian Fully ready for re-review.

karla-vm (approved):

PR #4538, opened 2022-10-31
MDCT-2048 - cover auth ssm branch and api key unverified

gmrabian:
## Description
Subtask: [MDCT-2048](https://qmacbis.atlassian.net/browse/MDCT-2048) adds test coverage on `authorization.ts` in the backend

### How to test
run `yarn coverage` and check out the improved coverage over main!

## Code author checklist

### DIFF
--- a/services/app-api/utils/auth/authorization.ts
+++ b/services/app-api/utils/auth/authorization.ts
@@ -35,10 +35,10 @@ const loadCognitoValues = async () => {
     const userPoolClientId = await ssm
       .getParameter(userPoolClientIdParams)
       .promise();
-    if (userPoolId.Parameter?.Value && userPoolClientId.Parameter?.Value) {
-      process.env["COGNITO_USER_POOL_ID"] = userPoolId.Parameter?.Value;
+    if (userPoolId?.Parameter?.Value && userPoolClientId?.Parameter?.Value) {
+      process.env["COGNITO_USER_POOL_ID"] = userPoolId.Parameter.Value;
       process.env["COGNITO_USER_POOL_CLIENT_ID"] =
-        userPoolClientId.Parameter?.Value;
+        userPoolClientId.Parameter.Value;
       return {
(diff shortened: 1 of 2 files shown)


services/app-api/utils/auth/authorization.ts
```diff
@@ -64,7 +64,8 @@ export const isAuthorized = async (event: APIGatewayProxyEvent) => {
     try {
       payload = await verifier.verify(event.headers["x-api-key"]);
     } catch {
-      console.log("Token not valid!"); // eslint-disable-line
+      // verification failed - unauthorized
+      return false;
```
braxex, line 68: Question: I have no problem with just swallowing an error here, but per line 72, the only output of this is `!!payload`, which I'm assuming would be falsy if there's an error during the try block. So do we even need the try/catch here and if not, should we just check for headers, and if there are headers, set payload? In fact, what will `verifier.verify` do if we pass it something undefined? Maybe I'm not seeing the big picture, but could this whole thing be reduced to just like:

```js
const isAuthed = await verifier.verify(event?.headers?.["x-api-key"]);
return !!isAuthed
```

braxex (approved):
Approved with one unimportant question.

services/app-api/utils/auth/authorization.ts
braxex: If we aren't gonna do anything with the error, we could just remove line 68 and let line 72 handle the return false and just add to line 67 a note that we're consciously swallowing any errors that happen here as they will bubble up a not authorized fallback response.

gmrabian: Yep so I tested this without the catch and if you send an invalid token (in this case it was just old, not even a bad format) it throws an internal error, which returns a `502 Bad Gateway` to the user. If we catch it, it just returns `403 Unauthorized` which is more what I would expect. I did clarify the variable names, but left the try/catch logic

scwambach (approved):
lgtm

PR #4544, opened 2022-11-01
externalize missing entity message

braxex:
## Description

There was one more piece of content that needed to be externalized to the .json template for the DrawerReportPage.

### How to test

1. Create a new report. Without adding any plans, navigate to a DrawerReport Page in section D.
2. Confirm that the missing plans message shows up correctly.

### Changed Dependencies

n/a

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/reports/DrawerReportPage.tsx
+++ b/services/ui-src/src/components/reports/DrawerReportPage.tsx
@@ -1,5 +1,4 @@
 import { useContext, useState } from "react";
-import { Link as RouterLink } from "react-router-dom";
 // components
 import {
   Box,
@@ -7,8 +6,6 @@ import {
   Flex,
   Image,
   Heading,
-  Link,
-  Text,
   useDisclosure,
 } from "@chakra-ui/react";
 import {
@@ -18,7 +15,7 @@ import {
   ReportPageIntro,
 } from "components";
 // utils
-import { filterFormData, useUser } from "utils";
+import { filterFormData, parseCustomHtml, useUser } from "utils";
 import {
   AnyObject,
   EntityShape,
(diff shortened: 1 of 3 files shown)


services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -137,16 +113,11 @@ export const DrawerReportPage = ({ route }: Props) => {
...
-          {missingEntitiesMessage}{" "}
-          <Link as={RouterLink} to={missingEntitiesLink.href}>
-            {missingEntitiesLink.text}
-          </Link>
-        </Text>
-      )}
+      <Box sx={sx.missingEntityMessage}>
+        {entities?.length
+          ? entityRows(entities)
+          : parseCustomHtml(verbiage.missingEntityMessage || "")}
```
braxex, line 119: Note for reviewer: Changed this to just parse whatever custom html we pass it, similar to other text areas on form pages.

services/ui-src/src/forms/mcpar/mcpar.json
```diff
@@ -1458,7 +1458,20 @@
...
+                "content": "This program is missing plans. You won’t be able to complete this section until you’ve added all the plans that participate in this program in section A.7. "
+              },
+              {
+                "type": "internalLink",
+                "content": "Add Plans",
+                "props": {
+                  "to": "/mcpar/program-information/add-plans"
+                }
+              }
+            ]
```
braxex, line 1474: Note for reviewer: Same basic section added for each DrawerReportPage in the template.

scwambach (approved):
does the thing you said it would do!

services/ui-src/src/forms/mcpar/mcpar.json
tblackwe: previously the content was in a shared variable, now it is repeated several times. Is there a reason to not have a const missingEntityMessage?

karla-vm (approved):

services/ui-src/src/forms/mcpar/mcpar.json
braxex: yeah i considered doing that, but the reality is that we already have two types (plans (most of them) and bss entities (only 1 instance)) that are being used, and there will likely be others added in the future. because we have to do this in json, we'd have to set the plans-only verbiage somewhere in the form and then reference it, which wouldn't be all bad, but if we wanted to diverge from that on a single page we'd then be overwriting it, etc. it just seemed like more logic than it was worth to avoid repeated lines in a json file that's already 3k+. 🤷

PR #4557, opened 2022-11-01
MDCT-2024 partial - rename get started and review and submit to be mcpar specific

gmrabian:
## Description
Part of [MDCT-2024](https://qmacbis.atlassian.net/browse/MDCT-2024) - rename `GetStartedPage` and `ReviewSubmitPage` to prefix with "Mcpar" because they are specific to that report, even in their pathing. This way we can make additional getstarted/reviewsubmit pages and not have issues yay

### How to test
Stuff still works (specifically the two aforementioned pages)

## Code author checklist


karla-vm (approved):
✨

services/ui-src/src/components/index.ts
```diff
@@ -71,8 +71,8 @@ export { HelpPage } from "./pages/Help/HelpPage";
 export { HomePage } from "./pages/Home/HomePage";
 export { NotFoundPage } from "./pages/NotFound/NotFoundPage";
 export { ProfilePage } from "./pages/Profile/ProfilePage";
-export { GetStartedPage } from "./pages/GetStarted/GetStartedPage";
-export { ReviewSubmitPage } from "./pages/ReviewSubmit/ReviewSubmitPage";
+export { McparGetStartedPage } from "./pages/McparGetStarted/McparGetStartedPage";
+export { McparReviewSubmitPage } from "./pages/McparReviewSubmit/McparReviewSubmitPage";
```
braxex, line 75: Thought: Would it make sense to leave the folders as `GetStarted` and `ReviewSubmit` since there will be others?

gmrabian: That's a good call, I hadn't considered that. Let me make the switch

braxex (approved):

PR #4580, opened 2022-11-02
hiding accordion

scwambach:
## Description

removed accordion on the help page

### How to test

go take a look at the help page

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/pages/Help/HelpPage.tsx
+++ b/services/ui-src/src/components/pages/Help/HelpPage.tsx
@@ -25,7 +25,7 @@ export const HelpPage = () => {
           cardprops={sx.card}
         />
       </Box>
-      {accordionItems.length > 1 && (
+      {accordionItems.length > 0 && (
         <Box sx={sx.faqAccordionBox}>
           <FaqAccordion accordionItems={accordionItems} />
         </Box>
--- a/services/ui-src/src/components/pages/Help/HelpPage.tsx
+++ b/services/ui-src/src/components/pages/Help/HelpPage.tsx
@@ -25,9 +25,11 @@ export const HelpPage = () => {
           cardprops={sx.card}
         />
       </Box>
-      <Box sx={sx.faqAccordionBox}>
-        <FaqAccordion accordionItems={accordionItems} />
-      </Box>
+      {accordionItems.length > 1 && (
+        <Box sx={sx.faqAccordionBox}>
+          <FaqAccordion accordionItems={accordionItems} />
+        </Box>
+      )}
     </PageTemplate>
   );
 };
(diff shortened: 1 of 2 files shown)


gmrabian (approved):
Works, looks good.
If we want to remove it, is it worth leaving in the code? Not sure the reason for removing, but feel like we could just take it out of verbiage and HelpPage.tsx completely if it's not needed

scwambach:
> Works, looks good. If we want to remove it, is it worth leaving in the code? Not sure the reason for removing, but feel like we could just take it out of verbiage and HelpPage.tsx completely if it's not needed

@braxex correct me if I'm wrong, but I think that there will be new FAQ questions in the future.

services/ui-src/src/components/pages/Help/HelpPage.tsx
```diff
@@ -25,9 +25,11 @@ export const HelpPage = () => {
           cardprops={sx.card}
         />
       </Box>
-      <Box sx={sx.faqAccordionBox}>
-        <FaqAccordion accordionItems={accordionItems} />
-      </Box>
+      {accordionItems.length > 1 && (
```
braxex, line 28: Request: As-is, an accordion with only 1 item would not show, because the length is only equal to 1 and not greater than 1. What you probably want is `accordionItems.length`, which will evaluate to `true` if it is anything greater than 0.
```suggestion
      {accordionItems.length && (
```

braxex: Note: If we wanted to delete the `accordionItems` array from the verbiage file entirely right now we could do that too.

services/ui-src/src/verbiage/pages/help.ts
```diff
@@ -17,16 +17,5 @@ export default {
...
-      answer:
-        "The MCPAR online reporting portal will be available starting early November 2022.",
-    },
-    {
-      question: "How do I log into my IDM account?",
-      answer:
-        "CMCS staff will be reaching out to states directly to make sure that appropriate state staff are granted access in the very near future so that each state has access in advance of the web-based forms being ready for use. Once the web-based forms are ready, you’ll receive an announcement from CMS.",
-    },
-  ],
+  accordionItems: [],
```
braxex, line 20: Note: When deleting stuff like this, especially if I know I'll be adding some back soon, I sometimes leave some breadcrumbs for myself, like: 

```ts
//  accordion items are in the following format:
//  {
//    question: "",
//    answer: "",
//  }
```

braxex (changes requested):
Just one requested change, but otherwise looks good.

braxex (approved):

braxex:
@scwambach 🚢 it

PR #4594, opened 2022-11-02
Halp

gmrabian:
## Description
Brax must eat

## Code author checklist

### DIFF
--- a/services/ui-src/src/utils/validation/schemas.test.ts
+++ b/services/ui-src/src/utils/validation/schemas.test.ts
@@ -14,7 +14,7 @@ describe("Schemas", () => {
     "N/A",
     "Data not available",
   ];
-  const badNumberTestCases = ["abc", "N", "", "123:123", "!@#!@%"];
+  const badNumberTestCases = ["abc", "N", "", "!@#!@%"];
 
   const goodRatioTestCases = [
     "1:1",
@@ -48,12 +48,12 @@ describe("Schemas", () => {
     }
   };
 
-  test("Evalulate Number Schema using number scheme", () => {
+  test("Evaluate Number Schema using number scheme", () => {
     testNumberSchema(number(), goodNumberTestCases, true);
     testNumberSchema(number(), badNumberTestCases, false);
   });
 
-  test("Evalulate Number Schema using ratio scheme", () => {
+  test("Evaluate Number Schema using ratio scheme", () => {
     testNumberSchema(ratio(), goodRatioTestCases, true);
     testNumberSchema(ratio(), badRatioTestCases, false);
   });
--- a/services/ui-src/src/utils/validation/schemas.ts
+++ b/services/ui-src/src/utils/validation/schemas.ts
@@ -23,53 +23,6 @@ const valueCleaningNumberSchema = (value: string, charsToReplace: RegExp) => {
   });
 };
 
-// export const number = () =>
-//   mixed()
-//     .test({
-//       message: error.REQUIRED_GENERIC,
-//       test: (val) => val != “”,
-//     })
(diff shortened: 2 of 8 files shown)


scwambach (approved):

services/ui-src/src/utils/validation/schemas.test.ts
```diff
@@ -14,7 +14,7 @@ describe("Schemas", () => {
     "N/A",
     "Data not available",
   ];
-  const badNumberTestCases = ["abc", "N", "", "123:123", "!@#!@%"];
+  const badNumberTestCases = ["abc", "N", "", "!@#!@%"];
```
braxex, line 17: Question: Do we need the `""` test case here or can we remove it altogether?

services/ui-src/src/utils/validation/schemas.ts
```diff
@@ -17,30 +17,28 @@ export const textOptional = () => text().notRequired();
...
     .test({
       message: error.INVALID_NUMBER_OR_NA,
-      test: (val) => {
-        const replaceCharsRegex = /[,.]/g;
-        return (
-          ignoreCharsForSchema(val, replaceCharsRegex).isValidSync(val) ||
-          validNAValues.includes(val)
-        );
+      test: (value) => {
+        const validNumberRegex = /[0-9,.]/;
```
braxex, line 32: Info: We could replace `0-9` with `d` to be real regexy, but honestly I like 0-9 better. Why do I bring it up? Idk.

services/ui-src/src/utils/validation/schemas.ts
```diff
@@ -17,30 +17,28 @@ export const textOptional = () => text().notRequired();
...
+        if (value) {
+          const isValidStringValue = !!validNAValues.includes(value);
+          const isValidNumberValue = !!value?.match(validNumberRegex);
```
braxex, line 35: Suggestion: We can get just a boolean from a regex test by using `regex.test` instead, like this:

```suggestion
          const isValidStringValue = validNAValues.includes(value);
          const isValidNumberValue = validNumberRegex.test(value);
```

It has the added bonus of not having to use boolean coalescing.

services/ui-src/src/utils/validation/schemas.ts
```diff
@@ -17,30 +17,28 @@ export const textOptional = () => text().notRequired();
...
+          return isValidStringValue || isValidNumberValue;
+        }
+        return true;
```
braxex, line 38: Nit: This should really be `else return true`.
```suggestion
        } else return true;
```

braxex (dismissed):
We still need to copy this to the server-side validation schema. Thanks @karla-vm and @gmrabian and @ntsummers1 for all your work on this one.

services/ui-src/src/utils/validation/schemas.test.ts
gmrabian: still a bad number 🤷 we don't want to allow empty string

services/ui-src/src/utils/validation/schemas.ts
gmrabian: fair point. I like the readability of 0-9

braxex (approved):
Looking good. 

Does that masking issue happen in main? If no, I think I know the problem. Here's a hint for @gmrabian. Maybe your co-conspirator can help you puzzle it out.

>Monikers given, expressed or guessed,
Is character common? Regular, yes,
In thine own self begin and start at the first,
The answer you'll find: narrow is worst.

Let's make that change and see how the mask works afterward.

karla-vm (approved):
💯

karla-vm (approved):

PR #4595, opened 2022-11-02
some fields should be ratios

braxex:
just do it


gmrabian (changes requested):
First ratio adds the colon and lets you tab by (it does throw error on save)? Second one immediately throws. I believe there is another here (D1.X.8) that should be a ratio as well, no?


https://user-images.githubusercontent.com/57802560/199735876-f69bf544-3394-42fa-a341-0375a2e335a2.mov

braxex:
@gmrabian Ready for review.

gmrabian (approved):
works great

BearHanded (approved):
lgtm

PR #4690, opened 2022-11-04
nested field validation bug fix

braxex:
## Description

This fixes a validation error we were having where nested radio and checkbox fields were being marked as not required, but were still having a separate requirement applied to have a minimum of one item checked even if they were not active. We need both checks normally, so this was just a tweak to override the `min(1)` with `min(0)` the same way we override `required()` with `notRequired()` if it's not active. See video below to see what it was doing before.

https://user-images.githubusercontent.com/27705928/200033364-737edfc6-a344-4c33-8c77-492426ad7a5d.mov

### How to test

1. Make sure it doesn't do what it does in the video.

### Changed Dependencies

n/a

## Code author checklist


scwambach (approved):
it definitely does NOT do the thing in the video

karla-vm (approved):

gmrabian (approved):

PR #4697, opened 2022-11-04
1 -- partial 2024 (v): tweak how routes are stored and passed

braxex:
## Description

Big picture -- there are two main parts of the frontend: in-report and not-in-report. If you're inside a report, you have access to the ReportContext and your source of truth for anything and everything related to the report, include route information like is used to create the sidebar, the next/previous buttons, etc. should be coming from the ReportProvider. If you are not in the report, but need some information about reports in general (for instance, so the dashboard can know what kind of report json should be stored alongside a newly created report), you need to get that from elsewhere. 

Previously were playing it a little fast and loose with the difference between the two, but we've been planning on fixing that pre-deployment, and that's what this PR is here for.

Changes:
- The `useFindRoute` utility receives passed in props again and uses those to calculate previous/next routes. It's also been made a bit more flexible with some fallbacks so it can handle report information sometimes being unavailable.
- Flat routes, which were previously not stored in form templates, are now stored as part of form templates.
- `ReportPageFooter` and `StandardReportPage` get routing information from `ReportContext` now.
- Related test updates.

### How to test

1. Hop in that app, create a new program, and click all around.

### Changed Dependencies

n/a

## Code author checklist

### DIFF
--- a/services/app-api/utils/validation/schemaMap.ts
+++ b/services/app-api/utils/validation/schemaMap.ts
@@ -162,7 +162,7 @@ export const nested = (
       // look for parentOptionId in checked Choices
       value?.find((option: any) => option.key === parentOptionId),
     then: () => fieldSchema(), // returns standard field schema (required)
-    otherwise: () => fieldSchema().notRequired(), // returns not-required field schema
+    otherwise: () => fieldSchema().notRequired().min(0), // returns not-required field schema
   });
 };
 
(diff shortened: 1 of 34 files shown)


karla-vm (approved):

gmrabian (approved):
cool. Didn't see any problems

ntsummers1 (approved):

PR #4718, opened 2022-11-07
Cypress login

gmrabian:
## Description
Modifies our custom 'authentication' module to take advantage of [Cypress sessions](https://docs.cypress.io/api/commands/session). Sessions allow us to reuse the auth tokens from a given login, ensuring that we don't need to relogin for every test that does not explicitly want to test the login flow. Sessions only persist per `.spec.` file so a future enhancement is making these tokens count for all tests, not just per file. That said, this change alone reduces our cypress runtime by up to a minute.

BEFORE
Screen Shot 2022-11-04 at 2 06 10 PM

AFTER
Screen Shot 2022-11-07 at 11 07 12 AM

These two config changes help us take advantage of this feature. You can also view [this guide on sessions](https://www.cypress.io/blog/2021/08/04/authenticate-faster-in-tests-cy-session-command/) for a brief idea of what we went for.
```
experimentalSessionAndOrigin: true,
 testIsolation: "off",
```

To further take advantage of these features, we upgraded Cypress to version 10. See the migration guide and GitHub actions updates below.
[Cypress GitHub Action releases](https://github.com/cypress-io/github-action/releases) (v4.0 adds Cypress 10 support)
[Cypress v10 migration guide](https://docs.cypress.io/guides/references/migration-guide#Config-Option-Changes)
Key changes:
- `config.json` -> `cypress.config.js`
- ~`plugins/index.js`~ -> `cypress.config.js` (within the config for e2e tests)

Other notes on changes:
1. `cy.visit("/")` was moved to 'afterEach' statements because our authentication will land on the homepage. When we reuse a session, we need to start back at the same spot because the session will not rerun that step.
2. `cy.waitUntil` is a neat community package we picked up because it allows us to only save the session once our tokens are ready. (doing a manual while loop with `cy.wait` caused the wait to fire as many times as possible regardless of token readiness. Hence why the community made this package)

### How to test
_See changed dependencies below_

Run cypress tests locally and ensure they pass! (they pass in GitHub actions!)

### Changed Dependencies
I suggest removing your `node_modules` folder from `tests/cypress` and running `yarn` in that folder to reinstall with cypress 10

## Code author checklist

- [x] I have done the deep review and verified the items in the above checklist are g2g


tests/cypress/support/authentication.js
```diff
@@ -31,33 +31,46 @@ const adminUser = {
...
+    cy.get(cognitoLoginButton).click();
+
+    /**
+     * Waits for cognito session tokens to be set in local storage before saving session
+     * This ensures reused sessions maintain these tokens
+     * We expect at least three for the id, access, and refresh tokens
+     * We don't explicitly check for token names because they change
+     */
+    cy.waitUntil(() =>
+      cy.window().then((window) => window.localStorage.length > 3)
```
tblackwe, line 73: Does the prefix of the name change? I like the update using waitUntil, but I think the filter could be improved. Or am I being overzealous in the approach?

tests/cypress/tests/e2e/header.spec.js
```diff
@@ -9,10 +9,13 @@ const menuOptionLogOut = '[data-testid="header-menu-option-log-out"]';
 const cognitoLoginButton = "[data-testid='cognito-login-button']";
 
 beforeEach(() => {
-  cy.visit("/");
   cy.authenticate("stateUser");
 });
 
+afterEach(() => {
+  cy.visit("/");
```
tblackwe, line 16: we could speed the runtime slightly by putting a conditional visit here.
`if (cy.location("pathname") !== "/") cy.visit("/")`

gmrabian: adding this - thanks!

tests/cypress/support/authentication.js
gmrabian: The prefix should be the same. Can you think of a succinct way to check for it? I didn’t want to get into a forEach in here

gmrabian:
Dev note: I made many changes in an attempt to fix what I thought was a Cypress bug related to the failing Sidebar tests. It was actually the result of a change that got merged in from main

tblackwe (approved):

tblackwe (approved):
I've done the deep dive review. Great work Garrett, and cheers for being so receptive to feedback.

britt-mo (approved):

PR #4750, opened 2022-11-09
fix nested date validation errors

braxex:
## Description

The problem: Nested date validation was broken. 
Locations: We have nested date fields on D II. Financial performance (drawer) and D VII. Quality measures (modal).

Background: We have had some trouble with nested field validation in the past, specifically with nested child fields still being considered required even when the parent field is not checked. We had iterated on using the `fieldSchema` to do this, stripping away requirements one by one, but it makes more sense to just use the Yup base schema instead. This is because the Yup base schema is used as the baseline for our custom field schema types, to which we then add validation like `.required`, `.test`, `.min`, `.matches`, etc. So really as we were stripping away the validation additions from the `fieldSchema` we were just getting closer and closer to the Yup base schema. So now we're just going to use the Yup base schema.

Changes: 
- Updated frontend and backend validation schemas to use Yup base schema instead of pared back field schema.

NOTE: I am continuing to explore how we can better test this area of our app to prevent issues like this.

### How to test

1. Test both D II. Financial performance (drawer) and D VII. Quality measures (modal). Ensure the nested date fields submit to the database with both checked and unchecked parents. 
2. Test other nested field types:
    - Nested radio: (see B X. Program Integrity)
    - Nested text: (all over the place, see B X. Program Integrity (B.X.8a) or any "Other, specify" anywhere)
    - Nested number: (see D X. Program Integrity, drawer)

### Changed Dependencies

n/a

## Code author checklist


gmrabian (approved):
Looks good locally

tblackwe (approved):

scwambach (approved):

ntsummers1 (approved):

PR #4770, opened 2022-11-14
MDCT-2088 - Server-side validation failure should notify user

gmrabian:
## Description
[MDCT-2088](https://qmacbis.atlassian.net/browse/MDCT-2088) I have some reservations here and would like _your_ input.

With what is in this PR you do get an error banner whenever it fails. However,
1. It does not go away until you refresh the page
2. It does not tell you what happened (this is sometimes desired though)
3. It does not save the values you filled in. It's just lost and you have to re-enter. Very sad.
4. Autosave will change some or most of the above feel and behavior
5. I tried putting a condition on `StandardReportPage`'s `onSubmit` such that it won't navigate if `errorMessage` is set. This sets up a race condition between the state setter and the navigate. From what I saw the navigator won every time - meaning it'll let you navigate one page but then subsequent navigations blocks.

Wondering if we want to just log the error for support and not show it? I still think that's bad because people might not notice stuff isn't getting saved. The alternative though is a bit complicated. 

Ok thoughts time please comment.

### How to test
Comment out lines 52, 62, 73, and 84 of `services/ui-src/src/components/reports/ReportProvider.tsx`.

Add a `throw new Error("test error")` to line 31 of `services/app-api/handlers/reports/update.ts`
Try to submit a report. See the error banner and the associated issues mentioned above (Note #4 will not happen unless you add that condition)

If you navigate around you'll notice that the banner doesn't go away. 

Now uncomment lines 52, 62, 73, and 84 of `services/ui-src/src/components/reports/ReportProvider.tsx`.

If you try and submit the form, you'll see that the error banner still appears but as you navigate around the form, it will eventually get removed.

## Code author checklist


braxex:
@gmrabian Great questions. To your fourth point, I agree that autosave will change a lot of this behavior, so I'm going to pretend that autosave is in place for the purpose of my responses. 

>1. It does not go away until you refresh the page

I think that's the way it should be. However, on any successful submission, the error message should be wiped out. With autosave in place, this will mean that any subsequent valid change and submission of that field, or submission of other valid data will remove the error message.

>2. It does not tell you what happened (this is sometimes desired though)

The error message triggered by the server-side validation should read something like: "Report data is invalid and could not be saved. Please contact support." That way we will know from the error message what is happening.

>3. It does not save the values you filled in. It's just lost and you have to re-enter. Very sad.

This is also to be expected with our current implementation because we do not save invalid data to the database. With autosave in place, this will be limited to a single field at a time in most cases and at most one modal or one drawer, which will have an immediate feedback loop in place to show the user that the data was not entered.

>5. I tried putting a condition on StandardReportPage's onSubmit such that it won't navigate if errorMessage is set. This sets up a race condition between the state setter and the navigate. From what I saw the navigator won every time - meaning it'll let you navigate one page but then subsequent navigations blocks.

I wouldn't worry about this too much because with autosave in place the navigation will be completely separated from the submit event.

gmrabian:
For this to work fully it relies on the features in the `fields` branch. Remains a draft until that branch goes into autosave

braxex:
I accidentally closed and deleted this. That was bad. Now it's back.

braxex:
Closing this because we are pursuing alternatives with design.

PR #4805, opened 2022-11-16
Britt and Nicks Most Excellent MCPAR Dashboard Integration Tests

ntsummers1:
## Description

Fixes [MDCT-2074](https://qmacbis.atlassian.net/browse/MDCT-2074) and its children

### How to test

Make sure you checkout the Gherkin/Cucumber section in the Readme that Thomas wrote up! It's a great intro into how everything works and how to set it up. Once you've followed it, you can simply run (Make sure you swap passwordhere for the actual passwords!)

`CYPRESS_ADMIN_USER_PASSWORD='passwordhere' CYPRESS_STATE_USER_PASSWORD='passwordhere' yarn test`

That will bring up the cypress menu. Once everything spins up, you can click start integration test and then click all any test you'd like to see run. I'd recommend Dashboard and Form as that's what we're really adapting on this PR.

## Code author checklist


tests/cypress/support/step_definitions/formInputs.js
```diff
@@ -1,11 +1,83 @@
...
+        cy.get(`input[name^="${row[0]}"]`).eq(row[3]).type(row[2]);
+        break;
+      default:
+        cy.get(`[name='${row[0]}']`).type(row[2]);
+        break;
+    }
+  });
+});
+
+When("these form elements are edited:", (dataTable) => {
```
tblackwe, line 28: Is this step def significantly different that form elements filled?  I think it is some elements are cleared, but most are the same. Might be an opportunity for future improvement. 
I think we can simplify this by calling `.clear()` on the appropriate items, then using `Step` to call `these form elements are filled:`

see [these docs for more info](https://github.com/badeball/cypress-cucumber-preprocessor/blob/master/docs/cucumber-basics.md#nested-steps)

tests/cypress/tests/e2e/mcpar/dashboard.feature
```diff
@@ -0,0 +1,84 @@
+Feature: MCPAR Dashboard Page - Program Creation/Editing/Archiving
+
+    Scenario: State users can create programs
```
tblackwe, line 3: When you have `Examples` you should use `Scenario Outline`
I didn't even know you could get away with using Scenarios 😄

tests/cypress/tests/e2e/mcpar/dashboard.feature
```diff
@@ -0,0 +1,84 @@
...
+    Scenario: Admin users cannot create programs
+        Given I am logged in as an admin user
+        And I am on "/"
+        And these form elements are filled:
+            | state | dropdown | Minnesota |
+
+        When I click the "Go to Report Dashboard" button
+        Then there is no way to create a program
+
+    Scenario: Admin users can archive programs
```
tblackwe, line 51: same note as state user.

Background can be simplified as

> Given I am logged in as an admin user
>And I am on "/"
>And these form elements are filled:
>  | state | dropdown | Minnesota |

        When I click the "Go to Report Dashboard" button

tests/cypress/support/step_definitions/formInputs.js
britt-mo: this is a good suggestion! I'll work on it.

tests/cypress/support/step_definitions/formInputs.js
```diff
@@ -1,11 +1,83 @@
 import { When } from "@badeball/cypress-cucumber-preprocessor";
 When("these form elements are filled:", (dataTable) => {
   dataTable.rawTable.forEach((row) => {
-    cy.get(`[name='${row[0]}']`).type(row[1]);
+    switch (row[1]) {
```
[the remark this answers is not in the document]
ntsummers1: Agree! I think this would be a good future enhancement we can build on

tests/cypress/support/step_definitions/formInputs.js
ntsummers1: Updated and talked with Britt and Thomas about solution :)

tblackwe (approved):

tests/cypress/tests/e2e/mcpar/dashboard.feature
```diff
@@ -0,0 +1,91 @@
...
+            Then the program is archived
+
+    Rule: State User Logged In
+        Background: State User on MCPAR page
+            Given I am logged in as a state user
+            And I am on "/mcpar"
+
+        Scenario: State users can't see archived programs
+            Given I am on "/mcpar"
+            Then there are no active programs
```
gmrabian, line 88: Will this consistently pass? The sidebar tests and the form tests create programs, plus we create programs, so this won't always be the case, right?

tests/cypress/tests/e2e/mcpar/dashboard.stepdef.js
```diff
@@ -0,0 +1,22 @@
...
+});
+
+Then("the program is archived", () => {
+  cy.contains("Unarchive").should("be.visible");
+});
+
+Then("one program is archived and the other is unarchived", () => {
+  cy.findAllByRole("button", { name: "Archive" }).first().should("be.visible");
+  cy.findAllByRole("button", { name: "Unarchive" }).last().should("be.visible");
+});
```
gmrabian, line 18: Similar here: can we count on this if we know other tests and we as manual testers might create programs?

gmrabian (commented):
This is killer. Fun and surprisingly cathartic to watch.

I'll echo Thomas's note about putting these in the GitHub Actions, so that we can make sure they're working. 

Secondly I'm wondering if we want to test hydration at all on these 🤔 I guess it'd be tiresome to go through the entire report again just to verify everything is the same. What do you think?

ntsummers1:
> Secondly I'm wondering if we want to test hydration at all on these 🤔 I guess it'd be tiresome to go through the entire report again just to verify everything is the same. What do you think?

I totally agree that that would be slick. I think that might be a good future enhancement to add in. Maybe after confirming we've got all these initial tests sorted?

tests/cypress/tests/e2e/mcpar/dashboard.feature
ntsummers1: I've gone ahead and cleaned up the sidebar test so that it archives the program it creates. The form also archives the program it creates. That should keep it clean for this test to pass. However, if we went in there and added a program and didn't delete it then yes, this would fail. That could be ok, but I could also go a step further and make sure there are no active programs by the name of Test Program or Test Program 2 (The 2 created in the form test) to make this slightly more robust. Happy to do either or

gmrabian:
So looks like some of the errors in the tests here are because the created programs aren't showing in the dashboard. Right now our fetchAll is size limited to 1MB of matching records. This is known and is easy to fix, we just haven’t yet. So when you’re creating new programs they are showing up in the database, but the fetchAll only returns for first few so those’ll never show up.

gmrabian (approved):
most excellent!

PR #4806, opened 2022-11-16
MDCT-2118: Question Change for D3.VIII.8

karla-vm:
## Description

Closes [MDCT-2118](https://qmacbis.atlassian.net/browse/MDCT-2118).

Changing the format and verbiage of question `D3.VIII.8` under **Sanctions**.

### How to test

- Run the app locally
- Create a report
- Create a plan
- Navigate to `/mcpar/plan-level-indicators/sanctions`
- Add a sanction, check that the question is now a radio with an adorable little child
- Save it, make sure it's saved to the DB

### Changed Dependencies

N/A

## Code author checklist

### DIFF
--- a/services/ui-src/src/forms/mcpar/mcpar.json
+++ b/services/ui-src/src/forms/mcpar/mcpar.json
@@ -2983,12 +2983,36 @@
                 }
               },
               {
-                "id": "sanction_remediationDate",
-                "type": "date",
-                "validation": "date",
+                "id": "sanction_remediationCorrected",
+                "type": "radio",
+                "validation": "radio",
                 "props": {
                   "label": "D3.VIII.8 Remediation date non-compliance was corrected",
-                  "hint": "What day was it remediated, i.e. When was the non-compliance corrected? Select a date."
+                  "hint": "Was the non-compliance remediated?",
+                  "choices": [
+                    {
+                      "id": "J9FTVN9fJpgXX6X2xJiAiP",
+                      "label": "Yes",
+                      "children": [
+                        {
+                          "id": "sanction_remediationCorrectedDate",
+                          "type": "date",
+                          "validation": {
+                            "type": "date",
+                            "nested": true,
+                            "parentFieldName": "sanction_remediationCorrected"
+                          },
+                          "props": {
+                            "hint": "When was the non-compliance corrected? Select a date."
+                          }
+                        }
+                      ]
+                    },
+                    {
+                      "id": "oAcvChnL6FLUJorh8XfgyR",
+                      "label": "No"
+                    }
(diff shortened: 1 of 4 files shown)


tblackwe (approved):

services/ui-src/src/forms/mcpar/mcpar.json
```diff
@@ -2983,12 +2983,36 @@
                 }
               },
               {
-                "id": "sanction_remediationDate",
-                "type": "date",
-                "validation": "date",
+                "id": "sanction_remediationCorrected",
```
braxex, line 2986: Nit: I would maybe name this one `sanction_remediationCompleted`. Because remediation is the act of correcting something, "remediationCorrected" reads like "correctionCorrected", which is kinda weird imo. I also think `Remediation date non-compliance was corrected` doesn't make any sense, but that one's out of our hands. 🤷

services/ui-src/src/forms/mcpar/mcpar.json
```diff
@@ -2983,12 +2983,36 @@
...
                   "label": "D3.VIII.8 Remediation date non-compliance was corrected",
-                  "hint": "What day was it remediated, i.e. When was the non-compliance corrected? Select a date."
+                  "hint": "Was the non-compliance remediated?",
+                  "choices": [
+                    {
+                      "id": "J9FTVN9fJpgXX6X2xJiAiP",
+                      "label": "Yes",
+                      "children": [
+                        {
+                          "id": "sanction_remediationCorrectedDate",
```
braxex, line 2998: Blocking: This id should remain `sanction_remediationDate`. This is because there are already reports in the database that have data for this field id and if it changes we have to handle for both, so anywhere we reference it in the codebase it would have to be `fieldAValue || fieldBValue`, which could get out of hand quickly. Changing each instance of the following across the four files will fix it:

- `sanction_remediationCorrectedDate` goes back to `sanction_remediationDate`
- `remediationCorrectedDate` goes back to `remediationDate`

braxex (changes requested):

braxex:
@karla-vm We also need to check with David and see what he would like us to display as a fallback if there is no date in `remediationDate`, like "N/A" or "TBD" or something.

With date

Screen Shot 2022-11-16 at 15 25 22

Without date

Screen Shot 2022-11-16 at 15 26 06

braxex:
@karla-vm One more note. Because this is an update to the MCPAR json, we should also change the version on line 5 before sending it out.

gmrabian (approved):
Lgtm! Didn't see any data issues and it didn't break old reports

braxex (approved):

PR #4824, opened 2022-11-17
use formTemplate from database instead of local formTemplate

braxex:
## Description

We discovered that we had not made the final change needed to fully implement our move to using the `formTemplate` stored in the `reportJson` alongside each report in the database and were still relying on the local `formTemplate` to drive render of forms. This PR completes that final step.

Changes:
- `AppRoutes` no longer passes through the `ReportRoute` object to the `ReportPageWrapper` for rendering report pages.
- `ReportPageWrapper` no longer accepts a prop from `AppRoutes`. Instead, it determines what report route to render by comparing the pathname (from the `useLocation` hook, which is a new dependency) to the list of routes provided by `ReportContext` from the database. This ensures that the form rendered is the form that is stored in the database for any given report. If it cannot find a matching route for some reason, the form is not rendered.

Changes to testing:
- `useLocation` is now mocked in `ReportPageWrapper`.
- `ReportPageWrapper.test` has been greatly simplified because it no longer needs to provide different props to each page type -- only have a different route pathname mocked. 
- To facilitate this cleanup, the `mockReviewSubmitPageJson` has been moved to `setupJest` where it belongs, and needed downstream changes to both `ReportPageWrapper.test` and `routing.test` were made.
- An additional test was added to `ReportPageWrapper.test` to ensure that if a matching route cannot be found, the form is not rendered.

### How to test

1. Create a new report on the dashboard. 
2. Navigate around the report and observe that everything is going great.
3. Change the name of a field in the local `mcpar.json` of whatever page you're on. 
4. Refresh the page. You should not see the change on the page.

### Changed Dependencies

n/a

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/app/AppRoutes.tsx
+++ b/services/ui-src/src/components/app/AppRoutes.tsx
@@ -37,7 +37,7 @@ export const AppRoutes = () => {
             <Route
               key={route.path}
               path={route.path}
-              element={<ReportPageWrapper routeKey={route.path} />}
+              element={<ReportPageWrapper />}
             />
           ))}
           <Route path="/mcpar/*" element={<Navigate to="/mcpar" />} />
--- a/services/ui-src/src/components/app/AppRoutes.tsx
+++ b/services/ui-src/src/components/app/AppRoutes.tsx
@@ -37,7 +37,7 @@ export const AppRoutes = () => {
             <Route
               key={route.path}
               path={route.path}
-              element={<ReportPageWrapper route={route} />}
+              element={<ReportPageWrapper routeKey={route.path} />}
             />
           ))}
           <Route path="/mcpar/*" element={<Navigate to="/mcpar" />} />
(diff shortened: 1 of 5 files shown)


services/ui-src/src/components/app/AppRoutes.tsx
```diff
@@ -37,7 +37,7 @@ export const AppRoutes = () => {
             <Route
               key={route.path}
               path={route.path}
-              element={<ReportPageWrapper route={route} />}
+              element={<ReportPageWrapper routeKey={route.path} />}
```
braxex, line 40: Do we even need to pass this here? We could just use the pathname from the location. 🤷

gmrabian: yes and pathname is safer because it's using the DB report, not the local one

scwambach (approved):
can't find anything wrong with it 👍 👍 👍

britt-mo (approved):

gmrabian (approved):
Works! Refresh is all good. 
Noticed our `formTemplate` has all the fields duplicated 🤔 so I changed one chunk of text and the UI stayed the same. I found the duplicate text, changed it, it changed the UI....
Just thinking about my ETL ticket. If it replaces both then that's cool haven't figured out why it's duplicated

PR #4828, opened 2022-11-17
mdct-2121 -add second cta button

britt-mo:
## Description
Add a second call to action button if the entity count is greater than 1.

### How to test
see that with 0 or 1 entity, there is only one 'add' button. If two or more, there is a second 'add' button after the last entity.
 
## Code author checklist

### DIFF
--- a/services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
+++ b/services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
@@ -137,16 +137,17 @@ export const ModalDrawerReportPage = ({ route }: Props) => {
     verbiage.countEntitiesInTitle ? ` ${reportFieldDataEntities.length}` : ""
   }`;
 
+  const addEntityButton = (
+    <Button sx={sx.addEntityButton} onClick={addEditEntityModalOnOpenHandler}>
+      {verbiage.addEntityButtonText}
+    </Button>
+  );
+
   return (
     <Box data-testid="modal-drawer-report-page">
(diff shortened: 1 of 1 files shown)


services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
```diff
@@ -137,16 +137,17 @@ export const ModalDrawerReportPage = ({ route }: Props) => {
     verbiage.countEntitiesInTitle ? ` ${reportFieldDataEntities.length}` : ""
   }`;
 
+  const addEntityButton = (
+    <Button sx={sx.addEntityButton} onClick={addEditEntityModalOnOpenHandler}>
+      {verbiage.addEntityButtonText}
+    </Button>
+  );
```
braxex, line 144: Request: It turns out we'll need different styling for the second button. Because of that, we might want to just un-variable-ize this one and repeat `<Button />` for each one but with different styles. The second button should have 2rem top margin and 0 bottom margin.

braxex (changes requested):
Looking good -- just one requested change. 

Also, we should make sure that this doesn't cause any trouble with the new form integration tests for this page -- i.e. selecting the button will become a `getAll` vs. `get`

gmrabian:
Works! I'll approve after above changes

services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
britt-mo: ahhh okay

gmrabian (approved):

scwambach (approved):

PR #4833, opened 2022-11-18
autosave for text field (proof of concept for all field types)

braxex:
## Description

This PR is a proof of concept for autosave for all field types. It implements conditional autosave (only on StandardReportPage) for the `TextField` component and the two other field types based on it: `TextAreaField` and `NumberField`. It also allows for saving blank values to the database on blur if the field is invalid. This will cause the field to re-render as a freshly reset field on re-hydration.

Still to do:
- Change server-side validation rules for all TextField validation types to allow default values.
- Move autosave onSubmit to a utility function we can share among all the field types.
- Remove tests from StandardReportPage.

Changes:
- Added `autosave` prop to `Form.tsx` which can be passed in from any instance (right now it's only from `StandardReportPage` and that's all we need for autosave). This prop is passed on to fields via the `formFieldFactory`.
- Modified `formFieldFactory` utility method to accept an options object and pass them all through to the created fields, 'cause that's cleaner and easier to work with when you have this many options going in there.
- Added `autosave` prop to `TextField` (and also ChoiceListField, DateField, and DropdownField even though we aren't fully wiring those up yet, just to quiet some console errors).
- Added `onBlur` data submission method to `TextField` so anytime a user blurs off a field it triggers autosave logic.
- Because the `StandardReportPage` will no longer be saving data at the page level (it's all happening via autosave now), that component is _way_ simpler.
- However, because `TextField` needs to make a database call now, it got a bit more complicated. I'm considering making a utility that will help with this though. 
- `ReportPageFooter` needed to have the `type="submit"` and `form={form.id}` added back in because this is what triggers the form level onSubmit function (which triggers the form-level validation and error handling) and will eventually be what we use to trigger statusing checks and updates. However, because it no longer submits any data or performs any async functions, I removed the `submitting` spinner.

Tests to write:
- For each field type:
    - If form has specified "autosave" option and the user is authorized, the field is checked for validity and `updateReport` is called. If either condition is not met, the field is not checked for validity and `updateReport` is not called.
    - If field passes validity check, the field's value is passed as the data for the `updateReport` call.
    - If field does not pass validity check, the field's _default_ value is passed as the data for the `updateReport` call.
    - if field's default value is passed as the data for the `updateReport` call, the server-side validation passes the default value through and allows the default value to be set in the database.

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/fields/TextField.tsx
+++ b/services/ui-src/src/components/fields/TextField.tsx
@@ -44,6 +44,11 @@ export const TextField = ({
     form.setValue(name, value, { shouldValidate: true });
   };
 
+  // ON BLUR
+  const onBlurHandler = async (event: InputChangeEvent) => {
+    // TODO: SUBMIT ONLY THIS FIELD DATA HERE
+  };
+
   // prepare error message, hint, and classes
   const formErrorState = form?.formState?.errors;
   const errorMessage = formErrorState?.[name]?.message;
@@ -60,6 +65,7 @@ export const TextField = ({
         hint={parsedHint}
         placeholder={placeholder}
         onChange={(e) => onChangeHandler(e)}
+        onBlur={(e) => onBlurHandler(e)}
         errorMessage={errorMessage}
         value={displayValue}
         {...props}
--- a/services/ui-src/src/components/fields/TextField.tsx
+++ b/services/ui-src/src/components/fields/TextField.tsx
@@ -1,11 +1,17 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useContext } from "react";
 import { useFormContext } from "react-hook-form";
 // components
 import { TextField as CmsdsTextField } from "@cmsgov/design-system";
 import { Box } from "@chakra-ui/react";
+import { ReportContext } from "components";
 // utils
-import { parseCustomHtml } from "utils";
-import { InputChangeEvent, AnyObject, CustomHtmlElement } from "types";
+import { parseCustomHtml, useUser } from "utils";
+import {
+  InputChangeEvent,
+  AnyObject,
+  CustomHtmlElement,
+  ReportStatus,
+} from "types";
 
 export const TextField = ({
   name,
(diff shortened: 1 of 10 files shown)


services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -44,6 +56,30 @@ export const TextField = ({
...
+        const reportKeys = {
+          state: state,
+          id: report?.id,
+        };
+        const dataToWrite = {
+          status: ReportStatus.IN_PROGRESS,
+          lastAlteredBy: full_name,
+          fieldData: { [name]: fieldValue },
+        };
+        await updateReport(reportKeys, dataToWrite);
```
gmrabian, line 78: wondering if we need to await this. It should queue the api calls if we get multiple. Would have to check

braxex: hmm yeah we should check on that and see. you're right that it probably doesn't make sense to await this.

braxex:
@gmrabian @scwambach what do you think of the server-side validation change here and the others that would need to happen to make storing blank values a thing?

gmrabian:
@braxex getting back to this
1. Nice
2. NumberField didn't save (Section B.I)
3. Blanking out a TextField works and updates the db (!) but also the `Continue` stops you from moving on (because required). So I don't know if we want to let people Continue after doing so, and let the status indicators handle that "required/incomplete" case?
4. The save is real quick and I like that
5. It does throw a ~small~ significant wrench in the server-side validation...could we change the validation (or all validations) to allow a blank value? That way numbers still have to be numbers, strings strings, etc., but they can also be blank?
Actually now that I think about it more I don't think the backend needs the "required" case, right (maybe with the exception of expected metadata, where we could add required tags)? But like if we get a value in the backend and want to validate that it's a number, it doesn't have to be required, no? It could just be a number or empty...cuz all we care about is weird stuff that we don't want in the db...but empty don't hurt...and the frontend handles required/missing cases with indicators and stuff....

that's my Friday Morning Impressions

PR #4893, opened 2022-11-28
Uncomment quick tests

britt-mo:
## Code author checklist


scwambach (approved):

gmrabian:
Were these commented out because they take a long time? This is just too long for an integration test imo (still running)

Screen Shot 2022-11-29 at 2 32 22 PM

ntsummers1:
Reason this was taking so long is because of the 6 report query limit - So when it can't find the report it created it'll just fail over and over trying to fill out the rest of the report which is why it took soooo long. It's just timing out each time. We should look into implenting something so that time outs or repeated failures just end so cases like this dont happen but for now we should close this as we're attempting to fix the program limit issue

PR #4927, opened 2022-11-30
Get FormData and FormTemplate after a user clicks a Program

ntsummers1:
## Description

Currently theres a 1mb limit in the size of requests we can do, and a 6mb limit overall on how much data we can return. So that means when we call the dashboard it returns about 24 programs because of the sheer size of the request. To fix that, we've gone with removing the formTemplate and formData from the dashboard call. That way we can get way more programs to show on the dashboard, and then get the information for the rest of the program after a user has entered the details of that program.

### How to test

Navigate to a state users dashboard
Create a bunch of programs
Enter any and all of them!

## Code author checklist

### DIFF
--- a/services/app-api/handlers/reports/fetch.ts
+++ b/services/app-api/handlers/reports/fetch.ts
@@ -1,6 +1,6 @@
 import handler from "../handler-lib";
 import dynamoDb from "../../utils/dynamo/dynamodb-lib";
-import { StatusCodes } from "../../utils/types/types";
+import { AnyObject, StatusCodes } from "../../utils/types/types";
 import error from "../../utils/constants/constants";
 
 export const fetchReport = handler(async (event, _context) => {
@@ -59,10 +59,17 @@ export const fetchReportsByState = handler(async (event, _context) => {
   // Looping to perform complete scan of tables due to 1 mb limit per iteration
   do {
     [startingKey, results] = await queryTable(startingKey);
-    existingItems.push(
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      ...results.Items.map(({ formTemplate, fieldData, ...item }) => item)
-    );
+
+    /*
+     * Remove formTemplate and formData to get rid of excessive size that isn't needed
+     * on the dashboard when this call is used
+     */
+    const items: AnyObject[] = results.Items;
+    items.forEach((item: any) => {
+      delete item["formTemplate"];
+      delete item["formData"];
+    });
+    existingItems.push(...items);
   } while (startingKey);
 
   return {
--- a/services/app-api/handlers/reports/fetch.ts
+++ b/services/app-api/handlers/reports/fetch.ts
@@ -42,35 +42,29 @@ export const fetchReportsByState = handler(async (event, _context) => {
   };
 
   let startingKey;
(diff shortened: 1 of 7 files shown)


services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
```diff
@@ -93,9 +94,16 @@ export const DashboardPage = ({ reportType }: Props) => {
   }, [reportsByState]);
 
   const enterSelectedReport = async (report: ReportShape) => {
+    // Get report details here maybe?
```
ntsummers1, line 97:
```suggestion
```

services/app-api/handlers/reports/fetch.ts
```diff
@@ -42,35 +42,29 @@ export const fetchReportsByState = handler(async (event, _context) => {
...
-      startingKey
+  do {
+    [startingKey, results] = await queryTable(startingKey);
+    existingItems.push(
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      ...results.Items.map(({ formTemplate, fieldData, ...item }) => item)
     );
-    existingItems.push(...results.Items);
-  }
+  } while (startingKey);
```
gmrabian, line 66: Love the rewrite here, though I'm not quite following the map logic. How is it filtering?

ntsummers1: So in Javascript you can call out the fields you want to edit and overwrite them in. So you've likely seen something along the lines of this:

```
const object1 = {a: "value for a", b: "value for b"};

const object2 = {...object1, a: "updated value"};
```

object2 would then be `{a: "updated value", b: "value for b"};`


If we do a map in a similar fashion, we can call out the properties of the object we're mapping and just simply ignore them because we didn't return it. So using the spread operator I'm calling out all properties that aren't formTemplate and fieldData and returning that as I iterate through the array. It's a quick way to filter out properties that you don't want. Only downside is it leaves you with the unused vars scenario.

services/app-api/handlers/reports/fetch.ts
```diff
@@ -42,35 +42,36 @@ export const fetchReportsByState = handler(async (event, _context) => {
...
+     * Remove formTemplate and formData to get rid of excessive size that isn't needed
+     * on the dashboard when this call is used
+     */
+    const items: AnyObject[] = results.Items;
+    items.forEach((item: any) => {
+      delete item["formTemplate"];
+      delete item["formData"];
+    });
+    existingItems.push(...items);
+  } while (startingKey);
```
ntsummers1, line 73: This won't stop the 6mb limit from being triggered but the size of these requests has been reduced significantly. Thousands of reports would have to be in an individual states dashboard before they hit the limit cap of 6mb.

services/app-api/handlers/reports/fetch.ts
ntsummers1: Updated this to make it a lot more clear about whats happening. Totally not because it wouldn't deploy due to typescript linting errors

ntsummers1:
Moved to https://github.com/CMSgov/mdct-mcr/pull/4932 due to branch name conflicting with bucket naming scheme

PR #4932, opened 2022-12-01
Get FormData and FormTemplate after a user clicks a Program

ntsummers1:
## Description

Currently theres a 1mb limit in the size of requests we can do, and a 6mb limit overall on how much data we can return. So that means when we call the dashboard it returns about 24 programs because of the sheer size of the request. To fix that, we've gone with removing the formTemplate and formData from the dashboard call. That way we can get way more programs to show on the dashboard, and then get the information for the rest of the program after a user has entered the details of that program.

### How to test

Navigate to a state users dashboard
Create a bunch of programs
Enter any and all of them!

## Code author checklist

### DIFF
--- a/services/app-api/handlers/reports/fetch.ts
+++ b/services/app-api/handlers/reports/fetch.ts
@@ -66,8 +66,8 @@ export const fetchReportsByState = handler(async (event, _context) => {
      */
     const items: AnyObject[] = results.Items;
     items.forEach((item: any) => {
-      delete item["formTemplate"];
-      delete item["formData"];
+      delete item.formTemplate;
+      delete item.formData;
     });
     existingItems.push(...items);
   } while (startingKey);
--- a/services/app-api/handlers/reports/fetch.ts
+++ b/services/app-api/handlers/reports/fetch.ts
@@ -1,6 +1,6 @@
 import handler from "../handler-lib";
 import dynamoDb from "../../utils/dynamo/dynamodb-lib";
-import { StatusCodes } from "../../utils/types/types";
+import { AnyObject, StatusCodes } from "../../utils/types/types";
 import error from "../../utils/constants/constants";
 
 export const fetchReport = handler(async (event, _context) => {
@@ -59,10 +59,17 @@ export const fetchReportsByState = handler(async (event, _context) => {
   // Looping to perform complete scan of tables due to 1 mb limit per iteration
   do {
     [startingKey, results] = await queryTable(startingKey);
-    existingItems.push(
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      ...results.Items.map(({ formTemplate, fieldData, ...item }) => item)
-    );
+
+    /*
+     * Remove formTemplate and formData to get rid of excessive size that isn't needed
+     * on the dashboard when this call is used
+     */
+    const items: AnyObject[] = results.Items;
+    items.forEach((item: any) => {
+      delete item["formTemplate"];
+      delete item["formData"];
+    });
+    existingItems.push(...items);
   } while (startingKey);
 
   return {
--- a/services/app-api/handlers/reports/fetch.ts
+++ b/services/app-api/handlers/reports/fetch.ts
@@ -42,35 +42,29 @@ export const fetchReportsByState = handler(async (event, _context) => {
   };
 
   let startingKey;
-  let keepSearching = true;
   let existingItems = [];
   let results;
-  let count = 0;
 
-  const queryTable = async (keepSearching: boolean, startingKey?: any) => {
+  const queryTable = async (startingKey?: any) => {
     queryParams.ExclusiveStartKey = startingKey;
     let results = await dynamoDb.query(queryParams);
     if (results.LastEvaluatedKey) {
       startingKey = results.LastEvaluatedKey;
-      return [startingKey, keepSearching, results];
+      return [startingKey, results];
     } else {
-      keepSearching = false;
-      return [null, keepSearching, results];
+      return [null, results];
     }
   };
 
   // Looping to perform complete scan of tables due to 1 mb limit per iteration
-  while (keepSearching) {
-    count++;
-    if (count > 4) {
-      break;
-    }
-    [startingKey, keepSearching, results] = await queryTable(
-      keepSearching,
-      startingKey
+  do {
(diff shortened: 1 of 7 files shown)


services/app-api/handlers/reports/fetch.ts
```diff
@@ -42,35 +42,36 @@ export const fetchReportsByState = handler(async (event, _context) => {
...
+     * Remove formTemplate and formData to get rid of excessive size that isn't needed
+     * on the dashboard when this call is used
+     */
+    const items: AnyObject[] = results.Items;
+    items.forEach((item: any) => {
+      delete item["formTemplate"];
+      delete item["formData"];
+    });
+    existingItems.push(...items);
+  } while (startingKey);
```
ntsummers1, line 73: This won't stop the 6mb limit from being triggered but the size of these requests has been reduced significantly. Thousands of reports would have to be in an individual states dashboard before they hit the limit cap of 6mb.

services/ui-src/src/types/index.ts
```diff
@@ -138,7 +138,7 @@ export interface ReportKeys {
   id: string;
 }
 
-export interface ReportShape extends ReportKeys {
+export interface DashboardReportShape extends ReportKeys {
```
ntsummers1, line 141: Creating an interface called DashboardReportShape and having ReportShape extend off of it made things a LOT easier. It could be argued that reportShape could just have formTemplate and fieldData swapped to an optional param but going that route includes a lot more rewrites, type checks, and testing.

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
```diff
@@ -86,16 +92,21 @@ export const DashboardPage = ({ reportType }: Props) => {
...
     setReportsToDisplay(newReportsToDisplay);
   }, [reportsByState]);
 
-  const enterSelectedReport = async (report: ReportShape) => {
+  const enterSelectedReport = async (report: DashboardReportShape) => {
+    const reportKeys: ReportKeys = {
+      state: report.state,
+      id: report.id,
+    };
+    const foundReport: ReportShape = await fetchReport(reportKeys);
```
britt-mo, line 106: 'foundReport' refers to a report that the user entered, right? Not sure the naming reflects that. But it's a nit.

ntsummers1: Nah its a good call out. Any recommendations for what you would call it instead? Would `reportToEnter` make more sense?

services/app-api/handlers/reports/fetch.ts
```diff
@@ -42,35 +42,36 @@ export const fetchReportsByState = handler(async (event, _context) => {
...
+    [startingKey, results] = await queryTable(startingKey);
+
+    /*
+     * Remove formTemplate and formData to get rid of excessive size that isn't needed
+     * on the dashboard when this call is used
+     */
+    const items: AnyObject[] = results.Items;
+    items.forEach((item: any) => {
+      delete item["formTemplate"];
+      delete item["formData"];
```
braxex, line 70: Nit: Could these be accessed with dot notation instead? i.e. `item.formTemplate` and `item.formData`

services/ui-src/src/types/index.ts
braxex: This makes a lot of sense to me, but I would suggest naming it something more descriptive of what it actually is, rather than where it's used, maybe like `ReportMetadataShape`.

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
braxex: `selectedReport` 😀

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
```diff
@@ -86,16 +92,21 @@ export const DashboardPage = ({ reportType }: Props) => {
...
       newReportsToDisplay = reportsByState?.filter(
-        (report: ReportShape) => !report?.archived
+        (report: DashboardReportShape) => !report?.archived
       );
     }
     setReportsToDisplay(newReportsToDisplay);
   }, [reportsByState]);
 
-  const enterSelectedReport = async (report: ReportShape) => {
+  const enterSelectedReport = async (report: DashboardReportShape) => {
```
braxex, line 101: Thought: I think we should wrap this in a try/catch block in case the report cannot be loaded for whatever reason and add some error handling here that displays an error message if it can't be loaded and entered. Might have to be a follow-up item.

scwambach (approved):
working great on my end! 👍

services/ui-src/src/types/index.ts
```diff
@@ -152,9 +152,13 @@ export interface ReportShape extends ReportKeys {
   submittedBy?: string;
   submitterEmail?: string;
   submittedOnDate?: number;
+  formTemplate?: ReportJson;
+  fieldData?: AnyObject;
```
braxex, line 156: Request: I don't think these should be necessary anymore because this type should never have those two properties on it, right?

```suggestion
```

services/ui-src/src/types/index.ts
```diff
@@ -152,9 +152,13 @@ export interface ReportShape extends ReportKeys {
...
+  archived?: boolean;
+}
+export interface ReportShape extends DashboardReportShape {
```
braxex, line 159: Nit: Add the space here back in.
```suggestion
}

export interface ReportShape extends DashboardReportShape {
```

braxex:
@ntsummers1 This looks really good! Left a couple comments.

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
britt-mo: Maybe touchedReport?

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
ntsummers1: Agree - [Heres a ticket to add this in as a separate PR. ](https://qmacbis.atlassian.net/browse/MDCT-2143)

braxex (approved):
It's beautiful. 😢 Please make my life complete by merging. 🚢

PR #5117, opened 2022-12-20
Extract Section Heading (WIP)

karla-vm:


services/ui-src/src/components/export/ExportedSectionHeading.tsx
```diff
@@ -0,0 +1,47 @@
+// components
+import { Box, Heading } from "@chakra-ui/react";
+import { SpreadsheetWidget } from "components";
+// types
+
+// utils
+import { parseCustomHtml } from "utils";
+
+export const ExportedSectionHeading = ({ heading, verbiage }: any) => {
```
braxex, line 9: Can remove types here or strict 'em up -- doesn't matter to me.

services/ui-src/src/components/export/ExportedSectionHeading.tsx
```diff
@@ -0,0 +1,47 @@
+// components
+import { Box, Heading } from "@chakra-ui/react";
+import { SpreadsheetWidget } from "components";
+// types
+
```
braxex, line 5: Remove or fill

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
```diff
@@ -36,18 +36,7 @@ export const ExportedModalDrawerReportSection = ({
...
-        </Heading>
-      )}
-      {existingEntity && verbiage?.intro?.info && (
-        <Box sx={sx.intro}>{parseCustomHtml(verbiage.intro.info)}</Box>
-      )}
-
-      {existingEntity && verbiage?.intro?.spreadsheet && (
-        <Box sx={sx.spreadSheet}>
-          <SpreadsheetWidget description={verbiage.intro.spreadsheet} />
-        </Box>
```
braxex, line 50: Hmm... need to think about how to handle this. For modals, if it doesn't have any entities currently, it needs to only show the heading. Plenty of options -- I'm sure whatever you come up with will be great.

braxex: Could even just add another optional prop that's like: `shouldShowExtras` or something (but maybe with a better name). Prop could default to `= true` when initially passed into the args so standard and drawer don't have to pass anything in, but modal could pass in like `entityCount` so if it's 0, they don't get rendered and if it's greater than 0 they do. 🤔

braxex (approved):
Approved, with fixes as noted.

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
karla-vm: I went with no types for the parameters, but I'm open to suggestions (this is more because I'm not sure how to best do this one). For the modals, there is already a boolean that indicates whether we have entities so I'm just reusing that!

PR #5208, opened 2023-01-03
add count to modal drawer sections

britt-mo:
## Description

If there are cards to display for modal drawers, the count should display above the cards (same view as regular program). If there are no cards, the header should still appear as normal but instead of a count it should read: '0 -- No {entity} entered' in the error-darker color.

### How to test
Add some data, see that count shows up correctly, when there's no data, see that '0 -- No {entity} entered' shows.

## Code author checklist


braxex:
@britt-mo Is this ready for review or are you still working on it?

britt-mo:
> @britt-mo Is this ready for review or are you still working on it?

Good for review

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
```diff
@@ -17,21 +17,35 @@ export const ExportedModalDrawerReportSection = ({
   const sectionHeading = verbiage?.intro.subsection || name;
   const entityCount = report?.fieldData?.[entityType]?.length;
 
-  let emptyEntityMessage = "";
+  let emptyEntityMessage;
   switch (entityType) {
     case ModalDrawerEntityTypes.ACCESS_MEASURES:
-      emptyEntityMessage = `No access and adequacy measures have been entered for this program report.`;
+      emptyEntityMessage = " 0 -- No access measures entered";
```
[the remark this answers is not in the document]
braxex: Thought: Instead of including the number in the empty entity message, you could change line 18 to return either the current code or 0, like: 

```
const entityCount = report?.fieldData?.[entityType]?.length || 0;
```

Then `emptyEntityMessage` is just `No access measures entered` or whichever is right. Then we can simplify below (see my note there).

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
```diff
@@ -17,21 +17,35 @@ export const ExportedModalDrawerReportSection = ({
...
+    <Heading as="h3" sx={sx.dashboardTitle} data-testid="headerCount">
+      {verbiage.dashboardTitle}
+      {entityCount ? (
+        ` ${entityCount}`
+      ) : (
+        <Text as="span" sx={sx.notAnswered} data-testid="entityMessage">
+          {emptyEntityMessage}
+        </Text>
+      )}
+    </Heading>
```
braxex, line 46: Thought, continued from above: Then we could simplify this like:

```suggestion
    <Heading as="h3" sx={sx.dashboardTitle} data-testid="headerCount">
      {verbiage.dashboardTitle} ${entityCount} - 
      {!entityCount && (
        <Text as="span" sx={sx.notAnswered} data-testid="entityMessage">
          {emptyEntityMessage}
        </Text>
      )}
    </Heading>
```

`entityCount` always shows because the count number should always show, and the additional empty message only shows if there are no entities.

braxex (changes requested):

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
britt-mo: One thing here is that the `0` has to be red as well, but I do like this idea.

britt-mo: this change gives the desired effect:
```

  const entityCount = report?.fieldData?.[entityType]?.length;

  let emptyEntityMessage;
  switch (entityType) {
    case ModalDrawerEntityTypes.ACCESS_MEASURES:
      emptyEntityMessage = "No access measures entered";
      break;
    case ModalDrawerEntityTypes.SANCTIONS:
      emptyEntityMessage = "No sanctions entered";
      break;
    case ModalDrawerEntityTypes.QUALITY_MEASURES: {
      emptyEntityMessage = "No quality measures entered";
      break;
    }
    default:
      break;
  }

  const headerEntityCount = (
    <Heading as="h3" sx={sx.dashboardTitle} data-testid="headerCount">
      {verbiage.dashboardTitle} {entityCount} 
      {!entityCount && (
        <Text as="span" sx={sx.notAnswered} data-testid="entityMessage">
          0 - {emptyEntityMessage}
        </Text>
      )}
    </Heading>
  );
```

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
```diff
@@ -17,21 +17,33 @@ export const ExportedModalDrawerReportSection = ({
   const sectionHeading = verbiage?.intro.subsection || name;
   const entityCount = report?.fieldData?.[entityType]?.length;
 
-  let emptyEntityMessage = "";
+  let emptyEntityMessage;
```
tblackwe, line 20: The default case will leave this undefined. is that intentional?

tblackwe (approved):

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
britt-mo: It was a change I made when I was trying to refactor a bit and didn't change it back. I can set it to a string though

braxex (approved):

braxex:
Just gotta fix up those failing e2e tests, then you should be good.

PR #5290, opened 2023-01-11
remove some test users; add dedicated cypress test users

braxex:
what if this?

by assigning them to DC we can probably avoid lots of the mess created in envs.


braxex:
@tblackwe @BearHanded thoughts?

braxex:
> @tblackwe @BearHanded thoughts?

What do you think?

tblackwe (commented):
Approach looks good to me, but I am unfamiliar with how our accounts are administered.

How does the Incognito password tie into these 2 new users?

services/ui-auth/libs/users.json
```diff
@@ -249,48 +249,44 @@
...
         "Value": "true"
       },
       {
         "Name": "custom:cms_roles",
-        "Value": "mdctmcr-state-user"
-      },
-      {
-        "Name": "custom:cms_state",
-        "Value": "CA"
+        "Value": "mdctmcr-bor"
```
tblackwe, line 272: Seeking to understand: how this value used?

tblackwe (approved):

karla-vm (approved):

PR #5355, opened 2023-01-27
utility (wip draft)

braxex:
## Code author checklist


services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -1,50 +1,74 @@
...
+    updateReport: Function;
+  };
+  user: {
+    userName: string | undefined;
+    state: string | undefined;
+    isAuthorizedUser: boolean;
+  };
+}
+
+export const autosaveFieldData = async ({
```
braxex, line 28: Note: I compiled all the methods into one that handles everything autosave. It's a bit big, but it handles everything, so we have central location to look. My first commit made the baseline for it for TextField, then I realized we would need to be able to save multiple fields at once for ChoiceListFields, so I modified it accordingly.

braxex: Note: None of the tests are updated, and all of the fields except TextField and ChoiceListField are still using the old methods, so there will be failures unless we move those over. Just a proof of concept right now. 🤷

services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -70,23 +53,15 @@ export const TextField = ({
...
-      await updateReport(reportKeys, dataToWrite);
+    if (autosave) {
+      const fields = [{ name, value, hydrationValue, defaultValue }];
+      const reportArgs = { id: report?.id, updateReport };
+      const user = {
+        userName: full_name,
+        state,
+        isAuthorizedUser: !!(userIsStateRep || userIsStateUser),
+      };
+      await autosaveFieldData({ form, fields, report: reportArgs, user });
```
braxex, line 64: Note: All the autosave stuff in each field file slims down to just this: creating the args that go into the `autosaveFieldData` method.

services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -27,14 +20,6 @@ export const TextField = ({
 }: Props) => {
   const defaultValue = "";
   const [displayValue, setDisplayValue] = useState<string>(defaultValue);
-
-  /*
-   * Last value is the last value that was submitted to Autosave. We only want to submit
-   * To autosave if theres been a change
-   */
-  const [lastAutosaveValue, setLastAutosaveValue] =
-    useState<string>(defaultValue);
```
braxex, line 36: Note: I don't think we need this anymore because we can rely on the `hydrationData` updating from the database on every `updateReport` call as I noted in the other PR's comments, but I might be missing something, so let's chat about it.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -67,23 +64,10 @@ export const ChoiceListField = ({
...
-    const cleanedDataToSend = clearChildren(choices, initialDataToSend);
-
-    const reportKeys = createReportKeys(report?.id, state);
-    const dataToWrite = createDataToWrite(
-      { [name]: cleanedDataToSend },
-      full_name
-    );
-
-    await updateReport(reportKeys, dataToWrite);
-  };
```
braxex, line 85: Note: I don't think we'll need this anymore because of the new `getNestedFields` function.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -104,11 +88,7 @@ export const ChoiceListField = ({
     });
   };
 
-  const autosaveContainsParentChoice = (value: string) =>
-    lastAutosaveValue &&
-    lastAutosaveValue.some((autosave) => autosave.value === value);
```
braxex, line 109: Note: I don't think we'll need this anymore because of the new `getNestedFields` function.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -104,11 +88,7 @@ export const ChoiceListField = ({
...
-
-  const clearChildren = (choices: FieldChoice[], dataToSend?: AnyObject) => {
```
braxex, line 111: Note: I simplified this a bit because I don't think we'll need all the functionality with the new `getNestedFields` function.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -171,18 +144,53 @@ export const ChoiceListField = ({
     }
   };
 
-  // update field values
+  const getNestedFields = (choices: FieldChoice[]) => {
```
braxex, line 147: Note: This function compiles the field info needed for all the nested child fields to be blanked on autosave. It checks the parent choices to see which ones are unchecked, then dives into each of those and blanks all nested children, regardless of their previous values (including if they are radio/checkbox, it doesn't care what choices are checked/unchecked -- everything gets blanked). All that field info is returned in an array to the autosave handler to be called there.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -171,18 +144,52 @@ export const ChoiceListField = ({
...
-      userIsStateUser
-    );
-    if (willAutosave) {
-      setLastAutosaveValue(displayValue);
-      autosaveSelections(displayValue);
+    if (autosave) {
+      let fields = [
+        { name, value: displayValue, hydrationValue, defaultValue },
+        ...getNestedChildFieldsOfUncheckedParent(choices),
+      ];
```
braxex, line 185: Note: Slightly different than other fields because we have to pass all the nested children of the unchecked parent.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -171,18 +144,52 @@ export const ChoiceListField = ({
...
+    const nestedFields: any = [];
+    const compileNestedFields = (fields: FormField[]) => {
+      fields.forEach((field: FormField) => {
+        // for each child field, get field info
+        const fieldDefaultValue = ["radio", "checkbox"].includes(field.type)
+          ? null
+          : "";
+        const fieldInfo = {
+          name: field.id,
+          shouldClear: true,
```
braxex, line 158: Bug: Is saving regardless of field value. Might already be cleared and therefore not need to save.
```suggestion
          shouldClear: form.getValues(field.id) !== fieldDefaultValue,
```

PR #5360, opened 2023-01-30
Autosave Dynamic Fields

ntsummers1:
## Code author checklist

### DIFF
--- a/services/ui-src/src/utils/autosave/autosave.ts
+++ b/services/ui-src/src/utils/autosave/autosave.ts
@@ -1,8 +1,7 @@
-import { isEqual } from "lodash";
 import { FieldValues, UseFormReturn } from "react-hook-form";
 import { Choice, DropdownChoice, EntityShape, ReportStatus } from "types";
 
-type FieldValue = string | DropdownChoice | Choice[] | null;
+type FieldValue = string | DropdownChoice | Choice[] | EntityShape[] | null;
 
 type FieldTuple = [string, FieldValue];
 
@@ -26,6 +25,7 @@ interface Props {
     userName: string | undefined;
     state: string | undefined;
   };
+  fieldType?: string;
 }
 
 export const autosaveFieldData = async ({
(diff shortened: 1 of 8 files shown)


services/ui-src/src/utils/forms/forms.ts
```diff
@@ -42,6 +42,7 @@ export const formFieldFactory = (
       key: field.id,
       name: field.id,
       hydrate: field.props?.hydrate,
+      fieldType: field.type,
```
braxex, line 45: Error: Passing `fieldType` as an element attribute with `React.createElement()` won't work because of this error. If we need to access a field's type, we have to name it something else or set it directly in the component and pass it through that way. However, I think we have better options.
Screenshot 2023-01-30 at 17 39 05

services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -26,35 +25,50 @@ interface Props {
...
-    // TODO: do we need a deeper equality check here?
-    (field: FieldInfo) => {
-      const { value, hydrationValue, defaultValue } = field;
-      return value !== defaultValue && !isEqual(value, hydrationValue);
-    }
-  );
+    default:
+      // if field value hasn't changed from database value, don't autosave field
+      onlyChangedFields = fields.filter(
+        // TODO: do we need a deeper equality check here?
```
braxex, line 55: Note: Good call that we don't need deep equality checking here. From doing a little testing I think the baseline equality checking does just fine for us. 👍

services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -26,35 +25,50 @@ interface Props {
...
+  switch (fieldType) {
+    case "dynamic":
+      onlyChangedFields = fields.filter((field: FieldInfo) => {
+        const checkedValues = field.displayValues?.filter((el) => el.name);
+        return (
+          checkedValues?.length !== 0 &&
+          field.displayValues !== field?.hydrationValue
+        );
+      });
+      break;
```
braxex, line 50: Thought: If even one of the entity names has changed, we need to make the database call, in which case we might as well update all the entity names, so there's probably no need to go through each one individually. The top-level check of value vs. hydration value should be enough. 

We also don't need to pass through a separate `displayValues` prop here -- we can just pass it as the `value` prop that all fields use. That way we can just use the exact same code for dynamicFields that we use for everything else, _and_ avoid having to pass through any information about the field type.

services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -26,35 +25,50 @@ interface Props {
...
+  }
 
   // for each passed field, prepare for autosave payload if necessary
   const fieldDataToSaveArray: FieldTuple[] = await Promise.all(
     onlyChangedFields.map(async (field: FieldInfo) => {
-      const { name, defaultValue } = field;
-
+      const { name, defaultValue, value, displayValues } = field;
+      const toSetValue = fieldType === "dynamic" ? displayValues : value;
```
braxex, line 67: Note: I realized we could simplify this method a lot too.

braxex (commented):
I made a PR that might be helpful. It includes a way we can check `dynamicFields` in the exact same way we check all the other fields, as well as some simplification/cleanup of the autosave utility method. Check it out here and let me know what you think: https://github.com/CMSgov/mdct-mcr/pull/5363/files

braxex:
@ntsummers1 Closing for now.

PR #5363, opened 2023-01-31
autosave for dynamic fields, and some autosave utility cleanup too

braxex:
## Description

Fixes autosave for `dynamicField`. The interesting changes are all in `DynamicField` and `autosave.ts`.

Changes:
- Adds a new `overrideCheck` prop for fields so DynamicField and fields nested under ChoiceListField can selectively ignore the react-hook-form client-side validation check results and instead set the value we pass through. More on this in a note in the utility file.
- Removes unnecessary `displayValues` prop from `autosaveFieldData` utility method.
- Removes unnecessary `isAuthorizedUser` artifacts from each field type.

### How to test

Make a new program and start making and saving some fields. I haven't messed with ChoiceListFields at all, but Dynamic and all other standard types should work. With DynamicFields, check that multiple entities can be added, changed, deleted, etc. at once. Then make sure that they can be blanked out. There should be a client-side inline validation error/warning, but it should still allow you to save it as a blank entity name with the same unchanged id. Then it should show up everywhere in the app with a blank name. 

### Changed Dependencies

- Removes `lodash` since we don't need it.

## Code author checklist

### DIFF
--- a/services/ui-src/src/utils/autosave/autosave.ts
+++ b/services/ui-src/src/utils/autosave/autosave.ts
@@ -1,7 +1,7 @@
 import { FieldValues, UseFormReturn } from "react-hook-form";
-import { Choice, DropdownChoice, EntityShape, ReportStatus } from "types";
+import { EntityShape, ReportStatus } from "types";
 
-type FieldValue = string | Choice[] | DropdownChoice | EntityShape[] | null;
+type FieldValue = any;
 
 type FieldDataTuple = [string, FieldValue];
 
@@ -24,6 +24,7 @@ interface Props {
     userName: string | undefined;
     state: string | undefined;
   };
+  fieldType: string;
 }
 
 export const autosaveFieldData = async ({
@@ -31,6 +32,7 @@ export const autosaveFieldData = async ({
   fields,
   report,
   user,
+  fieldType,
 }: Props) => {
   const { id, updateReport } = report;
   const { userName, state } = user;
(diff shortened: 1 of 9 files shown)


services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -33,49 +31,38 @@ export const autosaveFieldData = async ({
...
     };
     await updateReport(reportKeys, dataToWrite);
 
-    // after successful autosave, set field values in form state
-    fieldDataToSaveArray.forEach((field: FieldTuple) => {
-      const [fieldName, fieldValue] = field;
-      form.setValue(fieldName, fieldValue, { shouldValidate: true });
+    // after successful autosave, update field values in form state
+    fieldsToSave.forEach(([name, value]: FieldDataTuple) => {
+      form.setValue(name, value);
```
braxex, line 65: Note: Just simplified.

services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -33,49 +31,38 @@ export const autosaveFieldData = async ({
...
+      .filter((field: FieldInfo) => field.value !== field.hydrationValue)
+      // determine appropriate field value to set and return as tuple
+      .map(async (field: FieldInfo) => {
+        const { name, value, defaultValue, overrideCheck } = field;
+        const fieldValueIsValid = await form.trigger(name);
+        // if field value is valid or validity check overriden, use field value
+        if (fieldValueIsValid || overrideCheck) return [name, value];
+        // otherwise, revert field to default value
+        return [name, defaultValue];
+      })
```
braxex, line 51: Note: Combined the filter and the map and simplified since we no longer need the additional checks.

braxex: Thought: We might want to make this a separate method that we call from `autosaveFieldData` for both clarity and ease of testing.

services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -37,45 +35,34 @@ export const autosaveFieldData = async ({
...
+  const fieldsToSave: FieldDataTuple[] = await Promise.all(
+    fields
+      // filter to only fields with changed values
+      .filter((field: FieldInfo) => field.value !== field.hydrationValue)
+      // determine appropriate field value to set and return as tuple
+      .map(async (field: FieldInfo) => {
+        const { name, value, defaultValue, overrideCheck } = field;
+        const fieldValueIsValid = await form.trigger(name);
+        // if field value is valid or validity check overriden, use field value
+        if (fieldValueIsValid || overrideCheck) return [name, value];
```
braxex, line 48: Note: The new `overrideCheck` prop can be used for:
- DynamicField: Where we want to ignore the result of a field validity check (to allow us to blank internal text inputs) and instead of falling back on a default value, just pass the actual value through.
- ChoiceListField nested children that we want to force-blank or reset when parent field is unchecked: Where we want to ignore the check and set it to a value we choose, which will always be the default value, but which we can just pass as the value (see the change to ChoiceListField).

braxex:
@ntsummers1 @tblackwe @BearHanded @scwambach 

I think this should get us where we need to be for DynamicFields. The updates to this should help with nested children of ChoiceList too. Let's chat tomorrow.

services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -37,45 +35,47 @@ export const autosaveFieldData = async ({
...
+         */
+        if (!field?.defaultValue) return field.value !== field.hydrationValue;
+        return (
+          // Handles most cases where a user wants to update a field
+          field.value !== field.defaultValue ||
+          // Handles case where a user deletes their entry and blurs out of the field
+          (field.value === field.defaultValue &&
+            field.hydrationValue !== undefined &&
+            field.value !== field.hydrationValue)
+        );
```
ntsummers1, line 54: Reverted your code back to this cleanup and then expanded on it with an if statement that should handle any case a user is trying to update a field and we should be notified about it.

services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -26,56 +24,70 @@ interface Props {
     userName: string | undefined;
     state: string | undefined;
   };
+  fieldType: string;
```
braxex, line 27: Warning: Passing through fieldType here will break your implementation anytime you're passing through more than one field for autosave at a time (i.e. when nested child fields are being blanked). That's why we have all field-specific data in the `FieldInfo` type object that comes through. If we decide we need to pass field type through, we should do it for each field that's being saved.

braxex:
@ntsummers1 Closing for now.

PR #5369, opened 2023-01-31
Export PDF Feedback

karla-vm:
## Description

A couple of UI feedback items from Design:

* Fixed a bug where the count of entities was showing two (2) zeros if there were no entities
* Entity responses were too far apart from each other, now they're closer together to show they're a single item
* Nested date range on a couple of cards wasn't showing the original response
* `Not Answered / Not Applicable` -> `Not answered / Not applicable`

### How to test

Create a PDF, check that it's all good.

### Changed Dependencies

N/A

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.tsx
+++ b/services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.tsx
@@ -99,7 +99,11 @@ export const EntityCardTopSection = ({
                   ? "D2.VII.7a Reporting Period and D2.VII.7b Reporting period: Date range"
                   : "Measure Reporting Period"}
               </Text>
-              <Text sx={sx.subtext}>{formattedEntityData.reportingPeriod}</Text>
+              <Text sx={sx.subtext}>
+                {formattedEntityData.reportingPeriod === "Yes"
+                  ? formattedEntityData.reportingPeriod
+                  : `No, ${formattedEntityData.reportingPeriod}`}
+              </Text>
             </GridItem>
           </Grid>
           <Text sx={sx.subtitle}>
--- a/services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.tsx
+++ b/services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.tsx
@@ -99,11 +99,7 @@ export const EntityCardTopSection = ({
                   ? "D2.VII.7a Reporting Period and D2.VII.7b Reporting period: Date range"
                   : "Measure Reporting Period"}
               </Text>
-              <Text sx={sx.subtext}>
-                {formattedEntityData.reportingPeriod === "Yes"
-                  ? formattedEntityData.reportingPeriod
-                  : `No, ${formattedEntityData.reportingPeriod}`}
-              </Text>
+              <Text sx={sx.subtext}>{formattedEntityData.reportingPeriod}</Text>
             </GridItem>
           </Grid>
           <Text sx={sx.subtitle}>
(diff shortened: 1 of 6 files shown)


scwambach (approved):
lgtm! 👍

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
```diff
@@ -18,14 +18,15 @@ export const ExportedModalDrawerReportSection = ({
   const { report } = useContext(ReportContext);
   const { emptyEntityMessage } = exportVerbiage;
   const entities = report?.fieldData?.[entityType];
-  const entityCount = entities?.length;
+  const entityCount = entities ? entities?.length : undefined;
```
braxex, line 21: Note: Hmm... I feel like `entities?.length` should return either the length count or undefined by default if `entities` is not defined. Is this ternary really necessary? Or could it just be the original? or maybe `entities?.length || 0` or `entities?.length || undefined`? I'm in a meeting so not running this locally lol.

services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.tsx
```diff
@@ -99,7 +99,11 @@ export const EntityCardTopSection = ({
                   ? "D2.VII.7a Reporting Period and D2.VII.7b Reporting period: Date range"
                   : "Measure Reporting Period"}
               </Text>
-              <Text sx={sx.subtext}>{formattedEntityData.reportingPeriod}</Text>
+              <Text sx={sx.subtext}>
+                {formattedEntityData.reportingPeriod === "Yes"
+                  ? formattedEntityData.reportingPeriod
+                  : `No, ${formattedEntityData.reportingPeriod}`}
```
braxex, line 105: Note: I would change this where it's being formatted initially here instead: https://github.com/CMSgov/mdct-mcr/blob/0bc7d19449faa3737ff21e3341a5c73e58f6cdcf/services/ui-src/src/utils/reports/entities.ts#L26-L30

services/ui-src/src/utils/other/export.tsx
```diff
@@ -208,6 +208,9 @@ const sx = {
   },
   entityBox: {
     marginBottom: "1rem",
+    "td, p": {
+      marginBottom: "0.25rem",
+    },
```
braxex, line 213: Note: Design requested that these be wrapped in a list item, but we can wait until we get pdf accessibility feedback on that to see if it will be necessary.

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
```diff
@@ -18,14 +18,15 @@ export const ExportedModalDrawerReportSection = ({
...
 
   return (
     <Box mt="2rem" data-testid="exportedModalDrawerReportSection">
       <Heading as="h3" sx={sx.dashboardTitle} data-testid="headerCount">
         {verbiage.dashboardTitle} {entityCount}
         {!entityCount && (
           <Text as="span" sx={sx.notAnswered} data-testid="entityMessage">
+            <br />
```
braxex, line 29: Question: What does this do? Would it be better to do it via styling instead of a line break? Line breaks can be finicky on mobile sizes and whatnot.

braxex (changes requested):

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
karla-vm: I'm using CSS now, but Design asked for the "not answered" texts to be in a new line.

services/ui-src/src/utils/other/export.tsx
```diff
@@ -48,15 +48,21 @@ export const renderDrawerDataCell = (
...
-        <Text sx={sx.entityName}>{entity.name}</Text>
-        {renderResponseData(
-          formField,
-          fieldResponseData,
-          entityResponseData,
-          pageType,
-          entityIndex,
-          notApplicable
-        )}
+        <ul>
```
braxex, line 51: Info: There's a Chakra way to do this too! See [Chakra List here](https://chakra-ui.com/docs/components/list). If we did it this way we could use `sx` prop for styling and I think also pass `"none"` as the `styleType`

braxex (approved):
Looking good! 😎

PR #5370, opened 2023-02-01
for nick

braxex:
## Code author checklist

### DIFF
--- a/services/ui-src/src/components/fields/ChoiceListField.tsx
+++ b/services/ui-src/src/components/fields/ChoiceListField.tsx
@@ -44,14 +44,14 @@ export const ChoiceListField = ({
   const shouldDisableChildFields = !!props?.disabled;
 
   // set initial display value to form state field value or hydration value
-  const hydrationValue = props?.hydrate;
+  const hydrationValue = props?.hydrate || defaultValue;
   useEffect(() => {
     // if form state has value for field, set as display value
     const fieldValue = form.getValues(name);
     if (fieldValue) {
       setDisplayValue(fieldValue);
     }
-    // else if hydration value exists, set as display value
+    // else set hydrationValue or defaultValue as display value
     else if (hydrationValue) {
       setDisplayValue(hydrationValue);
       form.setValue(name, hydrationValue, { shouldValidate: true });
--- a/services/ui-src/src/components/fields/DateField.tsx
+++ b/services/ui-src/src/components/fields/DateField.tsx
@@ -59,6 +59,9 @@ export const DateField = ({
   // if should autosave, submit field data to database on blur
   const onBlurHandler = async (event: InputChangeEvent) => {
     const { name, value } = event.target;
+    // if blanking field, trigger client-side field validation error
+    if (value === defaultValue) form.trigger(name);
+    // submit field data to database
     if (autosave) {
       const fields = [{ name, value, hydrationValue, defaultValue }];
       const reportArgs = { id: report?.id, updateReport };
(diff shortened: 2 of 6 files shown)


services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -44,14 +44,14 @@ export const ChoiceListField = ({
   const shouldDisableChildFields = !!props?.disabled;
 
   // set initial display value to form state field value or hydration value
-  const hydrationValue = props?.hydrate;
+  const hydrationValue = props?.hydrate || defaultValue;
```
braxex, line 47: Note: This is the central idea behind the easier check to stop the initial "every time you blur it triggers" thing. Basically, if `props?.hydrate` is undefined, i.e. the field has never been saved before, set hydrationValue to defaultValue. This way we don't need the more complex checking logic for most field types (except for dynamic)

services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -59,6 +59,9 @@ export const DateField = ({
   // if should autosave, submit field data to database on blur
   const onBlurHandler = async (event: InputChangeEvent) => {
     const { name, value } = event.target;
+    // if blanking field, trigger client-side field validation error
+    if (value === defaultValue) form.trigger(name);
```
braxex, line 63: Note: This makes sure we run the client-side validation on the blurred field if it's being blanked. Without it, the field doesn't get the red warning error when blank it telling you that you've been bad. But we want that error to show up to tell people to not do that. We still let them save it because we revert the field to the default value and then the server-side validation lets it through, but we want the client-side validation to yell at them.

braxex: _same change across multiple fields_

services/ui-src/src/components/fields/ChoiceListField.tsx
braxex: _same change across multiple fields_

services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -28,14 +28,14 @@ export const TextField = ({
   form.register(name);
 
   // set initial display value to form state field value or hydration value
-  const hydrationValue = props?.hydrate;
+  const hydrationValue = props?.hydrate || defaultValue;
   useEffect(() => {
     // if form state has value for field, set as display value
     const fieldValue = form.getValues(name);
-    if (fieldValue || fieldValue == "") {
```
braxex, line 35: Note: We're not doing this on other string fields without issue.

braxex, line 35: Note: I don't see any reason why we're doing this, especially since the other string-based fields don't use it. Can we remove it?

services/ui-src/src/utils/validation/schemas.ts
```diff
@@ -10,24 +10,16 @@ import { validationErrors as error } from "verbiage/errors";
 import { Choice } from "types";
 
 // TEXT - Helpers
-const testForEmptyValue = (value: any) => {
```
braxex, line 13: Note: This method has a bug in it where when you have a value that is equal to `""`, it gets caught by line 14 because `""` is falsy. This should work better now.

services/ui-src/src/components/fields/ChoiceListField.tsx
ntsummers1: Gave this a run. For whatever reason this is fine when I try it in Datefield, DropdownField, Numberfield and TextField but Choicelist HARD breaks with this change. The page will freeze up and become unresponsive. I suspect the useEffect gets triggered over and over again?

braxex:
@ntsummers1 Closing for now.

PR #5441, opened 2023-02-06
[MDCT-2222] - Spruce up footer to be inline with other apps

ntsummers1:
## Description
The MCR Footer isn't quite inline with the other applications. This changes it so that its more or less the same as how Carts does it.

### How to test
./dev local
Look at the bottom of the page and check it out!

## Code author checklist

### DIFF
--- a/services/ui-src/src/components/layout/Footer.tsx
+++ b/services/ui-src/src/components/layout/Footer.tsx
@@ -26,7 +26,7 @@ export const Footer = () => {
               <Flex sx={sx.footerTopRightTopFlex}>
                 <Box sx={sx.footerCMSBrandingLeft}>
                   <Box sx={sx.hhsLogo}>
-                    <img
+                    <Image
                       src={hhsLogo}
                       alt="Department of Health and Human Services, USA"
                     />
@@ -34,7 +34,7 @@ export const Footer = () => {
 
                   {!isDesktop && (
                     <Box sx={sx.hhsMedicaidLogoMobile}>
-                      <img
+                      <Image
                         src={medicaidLogo}
                         alt="Medicaid.gov: Keeping America Healthy"
                       />
@@ -151,7 +151,6 @@ const sx = {
     maxWidth: "100%",
     padding: "0 1rem",
     minWidth: "100%",
-    width: "100%",
     marginBottom: "1rem",
     ".desktop &": {
       marginBottom: 0,
@@ -185,7 +184,7 @@ const sx = {
     display: "inline-block",
     textAlign: "left",
     verticalAlign: "top",
-    fontSize: 14,
+    fontSize: "sm",
     lineHeight: "21px",
     marginTop: 0,
   },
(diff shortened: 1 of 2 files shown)


services/ui-src/src/components/layout/Footer.tsx
```diff
@@ -9,48 +9,57 @@ import mcrLogo from "assets/logos/logo_mcr_footer.png";
...
-                    <Image src={hhsLogo} alt="HHS logo" sx={sx.hhsLogo} />
+            </Box>
+            <Box sx={sx.footerTopRightContainer}>
+              <Flex sx={sx.footerTopRightTopFlex}>
+                <Box sx={sx.footerCMSBrandingLeft}>
+                  <Box sx={sx.hhsLogo}>
+                    <img
+                      src={hhsLogo}
+                      alt="Department of Health and Human Services, USA"
+                    />
```
braxex, line 32: Suggestion: Recommend using Chakra's `Image` component here and elsewhere in PR for consistency.

services/ui-src/src/components/layout/Footer.tsx
```diff
@@ -105,57 +114,93 @@ const sx = {
...
-  footerTopLogoFlex: {
-    paddingRight: "1rem",
-    ".mobile &": {
-      minWidth: "200px",
-      paddingRight: 0,
-      justifyContent: "center",
+  footerTopLeftContainer: {
+    flex: "0 0 50%",
+    maxWidth: "100%",
+    minWidth: "100%",
```
braxex, line 122: Question: Could we just use `width: "100%"` here? Same question with other container below.

services/ui-src/src/components/layout/Footer.tsx
```diff
@@ -105,57 +114,93 @@ const sx = {
...
-    ".mobile &": {
-      maxWidth: "10.125rem",
+  footerCMSBrandingLeft: {
+    display: "flex",
+    justifyContent: "center",
+    alignItems: "center",
+    maxWidth: "100%",
+    padding: "0 1rem",
+    minWidth: "100%",
+    width: "100%",
```
braxex, line 154: Question: Seems unnecessary to set all these widths the same. Can probably just set `width` to 100 and then auto.

services/ui-src/src/components/layout/Footer.tsx
```diff
@@ -105,57 +114,93 @@ const sx = {
...
-    marginY: "1.75rem",
-    justifyContent: "center",
+    maxWidth: "12.5rem",
+    marginLeft: "1.25rem",
+  },
+  hhsCopyText: {
+    display: "inline-block",
+    textAlign: "left",
+    verticalAlign: "top",
+    fontSize: 14,
```
braxex, line 188: Nit: Use theme value `"sm"` here.

braxex (approved):
Overall looks great. A couple nits and questions, but nothing blocking.

services/ui-src/src/components/layout/Footer.tsx
ntsummers1: Can do away with width I believe. Need minWidth and maxWidth though for how diff browsers can handle it. minWidth in particular though is the true one we could keep

braxex (approved):

PR #7469, opened 2023-02-09
MDCT-2241: Refactor Hardcoded MCPAR to Accept MLR

karla-vm:
## Description

Kicks off [MDCT-2241](https://qmacbis.atlassian.net/browse/MDCT-2241) effort and covers [MDCT-2214](https://qmacbis.atlassian.net/browse/MDCT-2214).
 
As we're adding `MLR` to this code base, we need to ensure that anything specific to `MCPAR` is just a little bit more flexible 💯 

### How to test

Nothing should really change.

### Changed Dependencies

N/A

## Code author checklist


services/app-api/handlers/reports/archive.ts
```diff
@@ -15,14 +15,19 @@ export const archiveReport = handler(async (event, context) => {
...
     delete currentReport.formTemplate;
 
     // toggle archived status in report metadata table
     const reportMetadataParams = {
-      TableName: process.env.MCPAR_REPORT_TABLE_NAME!,
+      // TODO: chain future reports here
+      TableName:
+        reportType === "MCPAR"
+          ? process.env.MCPAR_REPORT_TABLE_NAME!
+          : process.env.MLR_REPORT_TABLE_NAME!,
```
braxex, line 30: Suggestion: I'd probably make a global object or enumerator to hold the report table names, like:
```
const reportTables = {
  MCPAR: process.env.MCPAR_REPORT_TABLE_NAME!,
  MLR: process.env.MLR_REPORT_TABLE_NAME!,  
  ...
}
```

Then you can just import that global constant and reference what you want from it as keys on an object, like:
`TableName: reportTables[reportType],`

services/app-api/handlers/reports/archive.ts
```diff
@@ -15,14 +15,19 @@ export const archiveReport = handler(async (event, context) => {
   if (getCurrentReport?.body) {
     const currentReport = JSON.parse(getCurrentReport.body);
     const currentArchivedStatus = currentReport?.archived;
+    const reportType = currentReport.reportType;
```
braxex, line 18:
```suggestion
    const reportType = currentReport?.reportType;
```

services/app-api/handlers/reports/archive.ts
braxex: _Applies to all other similar areas with report tables and form buckets below_

braxex (changes requested):

services/app-api/utils/constants/constants.ts
```diff
@@ -20,3 +20,15 @@ export const buckets = {
...
 };
+
+// REPORTS
+
+export const reportTables = {
+  MCPAR: process.env.MCPAR_REPORT_TABLE_NAME!,
+  MLR: process.env.MLR_REPORT_TABLE_NAME!,
+};
+
+export const reportFormBuckets = {
```
braxex, line 31: Nit: Would probably name either `reportBuckets` to match `reportTables` or `formBuckets` to match the env variable. Not sure why the env variable is named that but 🤷.

services/app-api/handlers/reports/create.ts
```diff
@@ -44,15 +51,17 @@ export const createReport = handler(async (event, _context) => {
       if (validatedFieldData) {
         // post validated field data to s3 bucket
         const fieldDataParams: S3Put = {
-          Bucket: process.env.MCPAR_FORM_BUCKET!,
+          Bucket:
+            reportFormBuckets[reportType as keyof typeof reportFormBuckets],
```
braxex, line 55: Suggestion: For all the places across `create`, `archive`, `update`, and `fetch` where we're setting the `TableName:` or `Bucket:` param multiple times, we could get the report or bucket once at the top of the file, set it as a const, then reference it below instead of accessing that constant multiple times. Just makes it a bit cleaner.

_Applies to all similar instances below_

services/app-api/handlers/reports/fetch.ts
```diff
@@ -14,6 +14,7 @@ export const fetchReport = handler(async (event, _context) => {
 
   // get current report metadata
   const reportMetadataParams = {
+    // TODO: chain other report types
```
braxex, line 17: Blocking: Looks like we still need to make the change in this file like in the others.

braxex (changes requested):

services/app-api/handlers/reports/fetch.ts
karla-vm: I believe this one is out of scope since we currently don't have a way to determine the `reportType` on fetch, and I think we'd have to update the API calls for that to happen.

scwambach (dismissed):

scwambach (approved):

braxex (approved):

BearHanded (approved):

karla-vm:
Closing because this `deploy` is broken, refer to [this](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/11075#pullrequestreview-1308296634) PR instead.

PR #8212, opened 2023-02-10
MDCT-2213: Add "Enter MLR" Button to Home Page

karla-vm:
## Description

Closes [MDCT-2213](https://qmacbis.atlassian.net/browse/MDCT-2213).

Adding a new button `Enter MLR` to the MCR homepage and hiding the "available" verbiage behind a feature flag until we're ready to release it.

### How to test

Login as a state user to the deployed instance [here](https://d15iwqrff5qhla.cloudfront.net/).

Check that the `Enter MLR` button is visible on screen! Then ask me to toggle the flag in `LaunchDarkly` for a disappearing act 🪄 

### Changed Dependencies

N/A

## Code author checklist

- [x] I have done the lgtm review


services/ui-src/src/components/cards/TemplateCard.test.tsx
```diff
@@ -23,26 +23,38 @@ jest.mock("react-router-dom", () => ({
...
+    render(mcparTemplateCardComponent);
   });
 
   test("TemplateCard is visible", () => {
-    expect(screen.getByText(templateVerbiage.title)).toBeVisible();
+    expect(screen.getByText(mcparTemplateVerbiage.title)).toBeVisible();
   });
 
   test("TemplateCard download button is visible and clickable", async () => {
     const apiSpy = jest.spyOn(mockAPI, "getSignedTemplateUrl");
```
britt-mo, line 56: I read that const as `aSpicy` and was like...what....

britt-mo (dismissed):

services/ui-src/src/components/cards/TemplateCard.test.tsx
karla-vm: 🤣

services/ui-src/src/components/cards/TemplateCard.tsx
```diff
@@ -44,8 +47,12 @@ export const TemplateCard = ({
         )}
         <Flex sx={sx.cardContentFlex}>
           <Heading sx={sx.cardTitleText}>{verbiage.title}</Heading>
-          <Text>{cardText}</Text>
-
+          {/* TODO: Remove LD flag once MLR is released */}
+          {!mlrReport && templateName === "MLR" ? (
+            <Text>{verbiage.body.unavailable}</Text>
+          ) : (
+            <Text>{cardText}</Text>
```
braxex, line 54: Thought: It seems like we're splitting the logic kind of oddly here. `cardText` is set above and considers if the verbiage link is available or not. Maybe we should make a new variable on line 34 that does the check currently on line 51, then combine it with verbiage.link check on 35 to get the cardText. That new variable could be referenced elsewhere and cleans this up a bit.

services/ui-src/src/components/cards/TemplateCard.tsx
```diff
@@ -60,17 +67,20 @@ export const TemplateCard = ({
...
+              (!mlrReport && templateName === "MLR" ? (
+                <></>
+              ) : (
+                <Button
+                  sx={sx.formLink}
+                  onClick={() => navigate(verbiage.link.route)}
+                  rightIcon={
+                    <Image src={nextIcon} alt="Link Icon" height="1rem" />
+                  }
+                >
```
braxex, line 80: Request: Just use a chained `&&` here, like `condition && <Button />` so the button will only show if the condition is true.

services/ui-src/src/components/cards/TemplateCard.tsx
braxex: The const we set at the top could be an object, like: 
```
const enabledReports = {
  MCPAR: // logic for MCPAR enabled here,
  MLR: // logic for MLR enabled here, which is likely just mlrReportEnabled,
}
```

then the check for each one below is no longer `templateName === "MLR"`, it becomes: 
`enabledReports[templateName]` -- and if that's `true`, you're good to go.

braxex: Happy to walk through any/all of this. 👍

braxex (changes requested):
Looking good. A few requested changes. Happy to walk through it with you if my notes don't make sense.

tblackwe (approved):
LGTM

braxex (approved):

PR #11075, opened 2023-02-16
MDCT-2241 + MDCT-2214: MLR Kickoff

karla-vm:
## Description

Kicks off [MDCT-2241](https://qmacbis.atlassian.net/browse/MDCT-2241) effort and covers [MDCT-2214](https://qmacbis.atlassian.net/browse/MDCT-2214).
 
As we're adding `MLR` to this code base, we need to ensure that anything specific to `MCPAR` is just a little bit more flexible 💯 by doing the following:

* Adding a new path parameter `reportType` to the report API calls
* Creating a new utility to ensure all of the required path parameters are provided

### How to test

Nothing should really change.

### Changed Dependencies

N/A

## Code author checklist


tblackwe:
@ntsummers1 @karla-vm autosave code needs to be updated in this branch to leverage the reportType path. Currently cypress tests are failing because of `undefined` appearing in the path.

tblackwe (approved):
LGTM, ty for fixing autosave so quickly.

services/app-api/handlers/reports/fetch.ts
```diff
@@ -2,19 +2,33 @@ import handler from "../handler-lib";
...
+} from "../../utils/constants/constants";
 
 export const fetchReport = handler(async (event, _context) => {
   let status, body;
-  if (!event?.pathParameters?.state! || !event?.pathParameters?.id!) {
+  if (
+    !event?.pathParameters?.reportType! ||
+    !event?.pathParameters?.state! ||
+    !event?.pathParameters?.id!
+  ) {
```
ntsummers1, line 18: Since codeclimate is being mad about this and now that theres a few reasons to throw an error, maybe we pull this out to a function or something so this can read like "If (!validReport) throw new Error(error.NO_KEY)"

services/ui-src/src/components/pages/ReviewSubmit/McparReviewSubmitPage.tsx
```diff
@@ -96,6 +99,7 @@ export const McparReviewSubmitPage = () => {
 const PrintButton = () => {
   const { print } = reviewVerbiage;
   return (
+    // TODO: make the path route to the correct report type (in the future)
```
ntsummers1, line 102: Do we need to worry about this atm? And if we want to push this down the line is this linked to a JIRA ticket so we don't forget?

services/app-api/handlers/reports/fetch.ts
karla-vm: Good call! I added a method under `validation.ts` for this.

services/ui-src/src/components/pages/ReviewSubmit/McparReviewSubmitPage.tsx
karla-vm: This is actually no longer a to-do list item, we're going to be creating a separate Review and Submit page for MLR 💯

services/app-api/utils/validation/validation.ts
```diff
@@ -2,6 +2,22 @@ import * as yup from "yup";
 import { nested, endDate, schemaMap } from "./schemaMap";
 import { AnyObject } from "../types/types";
 import { error } from "../constants/constants";
+import { APIGatewayEvent } from "aws-lambda";
+
+// validate report keys
```
braxex, line 7: Suggestion: Since this isn't actually related to validation, we should move it to another utils file. I would either put it in utils under `dynamo` or make a new dir `handlers` and put it in a file under there.

services/app-api/utils/validation/validation.ts
```diff
@@ -2,6 +2,22 @@ import * as yup from "yup";
...
+export const validReport = (
```
braxex, line 8: Suggestion: When referenced in other files, `validReport` may seem like it's an actual report instead of a boolean representing whether an event has all necessary path parameters. Maybe we could rename `checkEventParams` or `hasEventParams` or something to be more clear about what this check is.

braxex: Note: It's also worth noting that this is only relevant for `/report/` endpoints, and will not work for any other API handler.

services/app-api/utils/validation/validation.ts
```diff
@@ -2,6 +2,22 @@ import * as yup from "yup";
...
+  event: APIGatewayEvent,
+  apiMethod: string | undefined
+) => {
+  const pathParameters = event.pathParameters!;
+  if (apiMethod === "create") {
```
braxex, line 13: Thought: This totally works. We could also do something by making it agnostic to what kind of handler it is and what parameters are required by passing them in (think an array of strings that we pass in, called `requiredParams`).  That array, say `["id", "state", "reportType"]` or `["id", "state"]`, could then be checked to make sure it exists. Something like (with silly types so it highlights the syntax lol):

```js
export const hasRequiredPathParams = ( eventParams: AnyObject, requiredParams: ArrayOfStrings ) => 
requiredParams.every((requiredParam: SingleString) => eventParams[requiredParam]);

// then calling it would look like this
const requiredParams = ["id", "state", "reportType"];
if (!hasRequiredPathParams(event.pathParameters, requiredParams) {
  // fail hard
} 
// continue!
```

services/app-api/utils/validation/validation.ts
```diff
@@ -2,6 +2,22 @@ import * as yup from "yup";
...
+    return pathParameters.reportType && pathParameters.state;
+  } else {
+    return (
+      pathParameters.reportType && pathParameters.state && pathParameters.id
+    );
```
braxex, line 18: Nit: A quick note on this pattern. No need to do `else {}` when using an `if → return` pattern, because if it got caught in the `if`, it will already have returned, so you can just return without the `else`, like:

```js
const boolChecker = () => {
  if (whatever) {
    return true;
  }
  return false;
}
```

services/app-api/handlers/reports/update.ts
```diff
@@ -38,22 +44,27 @@ export const updateReport = handler(async (event, context) => {
     // if current report exists, get formTemplateId and fieldDataId
     if (getCurrentReport?.body) {
       const currentReport = JSON.parse(getCurrentReport.body);
-      const { formTemplateId, fieldDataId } = currentReport;
+      const { formTemplateId, fieldDataId, reportType } = currentReport;
       if (formTemplateId && fieldDataId) {
         // if report not in archived state, proceed with updates
         if (!currentReport.archived) {
-          const state: string = event.pathParameters.state;
+          const state: string = event.pathParameters?.state!;
```
braxex, line 51: Nit: By the time we get here, `event.pathParameters.state` will always be defined. We used to be checking it above on line 24, but now that we're not, TS is mad, so it makes sense why we'd use `!` here to tell it so. But I think we don't need to use `?` for optional chaining here, because `event.pathParameters` will always be defined too. `event.pathParameters.state!` should work.

braxex (changes requested):

services/app-api/utils/validation/validation.ts
karla-vm: Good call, I moved this to its own utility under `dynamo` and that makes way more sense.

karla-vm:
@braxex updated the description a bit.

braxex (dismissed):

scripts/deploy-support.sh
```diff
@@ -22,7 +22,7 @@ deploy() {
   service=$1
   pushd services/$service
   install_deps
-  serverless deploy  --stage $stage
+  serverless deploy  --stage $stage --enforce-hash-update
```
BearHanded, line 25: I think we don't actually need this, based on the other stuff we wound up digging into with the API endpoints.

scripts/deploy.sh
```diff
@@ -27,7 +27,7 @@ deploy() {
   service=$1
   pushd services/$service
   install_deps
-  serverless deploy  --stage $stage
+  serverless deploy  --stage $stage --enforce-hash-update
```
BearHanded, line 30: same here

BearHanded (changes requested):

ntsummers1 (dismissed):

services/app-api/serverless.yml
```diff
@@ -165,33 +165,35 @@ functions:
     handler: handlers/reports/fetch.fetchReport
     events:
       - http:
-          path: reports/{state}/{id}
+          path: reports/{reportType}/{state}/{id}
```
BearHanded, line 168: lets go `/reporting/`!

BearHanded (changes requested):
Update endpoints with a tempor

PR #11080, opened 2023-02-21
Autosave txtfix

ntsummers1:
### Description

Kim found this bug here:

https://user-images.githubusercontent.com/19439679/220405724-8e836664-3470-4596-bf90-3ca674900ea1.mp4

This PR fixes this!
https://user-images.githubusercontent.com/19439679/220407256-598ee713-6665-4224-aa19-7a44b39e8c0e.mov

### How to test

- ./dev local
- Create a new program for MCR
- Navigate to anything with a textbox, particularly one that is nested in a choicefield
- fill it out
- navigate to another page
- come back and see that it captured your input correctly!

### Author checklist

### DIFF
--- a/services/ui-src/src/components/fields/TextField.tsx
+++ b/services/ui-src/src/components/fields/TextField.tsx
@@ -39,7 +39,7 @@ export const TextField = ({
     // else set hydrationValue or defaultValue as display value
     else if (hydrationValue) {
       setDisplayValue(hydrationValue);
-      form.setValue(name, hydrationValue);
+      form.setValue(name, hydrationValue, { shouldValidate: true });
     }
   }, [hydrationValue]); // only runs on hydrationValue fetch/update
 
--- a/services/ui-src/src/components/fields/TextField.tsx
+++ b/services/ui-src/src/components/fields/TextField.tsx
@@ -34,7 +34,6 @@ export const TextField = ({
     const fieldValue = form.getValues(name);
     if (fieldValue) {
       setDisplayValue(fieldValue);
-      form.setValue(name, fieldValue);
     }
     // else set hydrationValue or defaultValue as display value
     else if (hydrationValue) {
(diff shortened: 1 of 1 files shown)


tblackwe (approved):

services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -34,11 +34,12 @@ export const TextField = ({
     const fieldValue = form.getValues(name);
     if (fieldValue) {
       setDisplayValue(fieldValue);
+      form.setValue(name, fieldValue);
```
braxex, line 37: Question: I don't think this would be necessary because on line 34 we get the field value from the form and then here we set the field value in the form to the value we just got from it. I'm guessing the fix still works without this?

services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -34,11 +34,12 @@ export const TextField = ({
...
     }
     // else set hydrationValue or defaultValue as display value
     else if (hydrationValue) {
       setDisplayValue(hydrationValue);
-      form.setValue(name, hydrationValue);
+      form.setValue(name, hydrationValue, { shouldValidate: true });
```
braxex, line 42: Comment: Assuming this is the fix -- it seems like the issue is actually just that the form was not revalidating the field after it's hydrated, which makes sense.

ntsummers1: Yup! Real easy one

services/ui-src/src/components/fields/TextField.tsx
ntsummers1: Still works. Initial fix of adding that did sometimes fix the issue but you are right in that really we just need the other line.

braxex (approved):
Pulled it down, ran it. Everything looks good. Glad to see it was such a simple fix! Thanks for working it so quickly.

tblackwe (approved):

PR #11082, opened 2023-02-21
Clean up reporting APIs.

oweingrod-coforma:
## Summary

This PR cleans up the MCR report CRUD apis to handle errors more cleanly and expose more useful information about errors to the client.

One somewhat breaking change is made—routes will now return a `400 BAD REQUEST` status when required arguments are missing from path parameters. Previously, they returned `500 INTERNAL SERVER ERROR` responses which felt undescriptive.

### Related ticket
N/A

### How to test
Additional manual testing likely unnecessary—all existing tests pass or have been updated.

### Important updates
N/A

### Author checklist

- [X] I have performed a self-review of my code
- [X] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary
- [X] I have updated the documentation, if necessary

### DIFF
--- a/services/app-api/handlers/reports/update.ts
+++ b/services/app-api/handlers/reports/update.ts
@@ -1,7 +1,5 @@
 import handler from "../handler-lib";
 import { fetchReport } from "./fetch";
-import { archiveReport } from "./archive";
-// utils
 import dynamoDb from "../../utils/dynamo/dynamodb-lib";
 import { hasPermissions } from "../../utils/auth/authorization";
 import s3Lib from "../../utils/s3/s3-lib";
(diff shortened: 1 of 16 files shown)


services/app-api/handlers/reports/update.ts
```diff
@@ -14,132 +12,185 @@ import { StatusCodes, UserRoles } from "../../utils/types/types";
...
+   * TODO: Is this code necessary given the discrete archive route?
+   * if (hasPermissions(event, [UserRoles.ADMIN])) {
+   *  const { statusCode: archiveStatus, body: archiveBody } =
+   *    await archiveReport(event, context);
+   *  return {
+   *    status: archiveStatus,
+   *    body: JSON.parse(archiveBody),
+   *  };
+   * }
+   */
```
braxex, line 48: Not as such, but we will to check if the user is admin in the archive handler and throw unauthorized if not, similar to how we check permissions at the beginning of the others. 👍

BearHanded (approved):
Much cleaner 👍

services/app-api/handlers/reports/update.ts
oweingrod-coforma: Removed this comment block. Thanks @braxex

braxex (changes requested):
Two things:
- We need to add the authorization check into the archive handler so it checks if the user is an admin.
- Since the update handler is no longer conditionally triggering archive/unarchive, we need to change the archive/unarchive button in the frontend to not hit the update endpoint, and instead hit the archive endpoint. This will require some serverless work and setting up a new frontend handler to do some fun stuff for us.

oweingrod-coforma:
> Two things:
> 
> * We need to add the authorization check into the archive handler so it checks if the user is an admin.
> * Since the update handler is no longer conditionally triggering archive/unarchive, we need to change the archive/unarchive button in the frontend to not hit the update endpoint, and instead hit the archive endpoint. This will require some serverless work and setting up a new frontend handler to do some fun stuff for us.

Yay, more tasks!! :)

oweingrod-coforma:
@braxex @BearHanded added check to permissions & tests, but can someone double check the mocks to make sure I didn't mess with the existing tests too much? Wasn't sure how to specifically test for admin credentials.

oweingrod-coforma:
@braxex See my last three commits for fixes for the frontend, as well as a new handler.

PR #11094, opened 2023-02-24
Nested Choicelist Fix and saving on continue

ntsummers1:
### Description

Autosave has been updated with quite a few new fixes:

- Hitting enter on the form will autosave the formpage before navigating you to the next page
- Validation will no longer stop you form continuing to the next page
- Choicelists and their children have been revamped to clear any choices or text boxes underneath them if you uncheck them
- Validation on Choicelists has been slightly relaxed to not be so angry all the time.

### How to test

./dev local
Sign in as a state user
Create a report
Open report
Start filling out fields! Pay particular attention to checkboxes and radiofields that have nested options under them (Like page B3!)

### Author checklist

### DIFF
--- a/services/ui-src/src/components/fields/TextField.tsx
+++ b/services/ui-src/src/components/fields/TextField.tsx
@@ -29,6 +29,7 @@ export const TextField = ({
 
   // set initial display value to form state field value or hydration value
   const hydrationValue = props?.hydrate || defaultValue;
+
   useEffect(() => {
     // if form state has value for field, set as display value
     const fieldValue = form.getValues(name);
@@ -37,8 +38,13 @@ export const TextField = ({
     }
     // else set hydrationValue or defaultValue as display value
     else if (hydrationValue) {
-      setDisplayValue(hydrationValue);
-      form.setValue(name, hydrationValue, { shouldValidate: true });
+      if (props.clear) {
+        setDisplayValue(defaultValue);
+        form.setValue(name, defaultValue);
+      } else {
+        setDisplayValue(hydrationValue);
+        form.setValue(name, hydrationValue, { shouldValidate: true });
+      }
     }
   }, [hydrationValue]); // only runs on hydrationValue fetch/update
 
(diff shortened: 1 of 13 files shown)


services/ui-src/src/components/forms/Form.tsx
```diff
@@ -70,7 +71,10 @@ export const Form = ({
     <FormProvider {...form}>
       <form
         id={id}
-        onSubmit={form.handleSubmit(onSubmit as any, onErrorHandler)}
+        onSubmit={form.handleSubmit(
+          onSubmit as any,
+          onError || (onErrorHandler as any)
```
braxex, line 76: Nit: Could tighten up these types. Though I know they are the internal RHF types, and the error handler one will probably get mad.

services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -37,8 +38,13 @@ export const TextField = ({
...
     // else set hydrationValue or defaultValue as display value
     else if (hydrationValue) {
-      setDisplayValue(hydrationValue);
-      form.setValue(name, hydrationValue, { shouldValidate: true });
+      if (props.clear) {
+        setDisplayValue(defaultValue);
+        form.setValue(name, defaultValue);
+      } else {
+        setDisplayValue(hydrationValue);
+        form.setValue(name, hydrationValue, { shouldValidate: true });
```
braxex, line 46: Question: This is the only one of the other spots where we do this that we pass `shouldValidate: true`. Is there a reason for that?

braxex (commented):
LGTM. Just a couple questions. Since I was somewhat involved in this I'll hold off on an actual approval and just leave my comments.

services/ui-src/src/components/fields/TextField.tsx
karla-vm: I'm not sure I follow your question, I see this on other field types.

karla-vm (dismissed):
LGTM! 
Got a couple of questions that need resolving from @braxex but otherwise looking great ⭐

tblackwe (approved):
LGTM

braxex (approved):

PR #11101, opened 2023-03-01
Add (optional) styling to fields that are not required 

britt-mo:
## Add `(optional)` styling to fields that are not required
[jira ticket](https://qmacbis.atlassian.net/jira/software/c/projects/MDCT/boards/232?modal=detail&selectedIssue=MDCT-2237)

## Summary
Added '(optional)' text to fields. New attribute `styleAsOptional` can be added to mcpar.json to add `(optional)` at the end of the title for a field.

### How to test
Add  `"styleAsOptional": true` in the `props:` object of any field except dynamic fields (should support text, number, radio/checkbox, date, dropdown, single checkbox, etc.)` and check in the browser that the targeted field's title now has `(optional)` at the end of it. This has nothing to do with the field's validation, as that is controlled by `validation` type.

### Author checklist

- [ x] I have performed a self-review of my code
- [ x] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary


services/ui-src/src/utils/other/styleAsOptional.ts
```diff
@@ -0,0 +1,4 @@
+import { parseCustomHtml } from "./parsing";
+
+export const labelTextWithOptional = (label: any) =>
```
braxex, line 3: Request: Strengthen label type to `string`.

services/ui-src/src/utils/other/styleAsOptional.ts
```diff
@@ -0,0 +1,4 @@
+import { parseCustomHtml } from "./parsing";
```
braxex, line 1: Request: Let's move this utility method out of `other/styleAsOptional` and put it in `parsing/parsing.ts` since that's what it's doing.

services/ui-src/src/forms/mcpar/mcpar.json
```diff
@@ -36,7 +36,7 @@
               {
                 "id": "contactName",
                 "type": "text",
-                "validation": "text",
+                "validation": "textOptional",
```
braxex, line 39: Blocking: Switch this back to `text` because we don't want to change it to optional for real.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -226,6 +230,9 @@ const sx = {
   ".ds-c-choice[type='checkbox']:checked::after": {
     boxSizing: "content-box",
   },
+  ".optional-text": {
+    fontWeight: "lighter",
+  },
```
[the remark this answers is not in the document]
braxex: Request: Put this style once in the styling object in `Form.tsx` so it applies to all fields where this class shows up.

services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -19,6 +19,7 @@ import {
   FormField,
   InputChangeEvent,
 } from "types";
+import { labelTextWithOptional } from "utils/other/styleAsOptional";
```
braxex, line 22: Request: Change to export util the same as all other utils through the `utils` dir, then import with the others above on lines 9-12.

braxex: _Note: Applies to all instances below._

services/ui-src/src/components/fields/ChoiceListField.tsx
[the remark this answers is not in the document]
braxex: _Note: Applies to all instances below._

services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -152,5 +158,12 @@ interface Props {
...
   sxOverride?: AnyObject;
+  styleAsOptional?: boolean;
   [key: string]: any;
 }
+
+const sx = {
+  ".optional-text": {
+    fontWeight: "lighter",
+  },
+};
```
braxex, line 169: Request: As noted above, this style should be moved to `Form.tsx` and only be written there once. Then we can remove the `sx` object here. Then you can revert the change on line 134 back to `sx={sxOverride}`.

braxex (changes requested):
I think we also need to do this work for NumberField and ChoiceField (the single checkbox).

services/ui-src/src/forms/mcpar/mcpar.json
britt-mo: good catch

britt-mo:
> I think we also need to do this work for NumberField and ChoiceField (the single checkbox).

@braxex I updated ChoiceField but NumberField uses TextField. When I added styleAsOptional to a `type: number`, I saw the '(optional)' text come up. Let me know if I'm not looking at the right thing!

services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -14,7 +14,7 @@ import {
   InputChangeEvent,
 } from "types";
 import { dropdownDefaultOptionText } from "../../constants";
-import { labelTextWithOptional } from "utils/other/styleAsOptional";
+import { labelTextWithOptional } from "utils/other/parsing";
```
braxex, line 17: Nit: Can import from `utils` with others on lines 10-14.

services/ui-src/src/utils/other/parsing.test.tsx
```diff
@@ -58,3 +95,14 @@ describe("Test parseCustomHtml", () => {
     expect(sanitizationSpy).toHaveBeenCalled();
   });
 });
+
+describe("Test labelTextWithOptional", () => {
```
braxex, line 99: Request: Right now this test is pulling in the enter component and report context, etc. and testing the utility where it's used in components. It would be better to move a testing like this to `TextField.test.tsx` where we already have all this mocked out and where we are testing the rest of the component. 

Here in `parsing.test.tsx`, we should just test the `labelTextWithOptional` method itself. We could do that by testing the input and output of the function. So something like: if you pass a label string of "test value", you get back `<span>test value <span class='optional-text'> (optional)</span></span>` or whatever comes back. 

Changing the tests like this helps keep concerns separate. The utility test file just tests the utilities and the component test file just tests the component.

services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -88,13 +92,11 @@ export const TextField = ({
   const parsedHint = hint && parseCustomHtml(hint);
   const nestedChildClasses = nested ? "nested ds-c-choice__checkedChild" : "";
   const labelClass = !label ? "no-label" : "";
-  const labelText = styleAsOptional ? labelTextWithOptional(label) : label;
+  const labelText =
+    label && styleAsOptional ? labelTextWithOptional(label) : label;
```
braxex, line 96: Question: As it currently stands, this code will make it so if there's no label, there will also be no "(optional)" on a field. That's probably what we want, but we should double-check with design since that's a bit of an edge case. Will you please reach out to design and see what they want to do if there's no label text? This happens a fair amount with things like nested text boxes under "Other, specify" and similar.

braxex (changes requested):
Looking good -- just a couple requests.

braxex:
> > I think we also need to do this work for NumberField and ChoiceField (the single checkbox).
> 
> @braxex I updated ChoiceField but NumberField uses TextField. When I added styleAsOptional to a `type: number`, I saw the '(optional)' text come up. Let me know if I'm not looking at the right thing!

Ah sure that makes sense. Ok, no worries!

braxex (approved):

services/ui-src/src/components/fields/ChoiceField.tsx
```diff
@@ -5,12 +5,14 @@ import { Choice as CmsdsChoice } from "@cmsgov/design-system";
...
 // utils
 import { AnyObject } from "types";
+import { labelTextWithOptional } from "utils";
 
 export const ChoiceField = ({
   name,
   label,
   hint,
   sxOverride,
+  styleAsOptional,
```
tblackwe, line 15: I'm sure I'm just missing it, but where is 'styleAsOptional' called? is it part of the report json?

tblackwe (approved):

services/ui-src/src/components/fields/ChoiceField.tsx
britt-mo: it's an attribute that gets added in the 'props' object in the mcpar.json

britt-mo: It's not getting used yet though so I couldn't add it.

braxex: It will be in MLR, but not yet in MCPAR.

PR #11126, opened 2023-03-08
[DRAFT] Statusing Front-end

ntsummers1:
### Author checklist


services/ui-src/src/components/pages/ReviewSubmit/McparReviewSubmitPage.tsx
```diff
@@ -54,38 +64,63 @@ export const McparReviewSubmitPage = () => {
   const submitForm = async () => {
     setSubmitting(true);
     if (isPermittedToSubmit) {
-      await submitReport(reportKeys);
+      const submissionDate = Date.now();
+      await submitReport(reportKeys, {
+        metadata: {
```
oweingrod-coforma, line 69: These changes are made in submit report now, so we don't have to add this data here

oweingrod-coforma (changes requested):
Looks good!

Only change—can we get a Cypress test to verify that these errors appear on the Review/Submit page?

tblackwe (approved):

PR #11131, opened 2023-03-09
Autosave patch - Dynamic Fields and Autocomplete

ntsummers1:
### Description

There is a duplication bug that was happening for any dynamic fields where if the user adds more then 1 item it will duplicate it. Now users can add back in as many fields as they like and it's saved appropriately

Additionally it was possible for users to autofill over multiple fields. While it showed other fields as being saved it wouldn't actually save that field if a user had not focused it. This makes it so the user can still autofill but only the field they are on

### Related ticket
Fixes [MDCT-2311](https://qmacbis.atlassian.net/browse/MDCT-2311) and [MDCT-2306](https://qmacbis.atlassian.net/browse/MDCT-2306)

#### Dynamic Field
Create a report
Go to Add a plan in the first sectiion
Add multiple plans using tabs, clicking, whatever,
Save and continue to the next page
come back
Verify everythings the same as you left it
Hard refresh
Verify everythings the same as you left it
Delete a record or 2
Hard refresh/continue away
Verify everythings the same as you left it

#### Autofill
Create a report
Go to the contact information
See that by adding a contact name it won't also add an email if you input an email as the contact name

### Author checklist

### DIFF
--- a/services/ui-src/src/utils/forms/forms.ts
+++ b/services/ui-src/src/utils/forms/forms.ts
@@ -42,7 +42,7 @@ export const formFieldFactory = (
       key: field.id,
       name: field.id,
       hydrate: field.props?.hydrate,
-      autoComplete: "one-time-code",
+      autoComplete: "one-time-code", // stops browsers from forcing autofill
       ...options,
       ...field?.props,
     };
--- a/services/ui-src/src/utils/forms/forms.ts
+++ b/services/ui-src/src/utils/forms/forms.ts
@@ -42,6 +42,7 @@ export const formFieldFactory = (
       key: field.id,
       name: field.id,
       hydrate: field.props?.hydrate,
+      autoComplete: "one-time-code",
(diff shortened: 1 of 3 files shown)


karla-vm (dismissed):

services/ui-src/src/utils/forms/forms.ts
```diff
@@ -42,6 +42,7 @@ export const formFieldFactory = (
       key: field.id,
       name: field.id,
       hydrate: field.props?.hydrate,
+      autoComplete: "one-time-code",
```
braxex, line 45: Nit: Can we add a note, like:
```suggestion
      autoComplete: "one-time-code", // stops browsers from forcing autofill
```

ntsummers1: We sure can

braxex (approved):

PR #11137, opened 2023-03-10
Number schema tweak

benmartin-coforma:
## Summary
Be a little more restrictive in what passes as a "number" in our `yup` schemas.

### Description
The only change here is a regex change, from "Must contain at least one digit, period, or comma" to "Starts with possibly a dash, then possibly a dot, then definitely a digit. Must not contain subsequent characters other than digits, periods, and commas."

This continues to allow values such as `"-.2"`, as well as obviously-incorrect values such as `"1,,,,,....2"`, but at least prevents values such as `"."`, `"2spooky4me"`, and `"Hello, world!"`. Essentially, we're upgrading from "Make sure it's a string" to "Make sure it looks kind of like a number". Stricter validation occurs later in the pipeline (I believe).

This change applies to validation on both the frontend and the backend.

### Related ticket
N/A

### How to test
This change is covered by existing unit tests; see the changed files for details. Execute the tests with `yarn test`, both in services/app-api and services/ui-src.

### Important updates
N/A

### Author checklist


oweingrod-coforma:
@benmartin-coforma there's some linter errors, but once those are fixed we can merge this without waiting for you to have full access, as unit tests run locally.

braxex:
>@benmartin-coforma there's some linter errors, but once those are fixed we can merge this without waiting for you to have full access, as unit tests run locally.

`pre-commit install` from readme should help you out with those linting errors

braxex:
@benmartin-coforma Where do we stand on moving forward on this one?

benmartin-coforma:
I think this PR is stuck, without a broader discussion of what we're trying to accomplish, and what risks we're OK with.

The biggest risk is that this (or any similar) change will invalidate a bunch of existing reports. I know we don't proactively re-validate these, but the risk is that a user who pulls up a report & tries to save it back down will be unable to do so, without editing some of the numeric fields.

I'm personally very risk-averse, so I want to avoid this, without at least knowing the size of the problem -- as in, how many fields on how many existing reports would fail the updated validation? I've written [a script that would perform that analysis](https://gist.github.com/benmartin-coforma/4edcec8a1580a2e8f460c0dbd9e11b0c) on all existing data, against a bunch of candidate regular expressions with varying strictness. But I don't know how exactly we would want to run that script against production data, and I don't know if the script actually works correctly.

And, given all the effort involved here, I don't know if it's worth our time to even run those checks. Perhaps I'm being too conservative, and we should skip the production data survey? But I can't make that call myself either.

If it's up to me to make these decisions, then I would sooner abandon this PR and delete the branch.

braxex:
@benmartin-coforma Great write-up. For right now let's close it. Write up a Jira ticket (bug) in the MCR tech debt epic explaining the issue and linking to the closed PR. Tag Davids and we can prioritize future work on it.

PR #11139, opened 2023-03-10
[MDCT-2215] Lock + Release MLR reports

oweingrod-coforma:
## Summary

This PR adds some functionality to release and lock MLR reports.

When reports are submitted they are locked, meaning they are **viewable** but not **editable**. When reports are "released", they are unlocked, and changes to the reports field data are saved in a new location in S3. (We also track these locations `reportMetadata.previousRevisions`)

This PR also cherry-picks the "submit" handler from the statusing feature. This changes the way in which we submit reports—instead of calling the update handler, we call the submit handler, which changes the requisite details in the report metadata.

### Description
Lastly, I've started adding more complete interfaces with simple validators for Dynamo data. Specifically, a new `isMLRReport` function which will cast any object to an `MLRReportMetadata` type if the correct data is in the report.

### Related ticket
[](https://qmacbis.atlassian.net/browse/MDCT-2215)

### How to test 
1. Create and submit an MLR report
2. Verify that report's metadata is set to "locked"
3. From the admin dashboard, select "release" for that report.
4. Verify that the report's metadata has a new field id, and the "previousRevisions" array has a value.
5. Enter the released report
6. Verify that option 1 has been selected for question G, and that question H is visible
7. Verify that all options for question H are selectable.

### Important updates
No changes to environment files, but this is not fully tested yet.

### Author checklist


oweingrod-coforma:
@karla-vm I removed some tests that are related to statusing (the submit handler will fail if the form is incomplete, but we don't have a concept of status yet). Just a heads up as to why those are removed.

karla-vm (approved):
Looks good! Just had a question on the `Archive` functionality in the deployed branch.

services/app-api/handlers/reports/release.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  ...mockProxyEvent,
+};
+
+describe("Test releaseReport method", () => {
+  beforeEach(() => {
+    // fail state and pass admin auth checks
+    mockAuthUtil.hasPermissions
+      .mockReturnValueOnce(true)
+      .mockReturnValueOnce(true)
+      .mockReturnValueOnce(true)
```
tblackwe, line 42: Can we find another way to do this? There is no context on when to return true vs false and is dependent on 5 tests running in the correct order.

services/app-api/handlers/reports/release.ts
```diff
@@ -0,0 +1,184 @@
+import handler from "../handler-lib";
+// utils
+import dynamoDb from "../../utils/dynamo/dynamodb-lib";
+import {
+  DynamoGet,
+  DynamoWrite,
+  isMLRReportMetadata,
```
tblackwe, line 7: will it only be MLR that can use this functionality? or will NAAAR reports get locked/unlocked as well?

EIther way we could use interface inheritance to standardize this method

services/ui-src/src/components/modals/AddEditReportModal.tsx
```diff
@@ -117,6 +120,7 @@ export const AddEditReportModal = ({
         fieldData: {
           ...dataToWrite.fieldData,
           stateName: States[activeState as keyof typeof States],
+          submissionCount: reportType === "MLR" ? 0 : undefined,
```
tblackwe, line 123: will NAAAR need submission counts? I would prefer the front end and back end both be program agnostic whenever possible, leveraging configuration flags rather than hardcoding report type toggles.

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
```diff
@@ -61,7 +61,7 @@ export const DashboardPage = ({ reportType }: Props) => {
   >(undefined);
   const [reportId, setReportId] = useState<string | undefined>(undefined);
   const [archiving, setArchiving] = useState<boolean>(false);
-  const [unlocking, setUnlocking] = useState<boolean>(false);
+  const [releasing, setReleaseing] = useState<boolean>(false);
```
tblackwe, line 64: nit: typo on setReleasing

tests/cypress/tests/e2e/mcpar/form.feature
```diff
@@ -3,7 +3,13 @@ Feature: MCPAR E2E Form Submission
     Scenario: A state user can fully create a form and submit it
         Given I am logged in as a state user
         When I submit a new MCPAR program
-        Then the program is submitted
+# Then the program is submitted
+
```
tblackwe, line 7: this belongs to another story.

tests/cypress/tests/e2e/mcpar/form.stepdef.js
```diff
@@ -129,6 +129,44 @@ Then("the program is submitted", () => {
   cy.contains("Successfully Submitted").should("be.visible");
 });
 
+When("I try to submit an incomplete MCPAR program", () => {
```
tblackwe, line 132: this belongs to another story.

tests/cypress/tests/e2e/mlr/form.stepdef.js
```diff
@@ -27,9 +27,7 @@ When("I submit a new MLR program", () => {
   cy.get('[data-testid="modal-submit-button"]').focus().click();
 });
 
-Then("the program is submitted", () => {
-  cy.contains("Successfully Submitted").should("be.visible");
-});
+Then("the program is submitted", () => {});
```
tblackwe, line 30: this belongs to another story.

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
oweingrod-coforma: oops! Thanks, find and replace

tests/cypress/tests/e2e/mcpar/form.feature
oweingrod-coforma: Meaning another stepdef file? I.e `submit.stepdef.js`

services/app-api/handlers/reports/release.test.ts
oweingrod-coforma: Agreed. I am new to jest mocking (previously would just do full e2e tests or write methods around stuff I couldn't run in test) so I was just following conventions here. Any thoughts on how this could be improved?

tests/cypress/tests/e2e/mcpar/form.feature
tblackwe: was submit part of this story? I saw the 'incomplete' program and (perhaps wrongly) assumed it was accidentally blended from statusing.

oweingrod-coforma: No, I needed the submit handler in this branch so I cherry-picked my commit from `statusing`

oweingrod-coforma: Long story short—MLR has unique behavior on submit. Rather than add that to an update flow which we were deprecating anyways, I moved the submit handler (and associated tests) over.

services/app-api/handlers/reports/release.test.ts
tblackwe: I'll take a look at get back to you. I don't know for certain either :D

tblackwe: the majority of the time we want permissions to return true, so setting that as the default response when we create mockAuthUtil

`const mockAuthUtil = require("../../utils/auth/authorization");
mockAuthUtil.hasPermissions.mockReturnValue(true);`

then later when we need it to return false we override it to be false
`test("Test release report without admin permissions throws 403", async () => {
    mockAuthUtil.hasPermissions.mockReturnValueOnce(false);
    mockDocumentClient.get.promise.mockReturnValueOnce({
      Item: mockDynamoDataMLRLocked,
    });
    const res = await releaseReport(releaseEvent, null);
    expect(res.statusCode).toBe(StatusCodes.UNAUTHORIZED);
    expect(res.body).toContain(error.UNAUTHORIZED);
  });`

This allows us to delete the `beforeEach` function call from the `describe` block

oweingrod-coforma: @tblackwe Makes sense! I'm going to write another PR towards `mlr` that has those changes, since this PR is closed already.

services/ui-src/src/components/modals/AddEditReportModal.tsx
oweingrod-coforma: Talked to DG about this—he said we're a long way off from knowing.

PR #11147, opened 2023-03-13
MLR Admin Actions Fix

karla-vm:
### Description

Admins should only be able to unlock a report that is not archived or that is locked.

### Related ticket

N/A

### How to test

- Create an MLR report
- Log in as admin
- Archive that report

`Unlock` should be disabled.

### Important updates

N/A

### Author checklist

### DIFF
--- a/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
@@ -147,7 +147,7 @@ const AdminUnlockButton = ({
     <Td>
       <Button
         variant="link"
-        disabled={report.locked === false}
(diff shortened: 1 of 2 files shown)


services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -147,7 +147,7 @@ const AdminUnlockButton = ({
     <Td>
       <Button
         variant="link"
-        disabled={report.locked === false}
```
tblackwe, line 150: you are inverting the locked value from previous. is that intentional?

karla-vm: oops, good catch!

PR #11160, opened 2023-03-16
Add `SectionHeader` form field type

oweingrod-coforma:
## Summary
JSON forms need to be able to add layout elements, like section headers. This PR adds that ability, alongside a `SectionHeader` component.

### Description
It’s a field type so it can be rendered as part of a form, but it’s actually not a field in the sense of needing to by hydrated or accepting input.

It wasn't specified, but I went ahead and gave it some customization. The component can take the prop `divider = "none" | "top" | "bottom"`, which changes the location of the `<hr />`.

Add it to your form with: 
```
  {
    "id": "sectionTest",
    "type": "sectionHeader",
    "validation": "text",
    "props": {
      "content": "1. Section Header Test",
      "divider": "top"
    }
  },
```

### Related ticket
[MDCT-2240](https://qmacbis.atlassian.net/browse/MDCT-2240)
### How to test
1. Add the above snippet to the field items of your MCPAR form
2. Verify the section header text appears
3. Verify that the divider options work correctly

### Author checklist

### DIFF
--- a/services/ui-src/src/components/forms/FormLayoutElements.test.tsx
+++ b/services/ui-src/src/components/forms/FormLayoutElements.test.tsx
@@ -1,0 +1,75 @@
+import { render, screen } from "@testing-library/react";
+import { axe } from "jest-axe";
+import { SectionHeader } from "./FormLayoutElements";
+
+const sectionHeaderComponentTopDivider = (
+  <SectionHeader
+    divider="top"
+    name="sectionHeader"
+    content="1. Section Header"
+    data-testid="test-section-header"
+  />
+);
+
+const sectionHeaderComponentBottomDivider = (
+  <SectionHeader
+    divider="bottom"
+    name="sectionHeader"
+    content="1. Section Header"
+    data-testid="test-section-header"
+  />
+);
+
+const sectionHeaderComponentNoDivider = (
+  <SectionHeader
+    divider="none"
+    name="sectionHeader"
+    content="1. Section Header"
+    data-testid="test-section-header"
+  />
+);
+
+describe("Test SectionHeader component", () => {
+  test("Top should make the section divider on the top.", async () => {
+    const { findByText, getByRole } = render(sectionHeaderComponentTopDivider);
+    const sectionHeader = screen.getByTestId("test-section-header");
+
+    expect(sectionHeader).toBeVisible();
+    expect(await findByText("1. Section Header")).toBeTruthy();
+    expect(sectionHeader.childNodes[0]).toMatchObject(getByRole("separator"));
+  });
+
+  test("Bottom should make the section divider on the bottom", async () => {
+    const { findByText, getByRole } = render(
+      sectionHeaderComponentBottomDivider
+    );
+    const sectionHeader = screen.getByTestId("test-section-header");
+
+    expect(sectionHeader).toBeVisible();
+    expect(await findByText("1. Section Header")).toBeTruthy();
+    expect(sectionHeader.childNodes[1]).toMatchObject(getByRole("separator"));
+  });
+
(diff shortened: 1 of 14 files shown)


tblackwe (commented):
top divider looks very close to the element above. 
image


bottom divider looks nicely spaced. 
image

braxex:
@oweingrod-coforma I think we'd probably want to change some types with this too. Like I don't think `sectionHeader` should have to have `validation` because that doesn't make much sense. It's not a form field that takes input. And that might actually cause some issues down the line with statusing. @tblackwe Thoughts?

What if we expanded a form's type to allow not just an array of form fields, but also some html elements as needed? Those could still be custom components like you have (though I guess they wouldn't need to be if we used the custom HTML ingestion we have built for other things), but they wouldn't have to follow the type restrictions that form fields have.

services/ui-src/src/components/forms/FormLayoutElements.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+    name="sectionHeader"
+    content="1. Section Header"
+    data-testid="test-section-header"
+  />
+);
+
+describe("Test SectionHeader component", () => {
+  test("Top should make the section divider on the top.", async () => {
+    const { findByText, getByRole } = render(sectionHeaderComponentTopDivider);
+    const sectionHeader = screen.getByTestId("test-section-header");
```
tblackwe, line 35: we are trying to only use getByTestId when absolutely necessary going forward. I have some suggestions on how to test this without test ids and child nodes

tblackwe: I have the code written locally but would like to review with you prior to committing.

tblackwe (changes requested):

services/ui-src/src/components/forms/FormLayoutElements.test.tsx
tblackwe: I am going to need to be done soon to make dinner, if you look at this again before tomorrow:

> describe("Test SectionHeader component", () => {
>   test("Top should make the section divider on the top.", async () => {
>     const { getByRole } = render(sectionHeaderComponentTopDivider);
> 
>     const heading = getByRole("heading", { name: "1. Section Header" });
>     const separator = getByRole("separator");
> 
>     expect(heading).toBeVisible();
>     expect(separator).toBeVisible();
>     expect(heading.compareDocumentPosition(separator)).toBe(
>       Node.DOCUMENT_POSITION_PRECEDING
>     );
>   });

oweingrod-coforma: This makes sense. Thanks for finding this for me!

oweingrod-coforma: Fixed!

BearHanded (dismissed):
My only comment is that chunks like this make me wonder if we will have future types that need to be included.
```
fields: Array<FormField | FormLayoutElement>,
```
If we start to accumulate more, we should have them extend a parent type. Not worth changing today.

oweingrod-coforma:
> My only comment is that chunks like this make me wonder if we will have future types that need to be included.
> 
> ```
> fields: Array<FormField | FormLayoutElement>,
> ```
> 
> If we start to accumulate more, we should have them extend a parent type. Not worth changing today.

I agree here. I was going to write it this way originally, but I decided against it because I knew it would change a ton of files.

oweingrod-coforma:
@tblackwe @braxex @BearHanded this is ready for re-review!

braxex (approved):
Haven't run it but it all looks reasonable to me. 👍

tblackwe (approved):

PR #11179, opened 2023-03-28
Make ReviewSubmit error status update more frequently when page hopping

ntsummers1:
### Description
The error message on the review submit page doesn't update if you race it by updating a field and then immediately navigating to the review and submit page. This makes it so the page always updates!

### How to test

Create a full filled out report
Go to Review Submit and see now error message at the top
Go to the first page on MCPAR
empty out a field, tab away, and immediately hit the review submit page in the side bar
See the error message pop in
Go back to the first page
Fill in the emptied field and then click the review submit page in the sidebar
Watch the error message disappear correctly!

### Author checklist


tblackwe (approved):
Validated fix on branch deployment

PR #11182, opened 2023-03-28
Add MLR export page

benmartin-coforma:
The meat of this commit is adding a bunch of ReportType-based branches to ExportedReportPage.tsx. This applies to verbiage (obviously the page title will be different), the presence of the combined data table (a MCPAR-specific feature), and which metadata we display.

Happily, the MLR data does render! Although there are future stories to tweak exactly how; some sections will differ further from the MCPAR.

Questions raised by this work:
 * Is the page title correct? I must check with Design before merging.
 * Is it appropriate to add `JSX` as a known global to ESLint? If not, should we exclude .TSX files from the no-undef rule? https://typescript-eslint.io/linting/troubleshooting/#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors
 * How should we handle verbiage for Export sub-components, which will not change between MCPAR and MLR (or, likely, NAAAR)? Currently, I am simply referring to the MCPAR verbiage. Instead we could recreate the exportVerbiageMap in each component. Or perhaps a broader, more involved solution, which may help simplify all components with ReportType-specific verbiage?
 * Am I going too far with the unit tests? We need to weigh thoroughness against execution time; a test that never fails is a waste of every dev's time forever.
 * Is `assertExhaustive` worthwhile? Defined in an appropriate place?

Related ticket: MDCT-2227

### Author checklist


services/ui-src/src/components/pages/Export/ExportedReportPage.tsx
```diff
@@ -93,6 +76,119 @@ export const ExportedReportPage = () => {
...
+                convertDateUtcToEt(report.lastAltered),
+                report.lastAlteredBy,
+                report.status,
+              ],
+            ],
+          }}
+        />
+      );
+    default:
+      assertExhaustive(reportType);
```
oweingrod-coforma, line 159: I like this alternative better than the assignment to `never`, thanks for adding!

oweingrod-coforma (approved):

oweingrod-coforma:
> Is the page title correct? I must check with Design before merging.

I think we can provide design with the entire PDF page rather than just parts, so I wouldn't have that block this merge.

> Is it appropriate to add JSX as a known global to ESLint?

I am in favor of this change, but @braxex and @BearHanded should weigh in. Does enabling it create lots of linter errors?

> How should we handle verbiage for Export sub-components

A lot of options here—easiest is to have "baseExportVerbiage" that contains all of the nonspecific data.

 In a perfect world, we could create class based verbiage generators that inherit from a base verbiage: 

```
Verbiage(reportType)
|____ExportVerbiage(reportType)
|__________MLRExportVerbiage(reportType)
|__________MCPARExportVerbiage(reportType)
```

then the verbiage map could become be a function: 
```
getExportVerbiage(reportType: ReportType): ExportVerbiage {
   switch(reportType) {
     case MCPAR: return MCPARExportVerbiage
     case MLR: return MLRExportVerbiage
     case NAAR: return ExportVerbiage
     default: handleErrorStuff
  }
}
```

These classes could implement all of the "render" functionality that we need them to (returning JSX with small copy and attribute access changes), and we could use the report context to select the correct verbiage.

If we have time for that, I think it'd be a spike ticket—I don't have enough recent React experience to know exactly how or where this should be implemented (thoughts @karla-vm @ntsummers1 @tblackwe @braxex @BearHanded?)

For right now, this is fine and IMO non-blocking, and we can finish the rest of the export page before worrying about this.

> Am I going too far with the unit tests?

Never too far!!!!

> Is assertExhaustive worthwhile? Defined in an appropriate place?

Yes and yes. Thank you for adding!

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
```diff
@@ -6,7 +6,7 @@ import { EntityCard, ReportContext } from "components";
 import { getFormattedEntityData } from "utils";
 import { EntityShape, ModalDrawerReportPageShape } from "types";
 // verbiage
-import exportVerbiage from "verbiage/pages/export";
+import exportVerbiage from "verbiage/pages/mcpar/mcpar-export";
```
tblackwe, line 9: does MLR not have export verbiage here?

services/ui-src/src/types/reports.ts
```diff
@@ -1,5 +1,12 @@
 import { AnyObject, CustomHtmlElement, FormJson } from "types";
 
+// REPORT TYPES
+export enum ReportType {
+  MCPAR = "MCPAR",
+  MLR = "MLR",
+  NAAR = "NAAR",
```
tblackwe, line 7: 2 `A` not enough. NEEDS MORE `A`! 

`NAAAR`

tblackwe (changes requested):
typo in `NAAAR`

Can we avoid assuming MCPAR report type in verbiage?

services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
karla-vm: @tblackwe we do not have a Modal Drawer for MLR currently.

services/ui-src/src/components/pages/Export/ExportedReportPage.tsx
```diff
@@ -12,19 +12,38 @@ import { Spinner } from "@cmsgov/design-system";
...
     (route: ReportRoute) => route
   );
 
+  const reportType = (report?.reportType ||
+    localStorage.getItem("selectedReportType")) as ReportType;
+
+  const exportVerbiageMap: { [key in ReportType]: any } = {
+    MCPAR: mcparVerbiage,
+    MLR: mlrVerbiage,
+    NAAR: undefined,
```
karla-vm, line 41:
```suggestion
    NAAAR: undefined,
```

services/ui-src/src/components/export/ExportedReportBanner.tsx
```diff
@@ -6,9 +6,24 @@ import { Box, Button, Image, Text } from "@chakra-ui/react";
...
 
 export const ExportedReportBanner = () => {
+  const { report } = useContext(ReportContext);
+  const reportType = (report?.reportType ||
+    localStorage.getItem("selectedReportType")) as ReportType;
+
+  const verbiageMap: { [key in ReportType]: any } = {
+    MCPAR: mcparVerbiage,
+    MLR: mlrVerbiage,
+    NAAR: undefined,
```
karla-vm, line 23:
```suggestion
    NAAAR: undefined,
```

karla-vm (dismissed):
Neat stuff, Ben! Other than Thomas's comment about the missing `A` in `NAAAR` you're good to go 💯

tblackwe (approved):

oweingrod-coforma (approved):

PR #11189, opened 2023-03-30
Add more custom elements and re-write JSON to use them

oweingrod-coforma:
## Summary

Add new custom elements, `ol`, `ul`, and `li`, which map to Chakra components. Re-write existing forms to use these custom elements.

Also, add code to allow the custom element parsing to handle nested children correctly.

### Author checklist

### DIFF
--- a/services/ui-src/src/utils/other/parsing.ts
+++ b/services/ui-src/src/utils/other/parsing.ts
@@ -1,21 +1,33 @@
-import React from "react";
+import React, { ReactElement } from "react";
 import { sanitize } from "dompurify";
 import parse from "html-react-parser";
 // components
 import { Link as RouterLink } from "react-router-dom";
-import { Heading, Link, Text } from "@chakra-ui/react";
+import {
+  Heading,
+  Link,
+  ListItem,
+  OrderedList,
+  Text,
+  UnorderedList,
+} from "@chakra-ui/react";
 // types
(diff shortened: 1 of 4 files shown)


karla-vm (approved):

services/ui-src/src/utils/other/parsing.ts
```diff
@@ -38,12 +49,34 @@ export const parseCustomHtml = (element: CustomHtmlElement[] | string) => {
...
   }
 };
 
+export function createElementWithChildren(
+  element: CustomHtmlElement
+): ReactElement {
+  const { type, content, as, props } = element;
+  const elementType: string = customElementMap[type] || type;
+  const elementProps = {
+    key: type + `${Math.floor(Math.random() * 100)}`,
```
tblackwe, line 65: Assume we are doing this because we need a unique key? What's the effect of we have a duplicate?

oweingrod-coforma: > Assume we are doing this because we need a unique key? What's the effect of we have a duplicate?

Yeah, previously this was just the index of the field in the list of all fields. Duplicate would cause an error at least in the console, not sure if a breaking one.

oweingrod-coforma: @tblackwe I changed this so we now generate a UUID for the repeated elements

PR #11203, opened 2023-04-05
MDCT-2313: MLR Statusing

karla-vm:
### Description

Adding the new MLR page types to the status completion calculator.

### Related ticket(s)

MDCT-2313

### How to test

WIP

### Important updates

N/A

---
### Author checklist

---


oweingrod-coforma:
@karla-vm We should try to create some fixtures like Thomas did, might help us figure out why this doesn't work

services/ui-src/src/utils/reports/reports.ts
```diff
@@ -115,6 +116,12 @@ export const compileValidationJsonFromRoutes = (
     // if drawer form present, add validation to schema
     const drawerFormFields = route.drawerForm?.fields.filter(isFieldElement);
     if (drawerFormFields) addValidationToAccumulator(drawerFormFields);
+    if (route.pageType === "modalOverlay") {
+      const overlayFormFields = (
+        route as ModalOverlayReportPageShape
+      ).overlayForm?.fields.filter(isFieldElement);
+      if (overlayFormFields) addValidationToAccumulator(overlayFormFields);
+    }
```
benmartin-coforma, line 124: It's a little unfortunate that this can't follow the same pattern as `standardFormFields`, `modalFormFields`, and `drawerFormFields`. I'm guessing that the string check on `pageType` here is saving us from a higher-impact change to the `ReportRoute` type?

oweingrod-coforma: Basically, yes. Because of the way the types are created, we need to cast `route` to the Modal Overlay page shape. We only want to do this if we know it's a modal overlay page, hence the string check.

services/ui-src/src/components/tables/EntityStatusIcon.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const reportFormValidation = Object.fromEntries(
+      Object.entries(report?.formTemplate.validationJson ?? {}).filter(
+        ([key]) => key.includes("report_") || key.includes("state_")
+      )
+    );
+
+    const formValidationSchema =
+      mapValidationTypesToSchema(reportFormValidation);
+    const formResolverSchema = object(formValidationSchema || {});
+    mapValidationTypesToSchema;
```
benmartin-coforma, line 26: I don't think this line is doing anything?

services/app-api/utils/validation/completionSchemas.ts
benmartin-coforma, line 1: Nice work in here!

benmartin-coforma (approved):
I haven't pulled this down & tested, but all of the code looks clean!

services/ui-src/src/components/tables/EntityStatusIcon.tsx
braxex: Yeah line 26 looks like it can be deleted.

ntsummers1 (changes requested):
Really nice work! Especially with the fixtures and updating the schemas and all that.

Gave it a run and was able to find this:

https://user-images.githubusercontent.com/19439679/233645560-68ef9662-94b2-4b8d-8d3d-d3c87b74abf1.mov

Steps to Reproduce:

1. Create a submission
2. Complete it and submit it
3. Log in as an admin and unlock it
4. Log in as a state user and open the submission back up

Is this a bug? 80% of me says it is but I could also juuust see it not being one. What do you think?

oweingrod-coforma:
> Really nice work! Especially with the fixtures and updating the schemas and all that.
> 
> Gave it a run and was able to find this:
> 
>  Screen.Recording.2023-04-21.at.9.09.42.AM.mov 
> Steps to Reproduce:
> 
> 1. Create a submission
> 2. Complete it and submit it
> 3. Log in as an admin and unlock it
> 4. Log in as a state user and open the submission back up
> 
> Is this a bug? 80% of me says it is but I could also juuust see it not being one. What do you think?

That is a bug! I think I forgot to some fields. Thank you!

services/ui-src/src/components/tables/EntityStatusIcon.tsx
oweingrod-coforma: Indeed it does not!

ntsummers1 (approved):

ntsummers1:
Great work!

ntsummers1 (approved):

braxex (approved):

PR #11204, opened 2023-04-06
MLR export page 1

benmartin-coforma:
### Description
Complete page 1 of the MLR PDF export. It was already in place and rendering almost how we wanted, thanks to prior work.

The main change here is to hide the question hint text; currently it is shown for all tables in all reports. Now, it is hidden (only) for the Primary Contact Information table of the MLR report.

There is also a small verbiage change.

### Related ticket(s)
MDCT-2309

---
### How to test
Create and fill out an MLR report, then view the PDF preview. You will see no gray explanatory text in the table on the first page. Other tables in the report will still have hint text.

Optionally, view the PDF export for a MCPAR report. It will still have hint text everywhere.

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/ui-src/src/components/export/ExportedReportFieldTable.test.tsx
+++ b/services/ui-src/src/components/export/ExportedReportFieldTable.test.tsx
@@ -9,7 +9,11 @@ import {
 } from "utils/testing/setupJest";
 import { ReportContext } from "components";
 import { ExportedReportFieldTable } from "./ExportedReportFieldTable";
-import { DrawerReportPageShape } from "types";
+import {
+  DrawerReportPageShape,
+  ReportContextShape,
+  StandardReportPageShape,
+} from "types";
 
 // Contexts
 const reportJsonFields = [{ ...mockNestedFormField, id: "parent" }];
@@ -22,12 +26,14 @@ const nestedParent = mockNestedFormField;
 nestedParent.props.choices[2].children = [{ ...mockFormField, id: "child" }];
 
 const mockStandardContext = { ...mockMcparReportContext };
+mockStandardContext.report = { ...mockStandardContext.report };
 mockStandardContext.report.fieldData = {
   ...mockStandardContext.report.fieldData,
   ...fieldData,
 };
 
 const mockDrawerContext = { ...mockMcparReportContext };
+mockMcparReportContext.report = { ...mockMcparReportContext.report };
 mockDrawerContext.report.fieldData = {
   ...mockDrawerContext.report.fieldData,
   plans: [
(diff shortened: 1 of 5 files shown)


services/ui-src/src/components/export/ExportedReportFieldTable.tsx
```diff
@@ -38,6 +39,11 @@ export const ExportedReportFieldTable = ({ section }: Props) => {
     ? twoColumnHeaderItems
     : threeColumnHeaderItems;
 
+  // The hint text is hidden for MLR page 1 (Point of Contact)
+  const reportType = report?.reportType as ReportType;
+  const hideHintText =
+    reportType === ReportType.MLR && section.form?.id === "apoc";
```
ntsummers1, line 45: Hey @oweingrod-coforma, I'm going to approve this for now since you're looking at adding another section to hide the hint text. I suggest though (And maybe @braxex would also have an opinion here) we move this out into the MLR.json file. For example, we could use a prop for this.

```
        "fields": [
          {
            "id": "contactName",
            "type": "text",
            "validation": "text",
            "props": {
              "label": "A. Contact name",
              "hint": "Enter the first and last name of the primary contact for this report.",
              "showHintInExport": false
            }
          },
```

That `showHintInExport` could be a passed prop so that we don't need to call out specific section, we can just reference the prop.

braxex: That is something we could do, but it would have to be on the `form` instead of on the individual `field`. Not a bad idea though. Maybe something for @benmartin-coforma @ntsummers1 @oweingrod-coforma to collaborate on.

services/ui-src/src/components/export/ExportedReportFieldTable.test.tsx
```diff
@@ -77,6 +83,61 @@ const emptyTableComponent = (
...
+      {
+        id: "mockFieldId",
+        props: {
+          label: "X. Mock Field label",
+          hint: "Mock Hint Text",
+        },
+      },
+    ],
+  },
+} as unknown as StandardReportPageShape;
```
ntsummers1, line 106: Swap this to using the mocks

ntsummers1 (approved):

services/ui-src/src/components/export/ExportedReportFieldTable.tsx
ntsummers1: I think this needs to be on a field by field basis unfortunately. There are 2 fields in the MLR figma that still have the hint text shown despite everything else not having hint text. Both of those fields are disabled, however, other disabled fields don't show their respective hint text. 

Example:
This field _doesn't_ show its hint text
```
{
  "id": "stateName",
  "type": "text",
  "validation": "text",
  "props": {
    "label": "E. State",
    "hint": "Auto-populates from your account profile.",
    "disabled": true
  }
},
```

This field _does_ show its hint text
```
{
  "id": "reportingPeriodStartDate",
  "type": "date",
  "validation": {
    "type": "date",
    "nested": true,
    "parentFieldName": "contractIncludesRemittance"
  },
  "props": {
    "label": "N. MLR reporting period start date",
    "hint": "Auto-populates from program reporting information.",
    "disabled": true
  }
},
```

And thats based off the figmas here: https://www.figma.com/file/GBktI8ZIJM9YYcdNva0ltg/MDCT---MCR?node-id=7591-162697&t=ZzRxhOyMtP2eQRp0-0

braxex: 🤦 Let's talk to design about this. Maybe we have to build it in at the form and field level so either can choose to hide it. Because show is the default, the property needs to be `hide....`. That said, I really don't like changing things like this solely for the purpose of a display-only non-interactive export view, so I'd like to push back on them and say you either get hint text or don't get hint text at the form level, not at the field level. The reporting period start date and end date being part of the entire report should be pretty obvious IMO. Start a convo with me and design please?

oweingrod-coforma: @braxex We could have the property default to `showHintTextOnExport=true` and then set it to false when we don't want it to appear.

oweingrod-coforma (approved):

services/ui-src/src/components/export/ExportedReportFieldTable.tsx
ntsummers1: Convo started for you!

benmartin-coforma: If we move this logic to the JSON file, we will do so under a different PR.

ntsummers1 (approved):

oweingrod-coforma (approved):

PR #11223, opened 2023-04-14
[MDCT-2399] Skip Hosted UI - use the existing cf var, and export it

BearHanded:
### Description

Thread the existing cf var back through for the Auth library.

**2 related things in this PR**
1. The env_example name for the existing variable was renamed to be used for the new one. This would have been fine, except the script that generates the env for vars from serverless.yaml checks that entries exist in the example before it saves them. Encourages good documentation I guess, but I wish there was a warning.
2. Removes the new variable that was just importing the same value from SSM that is generated from the cloudformation output. Less moving parts.

### Related ticket(s)

MDCT-2399

---
### How to test

Merge it to dev and see if the IDM flow works there now.

### Important updates

---
### Author checklist

---


ntsummers1 (approved):
Reviewed with Brendan, looks good. Fingers crossed this fixes everything!

PR #11225, opened 2023-04-17
Update isReportFormPage functions to more accurately assesses if in a form

ntsummers1:
### Description

Screenshot 2023-04-14 at 4 11 40 PM

This rouge report status header was showing up when going to the MLR dashboard. Lets get that fixed up!

---
### How to test

./dev local
Open up the MLR section
See theres no header!
Create a submission
See there is a header and it updates when you make changes to the submission!

---
### Author checklist

### DIFF
--- a/services/ui-src/src/forms/mcpar/index.ts
+++ b/services/ui-src/src/forms/mcpar/index.ts
@@ -18,12 +18,5 @@ export const mcparReportJson = {
   validationJson: compileValidationJsonFromRoutes(mcparReportRoutesFlat),
 };
 
-export const isMcparReportFormPage = (pathname: string): boolean => {
-  let i = 0;
-  for (i; i < mcparReportRoutesFlat.length; i++) {
-    if (mcparReportRoutesFlat[i].path === pathname) {
-      return true;
-    }
-  }
-  return false;
-};
+export const isMcparReportFormPage = (pathname: string): boolean =>
+  mcparReportRoutesFlat.some((route) => route.path == pathname);
--- a/services/ui-src/src/forms/mcpar/index.ts
+++ b/services/ui-src/src/forms/mcpar/index.ts
@@ -18,6 +18,12 @@ export const mcparReportJson = {
   validationJson: compileValidationJsonFromRoutes(mcparReportRoutesFlat),
 };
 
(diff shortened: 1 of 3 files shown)


services/ui-src/src/forms/mcpar/index.ts
```diff
@@ -18,6 +18,12 @@ export const mcparReportJson = {
...
-  pathname.includes("/mcpar/") && !nonFormPages.includes(pathname);
+export const isMcparReportFormPage = (pathname: string): boolean => {
+  let i = 0;
+  for (i; i < mcparReportRoutesFlat.length; i++) {
+    if (mcparReportRoutesFlat[i].path === pathname) {
+      return true;
+    }
+  }
+  return false;
+};
```
benmartin-coforma, line 29:
```suggestion
export const isMcparReportFormPage = (pathname: string): boolean => {
  return mcparReportRoutesFlat.map((route) => route.path).includes(pathname);
};
```

benmartin-coforma (changes requested):
Tested locally, works great.

services/ui-src/src/forms/mcpar/index.ts
benmartin-coforma: Alternatively, for short-circuiting behavior:
```ts
  return mcparReportRoutesFlat.some((route) => route.path == pathname);
```

ntsummers1: For sure! Since this code is pretty self explanatory the benefit or .map/.some/.forEach's readability against the speed was something I felt speed should win out on. That being said I'm happy to swap it out if you feel that it would be better to use this method.

ntsummers1: For comparison, heres a quick performance test I wrote:

Screenshot 2023-04-17 at 11 46 46 AM

benmartin-coforma: If I'm reading this right, the `array.some` approach would be able to execute 1900 times per second, even if we had 100,000 routes? That's absolutely fast enough that I would rather take the performance penalty than have an index-iterating for loop - especially given that our route count is more likely to be in the 10s than the 100,000s.

benmartin-coforma (approved):

PR #11231, opened 2023-04-26
Bugfix dynamodb local

tblackwe:
### Description
Updates dynamodb local to use direct override while we wait for serverless-dynamodb-local to be updated.

### Related ticket(s)

Bugfix
---
### How to test
Perform a fresh checkout
Follow the local build/deploy instructions
Validate that your dynamodb instance spun up without errors.

### Important updates

---
### Author checklist

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,23 +74,18 @@ In the terminal, any changes made to a program will show up as S3 updates with a
 
 #### DynamoDB Jar Not Found Workaround
 
-Currently (March 31, 2023) there is a bug which prevents local usage of DynamoDB.
+Currently (April 19th, 2023) there is a bug which prevents local usage of DynamoDB.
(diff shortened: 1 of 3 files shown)


oweingrod-coforma (approved):

oweingrod-coforma:
> I'm not going to block this, but I'm not a fan of it either.
> 
> On the one hand, it will be nice not to have to tell onboarding developers about the weird hack they have to do in their node_modules folder.
> 
> On the other hand, overriding NPM dependencies by pointing to specific github commits also feels like hack to me. When those dependencies cut new versions, will we notice? Will we remember to remove both overrides from our package.json?
> 
> With the manual node_module change, we know that our hacks & workarounds will be wiped out with the next update of the NPM package. That feels reassuring to me.

I'm going to disagree here—only because this is a local tool. If this were a real dependency, I definitely agree, but since this is local, we can probably stick with this pinned version for as long as possible—there aren't really security concerns with this library, since it's just local form info

PR #11235, opened 2023-04-27
 MDCT-2420: Disable All Fields + Buttons if MLR Report is Locked

karla-vm:
### Description

If an MLR report is in a `Locked` state, all of its action buttons and fields should be disabled. 

### Related ticket(s)

MDCT-2420

---
### How to test

- Log in as a state user
- Create a new MLR report
- Submit the report

Verify that all fields and buttons are disabled ✨ 

- Log in as an admin user
- Navigate to the submitted report

Ditto, everything is now disabled.

### Important updates

N/A

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/tables/EntityRow.tsx
+++ b/services/ui-src/src/components/tables/EntityRow.tsx
@@ -10,6 +10,7 @@ import unfinishedIcon from "assets/icons/icon_error_circle_bright.png";
 export const EntityRow = ({
   entity,
   verbiage,
+  locked,
   openAddEditEntityModal,
   openDeleteEntityModal,
   openEntityDetailsOverlay,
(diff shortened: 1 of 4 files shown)


services/ui-src/src/components/tables/EntityRow.tsx
```diff
@@ -141,7 +149,7 @@ const sx = {
   deleteButton: {
     padding: "0",
     background: "white",
-    "&:hover": {
+    "&:hover:disabled": {
       background: "white",
     },
```
braxex, line 154: Question: Do we want this behavior _only_ when hovering _and_ disabled, or both when hovering _and_ when hovering + disabled?

braxex: _applies to other locations in PR_

karla-vm: Both!

keeysnc (approved):
Looks good to me! The only thing I noticed was code climate flagged a lower percentage on diff coverage. It seems like it might be an out of scope issue though?

karla-vm:
> Looks good to me! The only thing I noticed was code climate flagged a lower percentage on diff coverage. It seems like it might be an out of scope issue though?

Yep! I think there's an issue with CodeClimate for this particular branch, so we're going to have a separate effort to look into it :)

PR #11256, opened 2023-05-11
add internal user

braxex:
### Description

The app now needs to support an 'Internal User' role, which has the exact same in-app permissions as the 'Help Desk' role. The difference between the two is that when provisioned access via IDM role request, Help Desk users gain IDM help desk permissions, and internal users do not. Having an Internal User role helps better guard user privacy.

### Related ticket(s)

MDCT-2456

---
### How to test

1. All tests are passing
2. You can log in as `internaluser@test.com` and have the same in-app permissions as a help desk user (read only access to all reports, but no write access of any kind).

### Important updates

n/a

---
### Author checklist

---


karla-vm (approved):
💯

ntsummers1 (approved):

braxex:
Deployed Cloudfront link for testing: https://dp3dftrk53f4b.cloudfront.net/

PR #11271, opened 2023-05-17
MDCT-2457, 2458, 2459: Restrict Report Access for State Users w/o Permissions

karla-vm:
### Description

Limiting report access and actions for state users that may not have permission for certain report types (MCPAR, MLR, NAAAR)

### Related ticket(s)

MDCT-2457
MDCT-2458
MDCT-2459

---
### How to test

Everything should work as expected for now

### Important updates

N/A

---
### Author checklist

---


services/ui-src/src/components/app/AppRoutes.tsx
```diff
@@ -21,8 +21,11 @@ import { ScrollToTopComponent, useUser } from "utils";
 import { Fragment } from "react";
 
 export const AppRoutes = () => {
-  const { userIsAdmin } = useUser().user ?? {};
+  const { userIsAdmin, userReports } = useUser().user ?? {};
   const mlrReport = useFlags()?.mlrReport;
+  // determine if the user has access to specific reports
+  const hasMcpar = userReports?.includes("MCPAR") || userIsAdmin;
+  const hasMlr = userReports?.includes("MLR") || userIsAdmin;
```
braxex, line 28: Nit: Would maybe name these `hasMcparAccess` or similar. Alternatively could make a lookup object that has something like: 
```
userReportAccess = {
  MCPAR: userReports?.includes("MCPAR") || userIsAdmin,
  MLR: userReports?.includes("MLR") || userIsAdmin,
}
```
then access like:
```
userReportAccess["MCPAR"] ? whatever : whatever
```

services/ui-src/src/components/cards/TemplateCard.tsx
```diff
@@ -67,7 +67,7 @@ export const TemplateCard = ({
               leftIcon={
                 <Image src={downloadIcon} alt="Download Icon" height="1.5rem" />
               }
-              isDisabled={isDisabled}
+              isDisabled={false}
```
braxex, line 70: Question: Why is this one always false?

services/ui-src/src/components/forms/AdminDashSelector.tsx
```diff
@@ -51,7 +53,7 @@ export const AdminDashSelector = ({ verbiage }: Props) => {
     selectedReport = selectedReport.replace("report-", "").toLowerCase();
     localStorage.setItem("selectedReportType", selectedReport);
 
-    if (userIsAdmin || userIsReadOnly) {
+    if (userCanReadOnly) {
```
braxex, line 56: Comment: I'm on the fence with ones like this -- it might be more clear to keep it as the two explicitly noted groups.

braxex: ^^ This applies to all similar spots where we collapse the user groups `userIsAdmin` and `userIsReadOnly` (which for the uninitiated are composed of [super admins (bo reps) and approvers] and [help desk and internal users] respectively) down to `userCanReadOnly`.

services/ui-src/src/utils/testing/mockUsers.tsx
```diff
@@ -20,6 +20,25 @@ export const mockStateUser: UserContextShape = {
...
+export const mockStateUserNoReports: UserContextShape = {
+  user: {
+    userRole: UserRoles.STATE_USER,
+    email: "stateuser@test.com",
+    given_name: "Thelonious",
+    family_name: "States",
+    full_name: "Thelonious States",
+    state: "MN",
+    userIsEndUser: true,
+    userReports: undefined,
```
braxex, line 41: Question: Do we need to also test for if a user has selected no reports when they provision their role? Idk if state users will have undefined (all other user types definitely will though), but i think they might have the IDM reports attribute equal to an empty string `""` if they didn't select any reports. We should check on this/test for it I think.

braxex (commented):
Left some comments and questions re: potential edge cases but didn't have time to explicitly test or go through all the tests and logic.

services/app-api/utils/auth/authorization.ts
```diff
@@ -92,3 +93,21 @@ export const hasPermissions = (
 
   return isAllowed;
 };
+
+export const hasAccess = (event: APIGatewayProxyEvent, reportType: string) => {
```
BearHanded, line 97: The naming of this method is pretty interchangeable with hasPermissions() at a glance. Maybe hasReportAccess() like the var name you used, or updating this one to hasReport, and the other hasRole?

BearHanded (changes requested):

keeysnc (commented):
Hey @karla-vm Im getting an error message when trying to entering mcpar online.
Here's the error:
Screenshot 2023-05-18 at 11 49 47 AM

braxex:
> @karla-vm Did some digging and I think this error is on my end. I can see the 403 permission error showing which seems like its intended for the `staterep` user.

@karla-vm @keeysnc Looks like a bug to me. State reps should be able to fetch, read, write, update, etc. reports.

services/ui-src/src/components/cards/TemplateCard.tsx
karla-vm: Originally, the top-level / incoming `isDisabled` prop was set to `false` and only used for this particular download button, so I assumed users need to be able to download the template regardless of their access.

services/ui-src/src/utils/testing/mockUsers.tsx
karla-vm: good call, I changed this unit test to have an array of an empty string, just for consistency across state user roles.

BearHanded (approved):

keeysnc (approved):
Nice! Works for me now.

braxex (approved):

BearHanded (approved):

PR #11290, opened 2023-05-26
MDCT-2580: Clear data user no longer needs from Drawers/Modals

ntsummers1:
### Description

Fix Drawers so that if a user clicks off the parent choice then that child is cleared and its cleared state is sent to the db

Double checked that Modals are also having their child questions be properly cleared if a parent choice is unselected

Removed the band-aid fix for Sanctions in EntityCardBottomSection

### Related ticket(s)

MDCT-2580

---
### How to test

- Create a program
- Add a plan
- Go to Sanctions
- Create a sanction and add other text
- Check the network tab and see other text data was sent
- Edit the sanction modal and choose an option without other text
- See the network tab has sent an empty string for the other text option
- Repeat with the Sanction Details drawer and the remediation choicelist (Yes, In Progress, No) question
- Create a BSS Entity
- Go to a Drawer page (Section D)
- Find a choicelist or anything with a child choice
- Repeat adding othertext or entering the child, saving, coming back and choosing a different parent, and saving to see the past child data has been removed.
- 

---
### Author checklist

### DIFF
--- a/services/ui-src/src/components/cards/EntityCard/EntityCardBottomSection.tsx
+++ b/services/ui-src/src/components/cards/EntityCard/EntityCardBottomSection.tsx
@@ -83,7 +83,7 @@ export const EntityCardBottomSection = ({
               !formattedEntityData?.noncomplianceInstances ||
               !formattedEntityData?.dollarAmount ||
               !formattedEntityData?.assessmentDate ||
-              !formattedEntityData?.remediationDate ||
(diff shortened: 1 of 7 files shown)


karla-vm (approved):

britt-mo:
Tested and confirmed correct change

britt-mo (approved):

PR #11294, opened 2023-05-31
[MDCT-2582] - Handle cases where there is no formfield props during export render

ntsummers1:
### Description

Old reports (Reports created longer than 4 months ago) have a question that has no label. Since then we've updated the way that the exported report is generated and when it came across something that had no label it broke. This makes sure the any cases where theres an undefined label/hint text that we account for it

### Related ticket(s)

MDCT-2582

---
### How to test

- Create a new program and make sure that the exported fields come back correctly.
- Remove Message Nick for how to reproduce the error that Georgia is seeing OR For in the mcpar.json file replace lines 596-601, 

```
    "id": "program_contractDate",
    "type": "date",
    "validation": "date"
```

with
``` 
    "id": "program_contractDate",
    "type": "date",
    "validation": "date",
    "props": {
        "hint": "Enter the date of the contract between the state and plans participating in the managed care program."
    }
```
Create a new program, and the checkout the export. Look for C1.I.1 question

---
### Author checklist

### DIFF
--- a/services/ui-src/src/components/export/ExportedReportFieldRow.tsx
+++ b/services/ui-src/src/components/export/ExportedReportFieldRow.tsx
@@ -17,7 +17,7 @@ export const ExportedReportFieldRow = ({
   const { report } = useContext(ReportContext);
   const reportData = report?.fieldData;
   const isDynamicField = formField.type === "dynamic";
-  const formFieldInfo = parseFormFieldInfo(formField?.props!);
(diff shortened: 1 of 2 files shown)


karla-vm (approved):
LGTM!

britt-mo (approved):

britt-mo:
Tested and confirmed correct change

PR #11314, opened 2023-06-16
[MDCT-2628] Populate form templates into DynamoDB

oweingrod-coforma:
### Description

Lambda function, which when run, populates the database with existing form templates based on form templates in the bucket.

### Related ticket(s)

MDCT-2628

---
### How to test

1. `./dev local`
1. (in another terminal) `cd services/app-api`
1. `serverless invoke local --function formTemplateTablePopulation`

### Important updates

Updates the `.env.example` file, please update your environment.

---
### Author checklist

---


oweingrod-coforma:
Updated this to include some utilities that allow you to query secondary indexes `LastAlteredIndex` and `IdIndex` to get the most recent template version or get the template version by id.

oweingrod-coforma:
Need to write tests for this, but it's ready for review

services/app-api/handlers/templates/populateTemplatesTable.ts
```diff
@@ -0,0 +1,153 @@
...
+export async function processReport(reportType: typeof REPORT_TYPES[number]) {
+  const reportBucket = reportBuckets[reportType as keyof typeof reportBuckets];
+  const formTemplates = await s3Lib.list({ Bucket: reportBucket });
+
+  logger.info(`Found ${formTemplates.length} possible form template versions.`);
+
+  const sortedTemplates = formTemplates
+    .sort(
+      (a, b) =>
+        (b.LastModified?.getDate() ?? 0) - (a.LastModified?.getDate() ?? 0)
```
benmartin-coforma, line 77: Typo: `getDate()` returns the day of the month; we probably want `valueOf()`.

Thought: Subtracting the dates directly would also work (that is, `b.LastModified - a.LastModified`), if not for the special casing we would need to handle `undefined`. _Do_ we have any undefined LastModified dates? I can't imagine why we would... but it's probably not worth the effort to find out either.

oweingrod-coforma: Oh thank you, good catch!

Our version of the AWS SDK types all of the object attributes (`LastModified`, `Key`, etc) as optional. Not sure why. Rather than mangling types (as I do later in the file), I just revert to 0 on the off chance the attribute is actually undefined.

oweingrod-coforma:
@benmartin-coforma @BearHanded @karla-vm This should get another review. Code coverage is mostly being brought down by functions that just wrap AWS api calls. Happy to add mocks for these, but I have tested this locally (and we can continue to test it before actually running on real data).

BearHanded (dismissed):
LGTM and works locally, worth noting it will generate versions for all of your past local forms if they're still in your s3 bucket folder.

BearHanded (approved):

BearHanded (dismissed):

benmartin-coforma (commented):
I have 3 main questions:

1. Is the sort order correct? Unless I am wrong, this script will give higher version numbers to older templates.
2. Is it wise to ignore templates that failed to process? It seems we should use `Promise.all` instead of `Promise.allSettled` followed by `isFulfilled`.
3. Are the type acrobatics worth it? It seems we could replace `isDefined`, `SomeRequired`, and `S3ObjectRequired` with a few judicious exclamation marks. Something like this:

```ts
  const sortedTemplateKeys = formTemplates
    .filter(t => t.LastModified !== undefined && t.Key !== undefined)
    .sort((a, b) => b.LastModified!.getTime() - a.LastModified!.getTime())
    .map((t) => t.Key!)
    .filter((key) => key.endsWith(".json") && key.split("/")[1].length === 2);
```

I also have a few nits:

1. This might be a good time to introduce the `ReportType` enum to app-api, copying it from ui-src. Instead of looping over `REPORT_TYPES`, we would loop over `Object.values(ReportType)`. Instead of typing parameters as `typeof REPORT_TYPE[number]`, we would type them as `ReportType`.
2. The `FormTemplate` type should probably include `md5Hash: string`, right?
3. The utility function `getTemplateVersionByHash` should never be used outside of this `populateTemplatesTables`, right? Would it be worth moving into the script, so that it is harder for future developers to reference, and easier for them to delete?
4. We may want to create a folder app-api/handlers/formTemplates, and move this script there. Hopefully this will reduce confusion between JSON form templates, and XLSX report templates.

oweingrod-coforma:
> I have 3 main questions:
> 
> 1. Is the sort order correct? Unless I am wrong, this script will give higher version numbers to older templates.
> 2. Is it wise to ignore templates that failed to process? It seems we should use `Promise.all` instead of `Promise.allSettled` followed by `isFulfilled`.
> 3. Are the type acrobatics worth it? It seems we could replace `isDefined`, `SomeRequired`, and `S3ObjectRequired` with a few judicious exclamation marks. Something like this:
> 
> ```ts
>   const sortedTemplateKeys = formTemplates
>     .filter(t => t.LastModified !== undefined && t.Key !== undefined)
>     .sort((a, b) => b.LastModified!.getTime() - a.LastModified!.getTime())
>     .map((t) => t.Key!)
>     .filter((key) => key.endsWith(".json") && key.split("/")[1].length === 2);
> ```
> 
> I also have a few nits:
> 
> 1. This might be a good time to introduce the `ReportType` enum to app-api, copying it from ui-src. Instead of looping over `REPORT_TYPES`, we would loop over `Object.values(ReportType)`. Instead of typing parameters as `typeof REPORT_TYPE[number]`, we would type them as `ReportType`.
> 2. The `FormTemplate` type should probably include `md5Hash: string`, right?
> 3. The utility function `getTemplateVersionByHash` should never be used outside of this `populateTemplatesTables`, right? Would it be worth moving into the script, so that it is harder for future developers to reference, and easier for them to delete?
> 4. We may want to create a folder app-api/handlers/formTemplates, and move this script there. Hopefully this will reduce confusion between JSON form templates, and XLSX report templates.

1. This was driving me insane. I thought the same, but when testing, this order gave the correct response (a brand new report getting the highest possible version number). Setting it to be `foo.b - foo.a` resulted in a new report being assigned version number 1. I'll take another look at it and see if something else isn't causing it.
2. We should still be logging errors for templates that had a problem with processing, and could handle them here if we want. My justification for leaving them alone is that the logic here is very limited — access an s3 object that we know exists (because its key came from the S3 api), load its body content, and get a hash of it. We could very easily group the rejected promises and log the errors, if you think that would be useful.
3. This is a thing we should talk about more broadly as a team, but when working with externally controlled types I like to try to convince myself that they are correct, even when they really don't seem to be. The types provided by our version of the AWS SDK are saying that these values might be undefined, so I want to take their word on it. Generally, I avoid `!` whenever I can (learned typescript in a repo with `no-non-null-assertion` turned on), so these types are necessary for handling array filtering in Typescript—it won't pick up changes within objects in an array unless you use a type predicate in the filter function. 

Nits:
1. Yes! Was going to do this as part of the api route change, which is next up.
2. Thank you! Will add.
3. This function will be used in the api route changes, to check if we need to create a new template version
4. Good idea!

benmartin-coforma:
> 3\. This function will be used in the api route changes, to check if we need to create a new template version

That doesn't sound right to me. After the dust has settled from this change, a new template version will only be created if the hash of the latest _stored_ version does not match the hash of the version in the API. That's one hash comparison, not an entire table scan.

Unless I'm misunderstanding, or missing an edge case?

oweingrod-coforma:
> > 3. This function will be used in the api route changes, to check if we need to create a new template version
> 
> That doesn't sound right to me. After the dust has settled from this change, a new template version will only be created if the hash of the latest _stored_ version does not match the hash of the version in the API. That's one hash comparison, not an entire table scan.
> 
> Unless I'm misunderstanding, or missing an edge case?

No, you're right! I'll move that block of code.

oweingrod-coforma:
@benmartin-coforma @BearHanded ready for re-review

BearHanded (approved):

benmartin-coforma (approved):

PR #11316, opened 2023-06-16
MDCT-2629: Clear Reports in Admin Selector Dashboard 

karla-vm:
### Description

Currently, reports by state are being "cached" in the report context for Admin users, so whenever an Admin switches from one state to another (e.g. AL to MN), they will see any reports of the previous state for a bit before rendering the second states' reports. 

Here we're clearing the reports by state from context whenever an Admin user returns to the Admin Dashboard Selector page.

### Related ticket(s)

MDCT-2629

---
### How to test

- Log in as two distinct state users, for example `stateuser` and `stateuser1`
- Create a new `MLR` report for each state (don't worry about filling it in or anything)
- Log in as a `adminuser`
- Navigate to the first state, view the report
- Return to the state selection dashboard
- Navigate to the second state

You shouldn't see the report from the first state on screen, not even for a **second** 👀 

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/forms/AdminDashSelector.tsx
+++ b/services/ui-src/src/components/forms/AdminDashSelector.tsx
@@ -1,9 +1,9 @@
-import { useState } from "react";
+import { useContext, useEffect, useState } from "react";
 import { useNavigate } from "react-router-dom";
 import { useFlags } from "launchdarkly-react-client-sdk";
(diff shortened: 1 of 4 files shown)


ntsummers1 (approved):
Tested locally and works great!

PR #11325, opened 2023-06-22
Snyk update

ateklu7:
Two workflow Jobs

Runs during PR creation on main and outputs the findings in GitHub Workflow console
Runs on Cron and creates JIRA ticket based on findings

JIRA ticket behavior

Existing ticket check is set to default 60days, this can be adjusted as needed.
The same Vulnerability can re-exist after the resolution when a the same package is added for this reason there is a check for Closed or Cancelled ticket and based on that a ticket get created.
SNYK Ignore

There is a .snyk file in the root directory that can hold a list of snyk ignore ids, a vulnerability can be manually added there following the same synaxt in the file, so when Snyk scan runs all the vulnerabilities in .snyk file will be ignored.

Important updates

Author checklist

 I have performed a self-review of my code
 I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary
 I have updated relevant documentation, if necessary


.github/workflows/snyk-test.yml
```diff
@@ -15,101 +15,45 @@ jobs:
...
-        env:
-          JIRA_BASE_URL: qmacbis.atlassian.net
-          JIRA_API_TOKEN: ${{ secrets.JIRA_SERVICE_USER_TOKEN }}
-          JIRA_USER_EMAIL: ${{ secrets.JIRA_SERVICE_USERNAME }}
+        - name: use the custom github  action to parse Snyk output
+          uses: Enterprise-CMCS/macfc-security-scan-report@v1.0.2
+          with:
+              jira-username: ${{ secrets.JIRA_SERVICE_USERNAME }}
+              jira-token: ${{ secrets.JIRA_SERVICE_USER_TOKEN }}
+              jira-host: 'qmacbis.atlassian.net'
```
oweingrod-coforma, line 50: Is it possible to add the MCR application tag here?

`jira-custom-fields: '{ "customfield_14154" : [{"id": "16958", "value": "MCR"}] }'`

oweingrod-coforma (changes requested):

.github/workflows/snyk-test.yml
ateklu7: the tag has been added

oweingrod-coforma (approved):

PR #11332, opened 2023-06-28
Patch Dependabot Alert #162: fast-xml-parser vulnerable to Prototype Pollution through tag or attribute name

karla-vm:
### Description

[Dependabot #162](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/security/dependabot/162)

### Related ticket(s)

N/A

---
### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-auth/package.json
+++ b/services/ui-auth/package.json
@@ -9,6 +9,8 @@
   "author": "",
   "license": "CC0-1.0",
   "dependencies": {
+    "@aws-sdk/client-sts": "3.362.0",
+    "@aws-sdk/client-cognito-identity-provider": "3.362.0",
     "fast-xml-parser": "^4.2.5"
   }
 }
--- a/services/ui-auth/package.json
+++ b/services/ui-auth/package.json
@@ -8,10 +8,7 @@
   },
   "author": "",
   "license": "CC0-1.0",
-  "devDependencies": {},
   "dependencies": {
-    "@aws-sdk/client-cognito-identity-provider": "3.362.0",
-    "@aws-sdk/client-sts": "3.362.0",
-    "fast-xml-parser": "4.1.2"
+    "@aws-sdk/client-sts": "3.362.0"
   }
 }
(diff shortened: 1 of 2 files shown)


benmartin-coforma (commented):
Question: It looks like `@aws-sdk/client-cognito-identity-provider` 3.362.0 depends on `@aws-sdk/client-sts` 3.362.0, which in turn depends on `fast-xml-parser` 4.2.5. I would have assumed that updating our version of client-cognito-identity-provider would _automatically_ pull in those dependencies, on `yarn install`.

If that's true, then I think we wouldn't need to list the child dependencies explicitly in our package.json? Which would also prevent the weirdness of explicitly requiring fast-xml-parser 4.1.2, while implicitly requiring version 4.2.5.

In short: could this be a one-line change to package.json, rather than a three-line change?

karla-vm:
> Question: It looks like `@aws-sdk/client-cognito-identity-provider` 3.362.0 depends on `@aws-sdk/client-sts` 3.362.0, which in turn depends on `fast-xml-parser` 4.2.5. I would have assumed that updating our version of client-cognito-identity-provider would _automatically_ pull in those dependencies, on `yarn install`.
> 
> If that's true, then I think we wouldn't need to list the child dependencies explicitly in our package.json? Which would also prevent the weirdness of explicitly requiring fast-xml-parser 4.1.2, while implicitly requiring version 4.2.5.
> 
> In short: could this be a one-line change to package.json, rather than a three-line change?

great call, Ben! I'll change that 🙏🏽

benmartin-coforma (approved):
Looks great! Thank you!

benmartin-coforma:
Oh no, I've cursed the deploy:

```
Deploying ui-auth to stage dependabot-162 (***)
Bundling with Webpack...
ERROR in ../../../services/ui-auth/libs/cognito-lib.js 5:4-56
Module not found: Error: Can't resolve '@aws-sdk/client-cognito-identity-provider' in '/home/runner/work/macpro-***/macpro-***/services/ui-auth/libs'
 @ ../../../services/ui-auth/handlers/createUsers.js 2:0-50 26:10-31 28:10-32 30:10-41

webpack compiled with 1 error
```

benmartin-coforma (approved):

oweingrod-coforma (approved):

PR #11350, opened 2023-07-01
Separate the latest formTemplates from the UI

benmartin-coforma:
### Description
We are working on moving form templates from the frontend to the backend, but there are many pieces in the UI that directly depend on knowledge of the latest form templates. This PR works to increase the separation, in a few ways.

1. Previously, the application generated routes based on the most recent form templates, for all report types. With this change, the application will generate routes based on the form template of the currently-selected report.
2. The Header and the main SkipNav behave differently based on whether we are on a report page (that is, whether the current route is defined by a form template). Previously, this determination was made based on the most recent form templates, for all report types. With this change, the determination will be made based on the form template of the currently-selected report.
3. To facilitate the Header and main SkipNav changes, the ReportContext will be extended to include a new flag, `isReportPage`. The main SkipNav will be encapsulated in a new component, MainSkipNav, so that it can access the ReportProvider.
4. The data we send to Tealium also differs based on whether we are on a report page, which was previously determined the same way as for the Header and main SkipNav. With this change, that determination will be a best guess, based on whether the current path begins with a report type.
5. Previously, the form template included both an array of nested routes, and a flattened array of those same routes. This effectively doubled the size of the JSON stored in form template S3 buckets, and effectively halved the maximum size of future reports. With this change, the flat routes will be computed on the frontend when a report loads. This may have adverse performance effects on page loads and report saves, but hopefully not too bad.
6. When creating a report, the UI includes in the API payload the most recent form template for that report type. This is precisely the behavior we want to change. _However_, this PR does not address it. What this PR will change is how the UI _gets_ the latest form template. Previously, it was passed through several layers, down to the AddEditReportModal. With this change, AddEditReportModal will pull it from a new utility file, formTemplate.ts. In the future, formTemplate.ts will be deleted entirely, and AddEditReportModal will no longer include a formTemplate in the createReport payload.
7. The ReportPageIntro component accepts any number of props, and passes them through to a Box component, which passes them through to a div. Previously, reportType was one of the props we were passing as a part of this props blob, which resulted in a console error every time we loaded the component. With this change, reportType will be dignified as an expected prop of ReportPageIntro, and will therefore not be passed through to the Box. This wasn't part of the linked story, but the console error was bothering me.

### Related ticket(s)
MDCT-2601

---
### How to test
This PR made changes to the routes of the application, and changes to report creation. It would be worthwhile to ensure reports can still be created, navigated around, saved, and loaded. Report routes are now flattened on the fly whenever a report is created, selected, fetched, updated, submitted, archived, or released.

Pressing tab on any page (without any already-focused elements) should bring up the SkipNav in the top left corner. It should behave appropriately on report pages and on non-report pages. (This is existing functionality)

The page header should include a "Leave Form" button when on a report page, and it shouldn't include one when not on a report page. (This is existing functionality)

The console shouldn't have any errors like "React doesn't recognize the prop \`reportType\` on a DOM element".

### Important updates
n/a

---
### Author checklist


oweingrod-coforma (approved):
Small nit on a test file but otherwise good! 

I recommend we override CC here. -0.1% of coverage is fine.

ajaitasaini (dismissed):
talked through the changes together, lgtm!

services/ui-src/src/utils/other/formTemplate.test.ts
```diff
@@ -0,0 +1,49 @@
+import {
+  mockFlattenedReportRoutes,
+  mockReportJson,
+} from "utils/testing/setupJest";
+import {
+  compileValidationJsonFromRoutes,
+  copyAdminDisabledStatusToForms,
+} from "./formTemplate";
+
+describe("Test copyAdminDisabledStatusToForms", () => {
```
oweingrod-coforma, line 10: I ended up moving this code to the backend as part of that refactor—will move these tests into my PR once this is merged

oweingrod-coforma (dismissed):

services/ui-src/src/utils/other/formTemplate.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    };
+    const result = copyAdminDisabledStatusToForms(mockAdminDisabledReportJson);
+
+    const testStandardPageForm = result.routes[0].form;
+    const testDrawerPageForm = result.routes[1].children![0].drawerForm!;
+    const testModalDrawerPageModalForm =
+      result.routes[1].children![1].modalForm!;
+    const testModalDrawerPageDrawerForm =
+      result.routes[1].children![1].drawerForm!;
+    const testModalOverlayPageForm = result.routes[1].children![1].modalForm!;
```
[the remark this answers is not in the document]
benmartin-coforma: Fixed. Since this file is about to be deleted (moved to the api), it would be worth adding this test to your branch for the backend work.

services/ui-src/src/utils/other/formTemplate.test.ts
benmartin-coforma: Oh, perfect!

ajaitasaini (dismissed):

benmartin-coforma:
This PR has been superseded by (#11393)[https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/11393]. This works around a resource rename constraint from CloudFormation.

PR #11370, opened 2023-07-12
MDCT-2464: Zustand State Management + Admin Banner Form

karla-vm:
### Description

This is an initial proof of concept for using [Zustand](https://docs.pmnd.rs/zustand/getting-started/introduction) as a state management library going forward. 

For this POC, a store has been created with an initial banner state (`undefined` by default) and there are two (2) functions: one for setting a new admin banner and one for clearing it. Another option is using the `setAdminBanner` to take an `undefined` value to clear banners, and both work exactly the same.

In my opinion, it was quite simple to set up (even with TypeScript) and I was able to plug in the new store functions/actions in place of the `useState` actions without changing much. The biggest lift I'd see is organizing our actions within a single store, but I've seen documentation and examples of folks having multiple stores for different contexts (in our case, we could have one store for banners, another store for form fields, etc.) and maintaining them. 

ALSO, there is no extra setup for unit testing! 💯 
_And_ I used `Redux DevTools` as part of this POC, which is a plug-and-play feature.

See [SPIKE document here](https://docs.google.com/document/d/1Raj0Z_XEwWI_ZT5qoX5Odb4bAp1nZ4IDNlrVieYQ-0w/edit#heading=h.m6o98tnzl5s9)

### Related ticket(s)

MDCT-2464

---
### How to test

- Login as admin
- Create a new banner! ✨ 
- Clear that banner! 🚫 

### Important updates

`zustand` has been installed at `ui-src`

---
### Author checklist

---

### DIFF
--- a/services/database/package.json
+++ b/services/database/package.json
@@ -25,15 +25,13 @@
     "typescript": "^4.9.3"
   },
   "dependencies": {},
-  "resolutions": {
-    "xml2js": "^0.5.0"
-  },
   "overrides": {
     "serverless-dynamodb-local": {
       "dynamodb-localhost": "https://github.com/99x/dynamodb-localhost#db30898f8c40932c7177be7b2f1a81360d12876d"
     }
   },
   "resolutions": {
-    "luxon": "3.3.0"
+    "luxon": "3.3.0",
+    "xml2js": "^0.5.0"
   }
 }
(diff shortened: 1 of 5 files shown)


benmartin-coforma:
I like it!

ajaitasaini (commented):
looks great! like that it has a simple setup

keeysnc (approved):
Nice! Lgtm

PR #11374, opened 2023-07-14
[MDCT-2712]: Update fields so they no longer trigger validations between same page types

ntsummers1:
### Description

Going between 2 pages of the same type would trigger validation during the same user session. This is ideal from a UX perspective but if I user goes between 2 different page types, like a ModalDrawer or ModalOverlay and then a StandardPageType, validation wouldn't trigger. This leads to an experience that isn't standard and inconsistant. We've decided to remove the validation triggering between the same page type and to only have it trigger from Review and Submit for now. Going forward we'll reimplement this in a standard way.

Initial bug:

https://github.com/Enterprise-CMCS/macpro-mdct-mcr/assets/19439679/24acf0d1-0f07-4abb-9cf3-3934fd54627c

New Fix:

https://github.com/Enterprise-CMCS/macpro-mdct-mcr/assets/19439679/f330e634-6588-4f64-a106-18c42ae7f6c9

### Related ticket(s)

MDCT-2712

---
### How to test

Create a MCPAR Report
Go between multiple pages and see that no validation triggers just by visiting it.
Clicking a page in Review and Submit page, however, does trigger validation and calls it out until a user updates those fields.

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/drawers/ReportDrawer.test.tsx
+++ b/services/ui-src/src/components/drawers/ReportDrawer.test.tsx
@@ -10,6 +10,7 @@ import {
   mockEmptyDrawerForm,
   mockModalDrawerReportPageVerbiage,
   mockStateUser,
+  RouterWrappedComponent,
 } from "utils/testing/setupJest";
 // utils
 import { useUser } from "utils";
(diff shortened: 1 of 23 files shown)


ajaitasaini (approved):
the double negative of **dontReset = false** threw me for a loop 😅 but looks good to me!

ntsummers1:
> the double negative of **dontReset = false** threw me for a loop 😅 but looks good to me!

I agree. I'm gonna be refactoring a lot of this going forward so I'm gonna note this and make it a priority to fix then.

PR #11375, opened 2023-07-18
Switch to a better-maintained fork of serverless-dynamodb-local

benmartin-coforma:
### Description
In production, our software runs on the cloud. But we use a set of libraries to simulate the cloud for local development. Two of those libraries, used to run DynamoDB locally, have fallen badly out of date. This PR will switch us to forks of both.

[`serverless-dynamodb-local`](https://github.com/99x/serverless-dynamodb-local) and its dependency [`dynamo-db-localhost`](https://github.com/99x/dynamodb-localhost) are published by [99x](https://opensource.99x.io/), but neither package has had an update in over two years. During this time, AWS made two breaking changes.
1. AWS changed the URL of their Dynamo download, from `http` to `https`. This change would be trivial to support on the library's side, and in fact [it has been fixed in the source](https://github.com/99x/dynamodb-localhost/commit/d4546c8110f1d5c2a454988c7e658e2f6a80d502), but that fix was never published. We worked around this in MDCT by overriding the NPM version with one pulled directly from GitHub.
2. AWS changed the [requirements for Access Key IDs](https://repost.aws/articles/ARc4hEkF9CRgOrw8kSMe6CwQ/troubleshooting-the-access-key-id-or-security-token-is-invalid-error-after-upgrading-dynamodb-local-to-version-2-0-or-greater), restricting them from containing underscores. Unfortunately, `serverless-dynamodb-local` mocks this ID with underscores. Once again this issue [has been fixed in source](https://github.com/99x/serverless-dynamodb-local/commit/0a39b1900ff4811c83d25d24d32fbb3aabc2a6bb), but I suspect we will be disappointed if we wait for the fix to be published to NPM.

[Raise](https://www.joinraise.org/) has forked both of these libraries ([the serverless plugin](https://github.com/raisenational/serverless-dynamodb) and the [supporting package](https://github.com/raisenational/aws-dynamodb-local)), and has committed to maintaining the forks. If we switch, we can remove our workarounds to consume unreleased fixes of 99x's libraries, and have some level of confidence that future DynamoDB breaking changes will be addressed promptly.

***Reviewers:*** What are some reasons we should  _not_ make this switch? We could continue to pin specific commits on 99x's packages instead.

### Related ticket(s)
n/a

---
### How to test
1. Run the app *locally* (not on cloudfront)
2. Create a report - this should save it to local DynamoDB
3. See your report's name on the dashboard - this is loaded from DynamoDB

### Important updates
n/a - Local dev should "just work" again.

---
### Author checklist


gmrabian (commented):
This time my DB started up correctly, but I got the same error when trying to talk to it. I tried a few different things, like viewing MCPARs as admin, and creating them as state user

`UnrecognizedClientException: The Access Key ID or security token is invalid.`

oweingrod-coforma:
Confirmed this is happening to me on a fresh clone. WEIRD

oweingrod-coforma:
@gmrabian Can you try with the fix I added earlier?

gmrabian (approved):
@oweingrod-coforma yes this fixed it for me! I can save/add/load stuff now!

BearHanded (approved):
LGTM and this gets us through the door at the moment. I don't see another option for now unless another fork becomes more popular or the original libs are resurrected.

cassandradanger:
i think this is a good option! worst case, Raise goes under and deletes this library and we have to revert back to 99x - but since it's for local development, I feel pretty comfortable going with this option!

PR #11394, opened 2023-07-31
Mdct 2650 Update destroy script

gmrabian:
### Description
Update destroy script

### Related ticket(s)
MDCT-2650

---
### How to test
Create a branch and let it deploy
Run this script from your local against that branch
Make sure all resources are cleaned up...in particular
- app-api stack finishes deleting
- api gateway execution logs are gone

### DIFF
--- a/scripts/destroy.sh
+++ b/scripts/destroy.sh
@@ -148,11 +148,27 @@ aws apigateway update-stage \
   echo "Removed certificate from stage and disassociated web acl"
 fi
 
+databaseStackName="database-$stage"
+
 # Trigger a delete for each cloudformation stack
 for i in "${stackList[@]}"; do
-  echo "triggering stack deletion for $i"
-  aws cloudformation delete-stack --stack-name $i
+  # database stack has a bucket that the app-api stack expects to exist while deleting
+  if [ $i == $databaseStackName ]; then
+    echo "skipping database delete until app-api is finished"
+  else
+    echo "triggering stack deletion for $i"
+    aws cloudformation delete-stack --stack-name $i
+  fi
 done
+
+# if database stack exists...
+# wait for app-api to delete and then delete database stack
+if [[ "${stackList[@]}" =~ "$databaseStackName" ]]; then
+  echo "waiting for app-api stack delete to complete before deleting database stack"
+  aws cloudformation wait stack-delete-complete --stack-name "app-api-$stage"
+  echo "triggering stack deletion for $databaseStackName"
+  aws cloudformation delete-stack --stack-name $databaseStackName
+fi
 
 # Delete Client Certificates associated with a branch
 certToDestroy=$(aws apigateway get-client-certificates | grep \"app-api-${stage}\" -B 2 |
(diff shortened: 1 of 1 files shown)


scripts/destroy.sh
```diff
@@ -123,29 +122,54 @@ for i in "${bucketList[@]}"; do
...
   aws s3 rm s3://$i/ --recursive
 done
 
-echo "Removing certificate from stage"
-
 restApiName=$stage-app-api
 
 restApiId=$(aws apigateway get-rest-apis | jq -r ".[] | .[] |  select(.name==\"$restApiName\") | .id|tostring")
 
+if [ "$restApiId" != "" ]; then
```
gmrabian, line 133: add if to prevent silent failing

scripts/destroy.sh
```diff
@@ -123,29 +122,54 @@ for i in "${bucketList[@]}"; do
...
   --rest-api-id $restApiId \
   --stage-name $stage \
   --patch-operations op=replace,path=/clientCertificateId,value="" \
   &>/dev/null
+  restApiArn="arn:aws:apigateway:us-east-1::/restapis/$restApiId/stages/$stage"
+
+  echo "Disassociating web acl from api gateway"
+
+  aws wafv2 disassociate-web-acl \
+    --resource-arn $restApiArn &>/dev/null
```
gmrabian, line 146: disassociate web acl to app-api can delete properly

scripts/destroy.sh
```diff
@@ -123,29 +122,54 @@ for i in "${bucketList[@]}"; do
...
 done
 
+# if database stack exists...
+# wait for app-api to delete and then delete database stack
+if [[ "${stackList[@]}" =~ "$databaseStackName" ]]; then
+  echo "waiting for app-api stack delete to complete before deleting database stack"
+  aws cloudformation wait stack-delete-complete --stack-name "app-api-$stage"
+  echo "triggering stack deletion for $databaseStackName"
+  aws cloudformation delete-stack --stack-name $databaseStackName
+fi
```
gmrabian, line 171: wait for app-api stack to delete before deleting db stack. the following error happens because the database stack gets deleted and the app-api has lambdas that rely on a bucket from that stack. If we wait, they will all delete properly rather than leaving stacks in the 'delete failed' state
Screenshot 2023-07-31 at 2 15 44 PM

scripts/destroy.sh
```diff
@@ -166,6 +190,13 @@ if [[ -n $apiGatewayLogGroupExists ]]; then
   aws logs delete-log-group --log-group-name $apiGatewayLogGroupName
 fi
 
+# Find hanging api-gateway execution logs
+apiGatewayExecutionLogRegex="^API-Gateway-Execution-Logs[^/]+\/$stage$"
+apiGatewayExecutionLogs=($(aws logs describe-log-groups --query "logGroups[*].logGroupName" --output json | jq -r --arg pattern "$apiGatewayExecutionLogRegex" '.[] | select(test($pattern))'))
+if [[ -n $apiGatewayExecutionLogs ]]; then
+  aws logs delete-log-group --log-group-name $apiGatewayExecutionLogs
+fi
```
gmrabian, line 198: delete hanging log group

BearHanded (approved):

PR #11397, opened 2023-08-02
Redirect fix for IDM login

keeysnc:
### Description
Create an endpoint to handle redirect on successful logout, and all the config that is required.
The endpoint exists outside the normal auth gates, so returning to it doesn't trap you in an endless loop in prod.

* This is similar to the fix that was done on QMR. https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/1911

---
### Important updates

---
### Author checklist

---


karla-vm (approved):

BearHanded (approved):
I think this has all the moving parts in places, vetted the env in the deployed branch and that /postLogout sends the user to the targetted resource. The rest of the config looks right, just a matter of deploying and cleaning up any oddities I think

PR #11405, opened 2023-08-03
backend update to allow for YoY copy

cassandradanger:
### Description

This PR updates the create report handler to accept an optional `copySourceId` request body value.

If `copySourceId` is passed, create handler should:
- Get the field data and form template for report with id
- If the form template version is different than the newest one, compute a list of all possible field names for each form template version
  - Remove any field data keys from the field data that do not exist in the “new” form template.
  - Combine this copied over field data to the newly created field data for the report.
  - Copied over data should not replace the new field data—use object destructuring to combine.
- Finish report creation as normal.

### Related ticket(s)

MDCT-2771
---
### How to test

MCPAR - create a new program, it should work. It should not fail. 
Currently, there's not a front end update to request a report to copy, so that will come next

### Important updates

---
### Author checklist

---


services/app-api/utils/formTemplates/formTemplates.ts
```diff
@@ -31,42 +30,28 @@ export async function getNewestTemplateVersion(reportType: ReportType) {
...
-};
-
 export async function getOrCreateFormTemplate(
   reportBucket: string,
   reportType: ReportType
 ) {
-  const currentFormTemplate = formTemplateForReportType(reportType);
-  const formTemplateWithAdminDisabled =
-    copyAdminDisabledStatusToForms(currentFormTemplate);
+  const currentFormTemplate = reportType === "MCPAR" ? mcparForm : mlrForm;
```
cassandradanger, line 40: should we update this to not be a ternary? future-proof it?

braxex: yeah. it's definitely gonna have to change soon anyway 'cause there's a variant of MCPAR coming soon called PCCM. it won't be a tremendous amount of work but it's gonna be different. i like using a map pattern for this kinda stuff:

```
...
const reportFormMap = {
  MCPAR: mcparForm,
  MLR: mlrForm,
}
const currentFormTemplate = reportFormMap[reportType];
if (!reportFormMap) throw error 
...
```

benmartin-coforma: If you look at the removed code on the left half of this file's diff, you'll see `formTemplateForReportType()`, which is precisely the type of future-proofing you're looking for.

I like it better than the map approach because if `if(!reportFormMap)` check is implicit in the switch-case. Plus, it will be enforced by Typescript instantly whenever a new `ReportType` is added.

benmartin-coforma (changes requested):
This PR would remove some code that is good:
* Use of `getFieldDataKey` in create.ts, and the state validation it requires
* Result unwrapping in `getNewestFormTemplateVersion()` in formTemplates.ts
* `formTemplateForReportType()` in formTemplates.ts
* The warning comment in `isFieldElement()` in formTemplates.ts, and the `isLayoutElement()` function entirely

I assume this is just version control weirdness, because the underlying `create-form-templates-2628` branch has continued to change. And unfortunately, it's not done changing quite yet.

The easiest way to deal with this will probably be to wait for `create-form-templates-2628` to merge into `main` (which should definitely happen today 😓 ), and take things from there.

services/app-api/utils/formTemplates/formTemplates.ts
braxex: Ooh yeah I like that approach. Let's do it. 👍

PR #11409, opened 2023-08-07
backend YoY update

cassandradanger:
### Description

This PR updates the create report handler to accept an optional `copySourceId` request body value.

If `copySourceId` is passed, create handler should:
- Get the field data and form template for report with id
- If the form template version is different than the newest one, compute a list of all possible field names for each form template version
  - Remove any field data keys from the field data that do not exist in the “new” form template.
  - Combine this copied over field data to the newly created field data for the report.
  - Copied over data should not replace the new field data—use object destructuring to combine.
- Finish report creation as normal.

### Related ticket(s)

MDCT-2771

---
### How to test

MCPAR - create a new program, it should work. It should not fail.
Currently, there's not a front end update to request a report to copy, so that will come next

### Important updates

---
### Author checklist

---


services/app-api/handlers/reports/create.ts
```diff
@@ -40,35 +42,37 @@ export const createReport = handler(async (event, _context) => {
   const requiredParams = ["reportType", "state"];
 
   // Return error if no state is passed.
-  if (!hasReportPathParams(event.pathParameters!, requiredParams)) {
+  if (
+    !event.pathParameters ||
+    !hasReportPathParams(event.pathParameters, requiredParams)
```
gmrabian, line 47: I'm getting a little mixed up on this logic. So we are wanting path parameters to not exist, but also passing them into a function that checks if they do exist (to then make sure they don't)?

services/app-api/utils/reports/index.ts
```diff
@@ -0,0 +1,76 @@
+import { getPossibleFieldsFromFormTemplate } from "../formTemplates/formTemplates";
```
gmrabian, line 1: nit: we don't use put logic in `index` files anywhere else (I don't think). We use them as folder aggregate files, so that we can more easily import from within the folder (see `app-api/utils/types/index` as an example). Just trying to be consistent

services/app-api/utils/reports/index.ts
```diff
@@ -0,0 +1,76 @@
...
+import s3Lib, { getFieldDataKey } from "../s3/s3-lib";
+import { AnyObject, State } from "../types";
+
```
gmrabian, line 4: not sure where best to put this, but does this function account for if a field type changes but the key does not? For example: we change a text box to text area, or we change a radio to a checkbox, or radio to text area. We would probably not (?) want to copy in that case, right?

services/app-api/utils/reports/index.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+    Object.keys(entity).forEach((entityKey) => {
+      if (!possibleFields.includes(entityKey)) {
+        if (
+          !entityKey.includes("name") &&
+          !["key", "value"].includes(entityKey)
+        ) {
+          delete entityData[index][entityKey];
+        }
+      }
```
gmrabian, line 66: could the nested if be simplified?
are we checking if the entityKey does not include the keywords 'name', 'key', or 'value'? What is the reasoning?

gmrabian (commented):
some comments for clarification

services/app-api/handlers/reports/create.ts
cassandradanger: The `hasReportPathParams` function checks for certain required parameters, but doesn’t affect the type inference of `event.pathParameters` . Adding an additional check for missing all path parameters entirely allows us to use fewer optionals in the handler.

services/app-api/utils/reports/index.ts
cassandradanger: per the spike: we probably wouldn't be altering a question - if a question needed updating, we would create a new question and any other validation changes to fields would fail on the front end.

karla-vm (approved):
LGTM!

services/app-api/utils/reports/reports.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  if (sourceFieldData) {
+    const possibleFields = getPossibleFieldsFromFormTemplate(formTemplate);
+    Object.keys(sourceFieldData).forEach((key: string) => {
+      // Only iterate through entities, not choice lists
+      if (Array.isArray(sourceFieldData[key])) {
+        pruneEntityData(
+          sourceFieldData,
+          key,
+          sourceFieldData[key],
```
BearHanded, line 33: Open question: Is there a reason sourceFieldData and sourceFieldData[key] are passed in here, rather than sourceFieldData, and a key, and letting the pruneEntityData know how to navigate that?

Do we see it always operating on an object and a known index in that object, or are we preparing for a case where two different objects are compared?

cassandradanger: oren says "We need to know the key to be able to check for and delete top level keys based on name", and this isn't too big of a deal

so i'm gonna resolve and merge, thank you!

PR #11429, opened 2023-08-15
[MDCT-2822] update with new rounding field

ajaitasaini:
### Related ticket(s)

MDCT-2822

---
### How to test

- [Login](https://d1pq7j5x6f2dig.cloudfront.net/)
- For both MCPAR and MLR, confirm that rounding is matching with [rounding sheet](https://docs.google.com/spreadsheets/d/1j6NdmOfIQ09Zb3JWFXXADNcXUhpD-ZVpYydjlPVDPec/edit#gid=0)

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/app-api/forms/mcpar.json
+++ b/services/app-api/forms/mcpar.json
@@ -712,7 +712,7 @@
                   "label": "C1.I.5 Program enrollment",
                   "hint": "Enter the total number of individuals enrolled in the managed care program as of the first day of the last month of the reporting year.",
                   "mask": "comma-separated",
-                  "decimalPlacesToRoundTo": 0  
+                  "decimalPlacesToRoundTo": 0
                 }
               },
               {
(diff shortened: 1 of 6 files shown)


ntsummers1 (dismissed):
Looks great! Wasn't able to get anything to break

services/app-api/forms/mcpar.json
```diff
@@ -2986,7 +3046,8 @@
                 "props": {
                   "label": "D3.VIII.6 Sanction amount",
                   "hint": "Add a dollar amount. Enter N/A for sanctions that do not carry financial penalties.",
-                  "mask": "currency"
+                  "mask": "currency",
+                  "decimalPlacesToRoundTo": 2
```
braxex, line 3050: Request: Rather than having to specify how many decimal places to round to for currency, we should default it to 2 in the code, which would then be overridable by this prop. There are 2 reasons for this:
1. If unspecified, as is the case with all of the current MCPAR reports that are live, it will default to 2 instead of defaulting to whatever the user enters, which is the desired behavior.
2. It's less work when we add lots of currency fields in reports because we don't have to specify 2.
3. Bonus reason: With it defaulting to 2 and customizable to something else, we are able to handle situations where we want to round to whole dollar values, like $1.23 round to $1 or something (think like tax prep software or something), which the BOs may want in the future.

britt-mo (dismissed):

braxex (approved):

ntsummers1 (approved):

PR #11433, opened 2023-08-16
Mdct 2768 YoY Copy frontend

cassandradanger:
### Description

The frontend portion of the YoY Copy feature. 
- Once a report has been submitted, it should show up in the "Select which report you'd like to copy" dropdown, as an option of a report to copy (in the New Program modal)
- Choosing to copy from a previous report or not, creating a new program should work as normal.
- If you copy from a previous report, all answers except "VII: Quality Measures" (we are still waiting on the final list of what should and should not be copied over - hence the feature flag)
- 
Screenshot 2023-08-14 at 4 54 58 PM

### Related ticket(s)

MDCT-2768, MDCT-2771

---
### How to test

- create a new program as normal (no copying)
- submit that program
- create another program, the previous program should show up in the dropdown menu
- choose it and create the new program
- skip to the end to Review & Submit, and see that only the "VII: Quality Measures" option still requires attention, and all the other questions have answers

### Important updates

---
### Author checklist

---


services/ui-src/src/components/reports/ReportProvider.tsx
```diff
@@ -171,6 +176,13 @@ export const ReportProvider = ({ children }: Props) => {
     );
   };
 
+  useEffect(() => {
+    const submittedReports = reportsByState?.filter(
+      (item) => item.status === "Submitted"
+    );
+    setSubmittedReportsByState(submittedReports);
+  }, [reportsByState]);
```
BearHanded, line 184: Does this need to be its own slice of state when it is filtered off another slice of state? I think this conceptually is fine here with react state, but might flag as an antipattern with redux.

See second comment around use case

services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -35,26 +36,31 @@ export const DropdownField = ({
...
-    if (typeof options === "string") {
-      const dynamicOptionValues = report?.fieldData[options];
-      if (dynamicOptionValues) {
-        const fieldOptions = dynamicOptionValues.map((option: EntityShape) => ({
+    if (options === "copyEligibleReports") {
+      dropdownOptions =
+        submittedReportsByState?.map((option) => ({
+          label: `${option.programName} ${convertDateUtcToEt(option.dueDate)}`,
+          value: option.fieldDataId!,
+        })) ?? [];
```
BearHanded, line 53: If this chunk is the only place we're using submittedReportsByState, would it make more sense to just filter for submitted reports here, rather than every time the reports array gets rev'd?

braxex: I think it makes more sense in the ReportContext because I don't want a field component to own something like that. May end up using it in other spots in the future too. Also not a very costly operation to filter them down in the state each time IMO.

services/ui-src/src/components/reports/ReportProvider.tsx
braxex: I like it better here for reasons in other comment :)

BearHanded (approved):

PR #11449, opened 2023-09-01
Update validation for text fields

gmrabian:
### Description
Previously you could submit an optional text field with whitespaces only and it would save to the db and silently fail validation.

Now validation for optional strings checks for whitespace.

https://github.com/Enterprise-CMCS/macpro-mdct-mcr/assets/57802560/8295a6a9-f69d-4eaa-a7c1-341c2ff9951c

Open to suggestions on making the code logic cleaner! It works as such
| String has length | String is only whitespace | Desired outcome |
| ----------------- | ------------------------- | ----------------- |
| empty                   | N/A                                     | valid |
| empty                   | N/A                                     | valid |
| has length            | is only whitespace            | not valid |
| has length            | is not only whitespace     | valid |

empty strings here are considered valid because we apply this logic to both optional and required fields, where the latter already has a `.required()` on its schema 

### Related ticket(s)

MDCT-2799

---
### How to test

- Login as state user
- Navigate to MLR, create new submission, edit
- Go to MLR Reporting
- Add program reporting information
- Verify that the text boxes that are required validate as expected (must be filled, cannot be spaces only)
- Verify the optional text box (last one) validates as expected (can be empty, cannot be spaces only)
- Verify the report cannot be created if you have spaces only in a box

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/utils/validation/schemas.test.ts
+++ b/services/ui-src/src/utils/validation/schemas.test.ts
@@ -6,6 +6,8 @@ import {
   validNumber,
   numberNotLessThanOne,
   numberNotLessThanZero,
+  text,
+  textOptional,
 } from "./schemas";
 
 describe("Schemas", () => {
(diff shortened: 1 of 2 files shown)


ntsummers1 (approved):

benmartin-coforma (approved):

britt-mo (approved):
Looked good in the browser

ajaitasaini (approved):
approving but i did have a question - what's the difference between the first and second row in the table you provided?

PR #11451, opened 2023-09-01
Add maxWidth to dashboard program name

gmrabian:
### Description

Add `maxWidth` property to the program name column such that it doesn't cause the page to scroll left and right at smaller sizes

https://github.com/Enterprise-CMCS/macpro-mdct-mcr/assets/57802560/bde8ff2e-8e19-4879-8285-3c802a333c8d

Open to suggestions!
If you remove `width` and only use `maxWidth` the width changes as you resize and the text shifts, which I think is not desirable

### Related ticket(s)

MDCT-2824

---
### How to test

- Login as state user
- Go to MCPAR
- Make a report with a really long name
- Resize the window to near the tablet/desktop breakpoint and verify the window does not push any information off the screen

---
### Author checklist

---


ntsummers1 (approved):
Clean

PR #11459, opened 2023-09-08
Mdct 2864 yoy specific

cassandradanger:
### Description

This PR is to update the YoY feature to only copy over the specific fields requested by stakeholders (can be found in the document linked in the ticket)

### Related ticket(s)

MDCT-2864

---
### How to test

- create and submit a MCPAR report
- create a new MCPAR report, copying that previously submitted report
- Go through the copied report and verify against the document linked in the Jira ticket

### Important updates

---
### Author checklist

---


gmrabian (approved):
looks good to me! 

I noticed the `reportType` got added to create and is now part of the copy process. Will this still work for existing reports that don't have the `reportType` attribute?

gmrabian (dismissed):

services/app-api/utils/reports/reports.ts
benmartin-coforma, line 1: Regarding the `// TODO: there has to be a better way` comment, I started putting together a suggestion inline but got a bit carried away. I think the problem would become easier to think about if we changed `pruneEntityData` from an imperative procedure to more of a functional one, which returns the pruned array. Similarly, the in-place pruning of individual entity objects can be made functional. The result would look something like this:

```ts
function prunedEntityArray(
  entityData: AnyObject[],
  possibleFields: string[]
) {
  function pruneEntityObject (entity: AnyObject): AnyObject {
    if (entity.key?.includes("state_excludedEntityIdentifiedInFederalDatabaseCheck")) {
      return {};
    }
    if (entity.key?.includes("program_prohibitedAffiliationDisclosure")) {
      return {};
    }
    return Object.fromEntries(Object.entries(entity).filter(isCopyableEntry));
  }

  // Delete any key existing in the source data not valid in our template, or any entity key that's not a name.
  function isCopyableEntry ([key, _value]: [string, any]): boolean {
      // Entities have "name" and "id" keys that are not accounted for in the form JSON. This carveout ensures we never remove them.
      if (["key", "value", "name", "id"].includes(key)) {
        return true;
      }
      if (possibleFields.includes(key)) {
        return true;
      }
      if (MCPARFieldIDBlocklist.wildcard.some((x) => key.toLowerCase().includes(x))) {
        return false;
      }
      if (MCPARFieldIDBlocklist.matchString.includes(key)) {
        return false;
      }
      return true;
  }

  return entityData
    .map(pruneEntityObject)
    .filter(entry => Object.keys(entry).length > 0);
}
```

That  reduces the parameter list from 4 to 2, which is nice, but it also omits the logic to delete the whole array if nothing is left. I think the best spot for that would be up a level, around line 38:
```ts
      if (Array.isArray(sourceFieldData[key])) {
        const prunedArray = prunedEntityArray(sourceFieldData[key], possibleFields);
        if (prunedArray.length === 0) {
          // Delete the whole key if there's nothing in it
          delete sourceFieldData[key];
        }
        else {
          sourceFieldData[key] = prunedArray;
        }
      }
```

benmartin-coforma: Please treat the above as an IDEA and not a REQUEST - if you'd rather leave it as-is, or do something else, I leave it to you.

services/app-api/utils/constants/constants.ts
```diff
@@ -55,3 +55,67 @@ export const bucketTopics = {
...
+    "sanction_dollarAmount",
+    "sanction_assessmentDate",
+    "sanction_remediationCompleted",
+    "sanction_remediationDate",
+    "sanction_correctiveActionPlan",
+    "qualityMeasure_reportingPeriod",
+    "qualityMeasure_reportingPeriodStartDate",
+    "qualityMeasure_reportingPeriodEndDate",
+  ],
+  wildcard: ["appeals", "grievances", "sanctions"],
```
benmartin-coforma, line 120: Question: I don't understand the purpose of the wildcard section here. Do we have auto-generated field IDs that can't be predicted? And if so, could the matching logic be made any more specific (For example, `key.startsWith(wildCard)` vs `key.includes(wildCard)`)? As it is, I'm a little bit worried about false positives.

services/app-api/utils/reports/reports.ts
```diff
@@ -15,32 +16,43 @@ export async function copyFieldDataFromSource(
...
+  validatedFieldData: AnyObject,
+  reportType: ReportType
 ) {
   const sourceFieldData = (await s3Lib.get({
     Bucket: reportBucket,
     Key: getFieldDataKey(state as State, copyFieldDataSourceId),
   })) as AnyObject;
 
-  if (sourceFieldData) {
+  if (sourceFieldData && reportType === ReportType.MCPAR) {
```
benmartin-coforma, line 27: Nit: When 90% of the body of a function is wrapped in an `if`, I tend to prefer early returns for the negative cases. For example, we could check the report type before even touching S3:
```ts
if (reportType !== ReportType.MCPAR) {
  return validatedFieldData;
}
```
The same style of "I don't have any work to do, so I'm bailing" early return could be used again to de-nest the rest of the function:
```ts
const sourceFieldData = ...
if (!sourceFieldData) {
  return validatedFieldData;
}
const possibleFields = ...
```

services/app-api/utils/reports/reports.ts
keeysnc: I was checking this out as well and I like Ben's idea, I was a bit curious though if `entityData` could be converted to an Array it may be a bit easier to map and filter out whats needed. Seems like thats similar to whats going on here though.

services/app-api/utils/constants/constants.ts
cassandradanger: basically, it's cutting down on listing out each id in the blocklist. According to the document, it sounds like the stakeholders will never want to copy over anything that regards sanctions, appeals or grievances due to the nature of the data. so we're not super concerned about false positives.
Is that cool? Or do you think it's still worth it to update the blocklist with each ID to ignore

benmartin-coforma: For me, that raises more questions. With the current code:
1. We would exclude 11 of the 12 fields in the Appeals Overview drawerForm - but we would include `plan_ltssUserFiledCriticalIncidentsWhenPreviouslyFiledAppeal`, as it doesn't include `appeals` exactly.
   * Although this is remediated by including that field in the matchString array
2. Similarly, we would exclude 4 of the 5 fields in the Grievances Overview drawerForm - but we would include `plan_ltssUserFiledCriticalIncidentsWhenPreviouslyFiledGrievance`, as it doesn't include `grievances` exactly.
   * Although this is also remediated by including that field in the matchString array
3. We include `program_standardAppealTimelyResolutionDefinition` and `program_expeditedAppealTimelyResolutionDefinition` - would those be intentional?
4. We include `program_grievanceTimelyResolutionDefinition` - probably not intentional?
5. The "sanctions" wildcard does nothing, as it doesn't appear in any field IDs
6. We include `sanction_interventionType-othertext` and `sanction_interventionTopic-othertext`
   * Although this could be fixed by changing the `sanctions` wildcard to `sanction` only
   * And if we did that, we may as well remove the sanction-specific IDs from the matchString array

I'm wondering if it would make more sense to include/exclude fields based on which page or form contains them - that would be a significant code change at this point, as it would require considering the form template rather than just the IDs. But with the current implementation, even if we patched up all these little edge cases, it really lends itself to creating more in the future. As the report changes over time, I anticipate bug reports like "this whole page copied from the previous year, except this one question, why?" and "this page wasn't supposed to copy from the previous year, except this one nested question did, why?"

BearHanded (dismissed):

gmrabian (dismissed):

britt-mo (approved):

BearHanded (approved):

PR #11460, opened 2023-09-13
Use an allowlist to decide which fields to copy

benmartin-coforma:
### Description
This PR contains three main changes:
1. Uses a field allowlist instead of a blocklist. This means that the default behavior for new fields (if we ever add any) will be _not_ to copy them. This is probably a safer choice than copying by default, because year-over-year copy is a feature we added after the product's initial release.
2. Lists every field explicitly. This is a net loss in terms of lines of code, but a win in terms of complexity. If a maintainer wants to know whether a particular field will be copied, they will be able to learn that by searching for the exact field ID. The copy algorithm is simpler too.
3. Distinguishes between entities and radio buttons/check box lists. This is done by breaking up the allowlist, one sublist per entity. This removes the need for any special case code for individual fields.
4. Copies fields one-by-one. This was just a little easier for me to reason about than the approach of deleting specific fields from the copy source before calling `Object.assign` on the remainer.

### Related ticket(s)
MDCT-2684

---
### How to test
1. Fill out a report completely
2. Create a new report, using the first as a copy source
3. Examine the new report, ensuring that all & only the correct fields are populated.

### Important updates
n/a

---
### Author checklist


services/app-api/utils/constants/constants.ts
```diff
@@ -56,66 +56,103 @@ export const bucketTopics = {
...
+    "plan_medicalLossRatioReportingPeriodEndDate",
+    "program_encounterDataSubmissionTimelinessStandardDefinition",
+    "plan_programIntegrityReferralPath",
+    "plan_smaMfcuConcurrentProgramIntegrityReferrals",
+    "plan_smaMfcuAggregateProgramIntegrityReferrals",
+    "plan_beneficiaryCircumstanceChangeReportingFrequency",
+  ],
+  qualityMeasures: [
+    "id",
+    "name", // TODO do quality measures have names? Delete this line if not.
```
BearHanded, line 137: Assuming this todo is for the future, not for today?

BearHanded (approved):

services/app-api/utils/constants/constants.ts
benmartin-coforma: It's today if you (or @cassandradanger ) know the answer to the question. I don't and I'm kind of swamped. There's no harm if `"name"` stays in this list despite not existing on entities - it's just less tidy.

cassandradanger: i don't believe accessMeasures, qualityMeasures,  have a "name" key, just "id" and "label"

cassandradanger (commented):
only issue i found is that these two items were copied over when they shouldn't be
A.5a Reporting period start date
A.5b Reporting period end date

gmrabian (commented):
Love the code cleanup! Overall no issues. Found one field that didn't copy:

`qualityMeasure_plan_measureResults` is in the allowlist but I did not see it copied over because it had this uuid appended to it

Screenshot 2023-09-13 at 4 45 00 PM

gmrabian:
> only issue i found is that these two items were copied over when they shouldn't be A.5a Reporting period start date A.5b Reporting period end date

I noticed this as well, but then I realized they had the updated dates and should be there (because you have to provide them when creating a program in the dashboard). Did you use the same dates for both reports? because that would make it look like it was copied

services/app-api/utils/constants/constants.ts
gmrabian: I did not see a `name` key in them either when I attempted to fill out everything I could

cassandradanger:
> > only issue i found is that these two items were copied over when they shouldn't be A.5a Reporting period start date A.5b Reporting period end date
> 
> @cassandradanger I noticed this as well, but then I realized they had the updated dates and should be there (because you have to provide them when creating a program in the dashboard). Did you use the same dates for both reports? because that would make it look like it was copied
> 
> If you make a report and change the dates is it as expected?
yup - just realized this right now - this is populates from the Create Program modal 🤦

cassandradanger:
> Love the code cleanup! Overall no issues. Found one field that didn't copy:
> 
> `qualityMeasure_plan_measureResults` is in the allowlist but I did not see it copied over because it had this uuid appended to it
> 
> Screenshot 2023-09-13 at 4 45 00 PM


The quality measure results shouldn't be copied over, so the end result is correct - but it shouldn't be in the allow list

PR #11466, opened 2023-09-18
YoY copy: show alert msg if reporting period is not available

cassandradanger:
### Description

add additional alert messaging for Quality & Performance Measures (per design)

"Not Answered" - when reporting period hasn't been reported
"Add the Measure Reporting Period for this quality and performance measure by editing the measure." -  if the measure still needs to be updated

Screenshot 2023-09-18 at 1 59 55 PM

### Related ticket(s)

MDCT-2945

---
### How to test- 

- create a copy of a submitted report
- go to the Quality & Performance Measures section
- see that all details about the measure have been copied over except the reporting period
- verify that the alerts regarding the reporting period are showing

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
+++ b/services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
@@ -115,11 +115,11 @@ export const EntityCard = ({
         />
         {openAddEditEntityModal && (
           <>
-            {!formattedEntityData.reportingPeriod ? (
+            {!formattedEntityData.reportingPeriod && (
               <Text sx={sx.missingReportingPeriodMessage}>
                 {verbiage.missingReportingPeriodMessage}
               </Text>
-            ) : null}
+            )}
             <Button
               variant="outline"
               size="sm"
--- a/services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
+++ b/services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
@@ -114,15 +114,22 @@ export const EntityCard = ({
           printVersion={!!printVersion}
         />
         {openAddEditEntityModal && (
-          <Button
-            variant="outline"
-            size="sm"
-            sx={sx.editButton}
-            leftIcon={<Image src={editIcon} alt="edit icon" height="1rem" />}
-            onClick={() => openAddEditEntityModal(entity)}
-          >
-            {verbiage.editEntityButtonText}
-          </Button>
+          <>
+            {!formattedEntityData.reportingPeriod ? (
+              <Text sx={sx.unfinishedMeasureMessage}>
+                {verbiage.unfinishedMeasureMessage}
+              </Text>
+            ) : null}
+            <Button
+              variant="outline"
+              size="sm"
+              sx={sx.editButton}
+              leftIcon={<Image src={editIcon} alt="edit icon" height="1rem" />}
+              onClick={() => openAddEditEntityModal(entity)}
+            >
+              {verbiage.editEntityButtonText}
+            </Button>
+          </>
         )}
         {entityStarted || entityCompleted || printVersion ? (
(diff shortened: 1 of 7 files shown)


services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
```diff
@@ -114,15 +114,22 @@ export const EntityCard = ({
...
-            onClick={() => openAddEditEntityModal(entity)}
-          >
-            {verbiage.editEntityButtonText}
-          </Button>
+          <>
+            {!formattedEntityData.reportingPeriod ? (
+              <Text sx={sx.unfinishedMeasureMessage}>
+                {verbiage.unfinishedMeasureMessage}
+              </Text>
+            ) : null}
```
braxex, line 122: Nit: Could use logical AND here instead of a ternary that returns `null`.

```suggestion
            {!formattedEntityData.reportingPeriod && (
              <Text sx={sx.unfinishedMeasureMessage}>
                {verbiage.unfinishedMeasureMessage}
              </Text>
            )}
```

braxex (dismissed):
Looks good to me. As mentioned, I'm curious if design would like anything changed in the PDF/export view (I'm guessing so).

services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
benmartin-coforma, line 122: Question: We only want this text to display for quality measures without reporting periods, but this code is also used to display other entity types - access measures, sanctions and so on - none of which have reporting periods. So on first read, it seems like this code would incorrectly display the unfinished measure message for those entity types as well. But when I tested in Cloudfront, I didn't see that problem.

It took me a while to figure out that this code _does_ display an unfinished measure message for other entity types - except that those entities don't have any verbiage defined for that message, so we're actually only displaying an empty paragraph.

This isn't a huge issue, but there are two effects. For one, the code is a little puzzling (at least, it took me a minute). For two, that paragraph does reserve a bit of extra space on the page despite being empty, due to its margin - which could be considered a styling bug. This screenshot shows how much space we're talking about:

image

I don't think this is blocking, but maybe the check could be made more specific? But I'm not sure what the best way to do that would be, because this is a ticket that really tests our separation of verbiage and component functionality.

braxex: Do we pass in the entity type at all here? If so, we could pretty easily add that as a condition for the text to show.

cassandradanger: I update this to be a && instead of a ternary - does that address this issue?

cassandradanger: ah yup, that was an oversight on my part - updated to check the entityType

benmartin-coforma (approved):
Looks good!

benmartin-coforma (approved):

services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
```diff
@@ -114,15 +114,23 @@ export const EntityCard = ({
...
-            variant="outline"
-            size="sm"
-            sx={sx.editButton}
-            leftIcon={<Image src={editIcon} alt="edit icon" height="1rem" />}
-            onClick={() => openAddEditEntityModal(entity)}
-          >
-            {verbiage.editEntityButtonText}
-          </Button>
+          <>
+            {entityType == "qualityMeasures" &&
```
braxex, line 118:
```suggestion
            {entityType == ModalDrawerEntityTypes.QUALITY_MEASURES &&
```

braxex (approved):
Looks great!

benmartin-coforma (approved):

PR #11470, opened 2023-09-22
update completion status requirements

cassandradanger:
### Description

David found a little bug with the YoY completion status.
On the VII: Quality Measures page, if a report was copied over, the Reporting Period and the Details are not copied over and show incomplete alerts. I made an update to create the Reporting Period error message, but it wasn't being taken into account for the cards completed status (the icon in the top left corner)

Screenshot 2023-09-22 at 2 56 53 PM

### Related ticket(s)

MDCT-2988

---
### How to test

Create a report copy
navigate to VII: Quality Measures
update the details of the measure (but not the reporting period)
see that the completion status is still incomplete
update the Reporting Period
that should fulfill all the requirements and the completion status should be Complete

this should be the case for the PDF as well

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
+++ b/services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
@@ -64,6 +64,8 @@ export const EntityCard = ({
     default:
       break;
   }
+  const entityDetailsAndReportingPeriodComplete =
+    entityCompleted && reportingPeriodCompletedOrOptional;
   return (
     <Card {...props} marginTop="2rem" data-testid="entityCard">
       <Box sx={sx.contentBox} className={printVersion ? "print-version" : ""}>
@@ -75,7 +77,7 @@ export const EntityCard = ({
         {!printVersion ? (
           <Image
             src={
-              entityCompleted && reportingPeriodCompletedOrOptional
+              entityDetailsAndReportingPeriodComplete
                 ? completedIcon
                 : unfinishedIcon
             }
@@ -93,14 +95,14 @@ export const EntityCard = ({
           >
             <Image
               src={
-                entityCompleted && reportingPeriodCompletedOrOptional
+                entityDetailsAndReportingPeriodComplete
                   ? completedIcon
                   : unfinishedIcon
               }
               alt={`entity is ${entityCompleted ? "complete" : "incomplete"}`}
               sx={sx.printVersionIcon}
             />
-            {entityCompleted && reportingPeriodCompletedOrOptional ? (
+            {entityDetailsAndReportingPeriodComplete ? (
               <Text className="completed-text">Complete</Text>
             ) : (
               <Text className="error-text">Error</Text>
--- a/services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
+++ b/services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
@@ -81,13 +81,17 @@ export const EntityCard = ({
                 ? completedIcon
                 : unfinishedIcon
             }
-            alt={`entity is ${entityCompleted ? "complete" : "incomplete"}`}
+            alt={`entity is ${
+              entityDetailsAndReportingPeriodComplete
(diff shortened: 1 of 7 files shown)


benmartin-coforma (changes requested):
This doesn't feel like the best solution. If the end goal is to have an extra check for completedness, which only applies to quality measures, it seems like the best spot would be with the rest of the quality-measure-specific code, in the switch-case on entityType.

Specifically, I think we could add the reportingPeriod check here (currently lines 59-61):
```ts
        entityCompleted =
          entityStarted &&
          formattedEntityData?.reportingPeriod &&
          validPerPlanResponses?.length === report?.fieldData?.plans?.length;
```

Then the rest of the changes in the file, if I'm understanding correctly, would be unnecessary.

cassandradanger:
> This doesn't feel like the best solution. If the end goal is to have an extra check for completedness, which only applies to quality measures, it seems like the best spot would be with the rest of the quality-measure-specific code, in the switch-case on entityType.
> 
> Specifically, I think we could add the reportingPeriod check here (currently lines 59-61):
> 
> ```ts
>         entityCompleted =
>           entityStarted &&
>           formattedEntityData?.reportingPeriod &&
>           validPerPlanResponses?.length === report?.fieldData?.plans?.length;
> ```
> 
> Then the rest of the changes in the file, if I'm understanding correctly, would be unnecessary.

Hey @benmartin-coforma ! I tried this at first, but the Details alert is dependent on the EntityComplete - so that alert was still showing after the Details were updated - so that's why I pulled out the reportingPeriod validation

benmartin-coforma:
@cassandradanger I see! I didn't understand the ramifications there - it's more complicated than I realized.

benmartin-coforma (commented):
I'm retracting my change request, with apologies!

services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
```diff
@@ -83,11 +92,15 @@ export const EntityCard = ({
...
+              src={
+                entityCompleted && reportingPeriodCompletedOrOptional
+                  ? completedIcon
+                  : unfinishedIcon
+              }
               alt={`entity is ${entityCompleted ? "complete" : "incomplete"}`}
               sx={sx.printVersionIcon}
             />
-            {entityCompleted ? (
+            {entityCompleted && reportingPeriodCompletedOrOptional ? (
```
gmrabian, line 103: suggestion: since `entityCompleted && reportingPeriodCompletedOrOptional` is used three times we could write it as a variable in this file for simplicity

gmrabian (approved):
verified locally. looks good!

gmrabian (approved):
lgtm

services/ui-src/src/components/cards/EntityCard/EntityCard.tsx
benmartin-coforma, line 1: I like this a lot better with the dedicated variable!

We do have a situation, now, where the alt text of the icon might not match the icon itself. Specifically, lines 84 and 102 should now refer to `entityDetailsAndReportingPeriodComplete` rather than `entityCompleted`.

Similarly, in the print version, the className of the Box containing the icon may not match the underlying icon. So line 90 should also reference the new variable rather than `entityCompleted`. That className affects the margin, so I think it would have shown up as a styling bug (in the PDF only).

braxex: Suggestion: It looks like `entityCompleted` is being used in a lot of places now, but it's name is now kind of a lie, right? As I understand it, `reportingPeriodCompletedOrOptional` has to also be truthy now for an entity to truly be "complete". Seems like maybe we should move that check into the switch statement where we're setting `entityCompleted` so that the original variable's name is real, and probably check everywhere in this file that it's being referenced to make sure we aren't using different metrics for completeness inn different places and can avoid bugs.

benmartin-coforma (approved):

PR #11478, opened 2023-10-05
Fix blank header cells for screen readers

gmrabian:
### Description

Adds descriptive labels to the table columns that have no visible header

Uses [Chakra's Visually Hidden component](https://chakra-ui.com/docs/components/visually-hidden)

Please comment with better ideas for carrying and parsing the hidden names. I considered just a text string like `hidden.Actions` and checking for that with `includes` but the object seems more...formal. Less hacky. I'm sure there is an even smarter way.

### Related ticket(s)

MDCT-3031
MDCT-2981

---
### How to test

Launch and login as state user
Go to MLR dashboard
Add a report if there are none
Turn on your MacOS screen reader
  Settings -> Accessibility -> VoiceOver -> Toggle On
From here you can test one of two ways:
1. Refresh the page and let VoiceOver start reading all content
  - When it gets to the empty columns (first and last) it should announce something relevant to that column
  - "Edit report name" for first column and "Actions" for last column
2. Click on or near the dashboard table (use tabs to move focus as needed)
  - Once in the table, use `ctrl` + `option` + `arrow keys` to move around
  - When you are at the heading or underneath the first or last column it should announce the column name (I think if you start at the top and go straight down it will not, but if you switch columns side to side it will)

Can also verify review and submit page has "actions" as its third header for the page table

Please reach out to me if you want to learn more about how to test this!

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/tables/Table.tsx
+++ b/services/ui-src/src/components/tables/Table.tsx
@@ -13,7 +13,11 @@ import {
 // utils
 import { sanitizeAndParseHtml } from "utils";
 // types
-import { AnyObject, TableContentShape } from "types";
+import {
+  AnyObject,
+  ScreenReaderOnlyHeaderName,
+  TableContentShape,
+} from "types";
 
 export const Table = ({
   content,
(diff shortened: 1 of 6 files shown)


BearHanded (approved):

karla-vm (approved):
💯 LGTM!

PR #11481, opened 2023-10-10
Add PCCM radio to MCPAR dashboard create modal

gmrabian:
### Description

At this time we have only added the radio and its metadata. It does not affect the actual report at all. That will be covered separately.

Feedback needed:
We probably want to make this disabled (read-only) after creation, right? So it should only be editable when first creating a report but not if you go back to change the title or dates.

Open to different naming ideas for any variables

### Related ticket(s)

MDCT-3027

---
### How to test

Create a MCPAR report as a state user
Verify the new field is in the DB using dynamodb admin `DYNAMO_ENDPOINT=http://localhost:8000 dynamodb-admin`
Verify the radio option hydrates in the frontend as expected

Create MLR reports and verify they are the same and create normally

---
### Author checklist

---

### DIFF
--- a/services/app-api/utils/types/reports.ts
+++ b/services/app-api/utils/types/reports.ts
@@ -1,4 +1,4 @@
-import { FormJson, ModalOverlayReportPageShape } from "./formFields";
+import { Choice, FormJson, ModalOverlayReportPageShape } from "./formFields";
 import { AnyObject, CompletionData, CustomHtmlElement, State } from "./other";
 
 // REPORT STRUCTURE
@@ -151,7 +151,7 @@ export interface MCPARReportMetadata extends ReportMetadata {
   reportingPeriodEndDate: number;
   dueDate: number;
   combinedData: boolean;
-  programIsPCCM: AnyObject[];
+  programIsPCCM: Choice[];
 }
 
 /**
--- a/services/app-api/utils/types/reports.ts
+++ b/services/app-api/utils/types/reports.ts
@@ -151,6 +151,7 @@ export interface MCPARReportMetadata extends ReportMetadata {
   reportingPeriodEndDate: number;
   dueDate: number;
   combinedData: boolean;
+  programIsPCCM: AnyObject[];
 }
 
 /**
--- a/services/ui-src/src/components/modals/AddEditReportModal.tsx
+++ b/services/ui-src/src/components/modals/AddEditReportModal.tsx
@@ -79,6 +79,7 @@ export const AddEditReportModal = ({
     const reportingPeriodEndDate = convertDateEtToUtc(
       formData["reportingPeriodEndDate"]
     );
+    const programIsPCCM = formData["programIsPCCM"];
 
     return {
       metadata: {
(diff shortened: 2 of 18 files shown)


services/ui-src/src/components/modals/AddEditReportModal.tsx
```diff
@@ -89,6 +90,7 @@ export const AddEditReportModal = ({
         combinedData,
         lastAlteredBy: full_name,
         copyFieldDataSourceId: copyFieldDataSourceId?.value,
+        programIsPCCM,
```
gmrabian, line 93: adds field to payload for saving to backend

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
```diff
@@ -144,6 +144,7 @@ export const DashboardPage = ({ reportType }: Props) => {
           ),
           combinedData: report.combinedData,
           copyFieldDataSourceId,
+          programIsPCCM: report?.programIsPCCM,
```
gmrabian, line 147: adds field to hydration data for reopening to edit

services/app-api/utils/types/reports.ts
```diff
@@ -151,6 +151,7 @@ export interface MCPARReportMetadata extends ReportMetadata {
   reportingPeriodEndDate: number;
   dueDate: number;
   combinedData: boolean;
+  programIsPCCM: AnyObject[];
```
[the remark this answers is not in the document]
karla-vm: I think this works as an `AnyObject`, does this need to be an array or can it be a single object?

BearHanded:
Do we need this behind a feature flag for the time being?

gmrabian:
> Do we need this behind a feature flag for the time being?

That's a good question and I think the answer is yes, we do. Not sure how that would work for the backend stuff then 🤔 would we set the types to optional or would we hardcode that value to false?
That, or we could stack the next set of changes on this branch and not merge them until we're ready

benmartin-coforma (commented):
I approve of the work so far, but before merging we need either a feature flag or a green light to go live.

The YoY Copy feature flag was implemented by duplicating `addEditMcparReport.json`, and choosing the old or new json depending on the feature flag. We could do something similar here, but we open ourselves up to combinatorial problems if we're not sure when YoY will go live before PCCM - that is, we would have 4 copies of the json instead of 2. (And if we don't clean these up, the next feature will have 8, and so on.)

Feature flagging on the frontend, such that it won't request PCCM behavior from the backend until go-live, allows us to merge in the backend changes unflagged, which is very nice. Not so nice is the work Garrett has put into testing the frontend - I do not know how we would get the flag into our jest & cypress tests, and I'm highly skeptical the effort would be worth it.

karla-vm (dismissed):

gmrabian:
Approvals for these changes were in a separate, stacked [PR](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/11490). We will merge them once we reapprove and are ready for this to move into pre-prod and prod environments.

karla-vm (approved):

PR #11507, opened 2023-10-30
Remove transitive dependency on vm2

benmartin-coforma:
### Description
This PR upgrades our version of `proxy-agent` to `6.3.0`. We don't use `proxy-agent` directly; it is a dependency pulled in by Serverless. And `proxy-agent` depends (also indirectly) on a package called `vm2`, which had a severe vulnerability this past summer.

I don't believe we ever had any security exposure due to `vm2`; it was only used in certain code paths, and I don't believe our use of Serverless ever caused any of those paths to be exercised. Nevertheless, this newer version of `proxy-agent` removes the `vm2` dependency, erasing any possibility of exposure.

Ideally, we would have been able to bump Serverless only. Upgrading the dependency under its feet carries a slight risk of breaking change; our current version of Serverless requests `proxy-agent` version 5-something. However, the `proxy-agent` changelog doesn't list any breaking changes between version 5 and version 6, so I believe we'll be fine there.

### Related ticket(s)
CMDCT-2711

---
### How to test
Smoke testing should suffice here.

### Important updates
n/a

---
### Author checklist

- ~[ ] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary~
- ~[ ] I have updated relevant documentation, if necessary~

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
   "resolutions": {
     "nwsapi": "^2.2.5",
     "xml2js": "0.5.0",
-    "json5": "2.2.3"
+    "json5": "2.2.3",
+    "proxy-agent": "^6.3.0"
   }
 }
--- a/services/ui-src/package.json
+++ b/services/ui-src/package.json
@@ -23,7 +23,6 @@
     "bootstrap": "^5.1.3",
     "date-fns": "^2.26.0",
     "date-fns-tz": "^1.2.2",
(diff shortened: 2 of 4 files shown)


gmrabian (approved):
deployed env is good; did a spot check on a few lambdas and didn't see any error logs. Thanks, Ben!

BearHanded (approved):

PR #11515, opened 2023-11-08
[CMDCT-2] Fix clear prop hiding users answers when they click cancel but come back

ntsummers1:
### Description

If you edit a field when inside of a drawer or a modal and then click cancel, it would end up showing that field as edited (Though the backend still had its original answer). This can be very jarring to the user, so this PR fixes the clear prop thats causing this discrepancy!

### Related ticket(s)

CMDCT-2

---
### How to test ###

Create a MCPAR form
Go to Access Measures (Or any drawer form or modal form you want)
Create an Access Measure
Click edit Access Measure to reopen to modal.
Edit a text field
Click Cancel
Click edit Access Measure again
See your text field has your original answer!

---
### Author checklist

### DIFF
--- a/services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
+++ b/services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
@@ -19,6 +19,7 @@ import {
   entityWasUpdated,
   getEntriesToClear,
   setClearedEntriesToDefaultValue,
+  resetClearProp,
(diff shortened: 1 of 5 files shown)


karla-vm (approved):
This looks good to me!

gmrabian (commented):
I wasn't able to replicate the bug in MCR dev 🐛 😶‍🌫️

PR #11519, opened 2023-11-14
Fix AV scan

gmrabian:
### Description

Our virus scan started failing after upgrading the lambda runtime to node 18 because node 16 and below included aws-sdkv2 and node18+ includes aws-sdkv3

V2 to V3 upgrade suggestion from [this forum](https://stackoverflow.com/questions/74549995/error-cannot-find-module-aws-sdk-in-nodejs-aws-lambda-function) and confirmed by [AWS announcement](https://aws.amazon.com/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/)

Explanation of changes:
- Import changed from all of `aws-sdk` to modulized `@aws-sdk/client-s3`
- v2 [required](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/using-async-await.html) `.promise()` on awaits and [v3 does not](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/using-async-await.html)
- The `client.{commandType}(params)` syntax has changed to
```
const command = new {CommandType}(params)
const response = client.send(command)
```

[AWS-SDKv3 S3 docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/)

### Related ticket(s)

CMDCT-3013

---
### How to test

- Get on ZScaler
- Go into AWS mcr-dev
- Go to S3 -> `uploads-virus-scan-3013-attachments-446712541566`
- Drag and drop a file into the bucket
- Go to lambda -> `uploads-virus-scan-3013-avScan` -> Monitor tab -> View Cloudwatch logs
- Go to latest log
- Confirm it scanned your file successfully
- In the bucket click on your object and verify it has the `CLEAN` tag

Verify MCPAR template was tagged CLEAN
- Login to https://d1oik0hwk47zmh.cloudfront.net/ as a state user
- Click `Download MCPAR Excel Workbook` on the homepage and ensure it downloads the xlsx file

---
### Author checklist

---

### DIFF
--- a/services/uploads/src/antivirus.js
+++ b/services/uploads/src/antivirus.js
@@ -52,7 +52,7 @@ function pathFromObjectKey(s3ObjectKey) {
   return `${downloadDir}/${path.basename(sanitizedFilename)}`;
 }
 
-function downloadFileFromS3(s3ObjectKey, s3ObjectBucket) {
+async function downloadFileFromS3(s3ObjectKey, s3ObjectBucket) {
   if (!fs.existsSync(downloadDir)) {
     fs.mkdirSync(downloadDir);
   }
@@ -70,21 +70,16 @@ function downloadFileFromS3(s3ObjectKey, s3ObjectBucket) {
   };
   const getObject = new GetObjectCommand(options);
 
-  return new Promise((resolve, reject) => {
-    s3.send(getObject)
-      .createReadStream()
-      .on("end", function () {
-        utils.generateSystemMessage(
-          `Finished downloading new object ${s3ObjectKey}`
-        );
-        resolve();
-      })
-      .on("error", function (err) {
-        console.log(err); // eslint-disable-line no-console
-        reject();
-      })
-      .pipe(writeStream);
-  });
+  try {
+    const response = await s3.send(getObject);
+    await response.Body.transformToWebStream().pipe(writeStream);
+    utils.generateSystemMessage(
+      `Finished downloading new object ${s3ObjectKey}`
+    );
+  } catch (err) {
+    utils.generateSystemMessage(`Error downloading new object ${s3ObjectKey}`);
+    throw err;
+  }
 }
 
 async function lambdaHandleEvent(event, _context) {
--- a/services/uploads/src/antivirus.js
+++ b/services/uploads/src/antivirus.js
@@ -14,6 +14,7 @@ const clamav = require("./clamav");
 const s3 = new S3Client();
 const utils = require("./utils");
 const constants = require("./constants");
+const { pipeline } = require("stream/promises");
 
 /**
  * Retrieve the file size of S3 object without downloading.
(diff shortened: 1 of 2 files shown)


services/uploads/src/antivirus.js
```diff
@@ -62,22 +69,19 @@ function downloadFileFromS3(s3ObjectKey, s3ObjectBucket) {
...
+    const response = await s3.send(getObject);
+    const readStream = response.Body.transformToWebStream();
+    await pipeline(readStream, writeStream);
+    utils.generateSystemMessage(
+      `Finished downloading new object ${s3ObjectKey}`
+    );
+  } catch (err) {
+    utils.generateSystemMessage(`Error downloading new object ${s3ObjectKey}`);
+    throw err;
+  }
```
benmartin-coforma, line 84: Gorgeous, huge improvement

services/uploads/src/clamav.js
```diff
@@ -1,25 +1,32 @@
...
 async function listBucketFiles(bucketName) {
+  const listObjects = new ListObjectsV2Command({
+    Bucket: bucketName,
+  });
   try {
-    const listFilesResult = await S3.listObjectsV2({
-      Bucket: bucketName,
-    }).promise();
-
+    const listFilesResult = await S3.send(listObjects);
```
benmartin-coforma, line 29: S3's List Objects is one of those sneaky methods that may truncate the result list and give you a continuation token. If there's a chance there will ever be more than 1000 files in this bucket, we'll need to check for the presence of `listFilesResult.NextContinuationToken`.

benmartin-coforma (commented):
I have not tested, but this looks about right!

services/uploads/src/clamav.js
gmrabian: @benmartin-coforma Good concern! I had not initially considered this, however it looks like there's no need to worry:
This function is only used to list the clamav definition files, which at the moment is five (only four of which get updated now, I think `daily.cld` is no longer used). These files get overwritten when there is an update, so they should never be more than this, plus or minus one
Screenshot 2023-11-15 at 9 24 10 AM

Additionally I checked with the CLI to see if it lists versions, which could make it lengthy. It does not. It only lists the latest version, so again, five items
`aws s3api list-objects-v2 --bucket uploads-main-avscan-446712541566`
Screenshot 2023-11-15 at 9 26 31 AM

karla-vm (dismissed):
Confirmed behavior in `CloudTamer` 💯 
Screenshot 2023-11-15 at 10 32 42 AM

services/uploads/src/clamav.js
benmartin-coforma: Okay, sounds good!

benmartin-coforma (approved):
LGTM!

BearHanded (approved):

karla-vm (approved):

PR #11522, opened 2023-11-15
[CMDC-3045] Add serverless plugins to reduce files on build/deploy - "EMFILE - too many open files error"

BearHanded:
### Description

Adds two recommended plugins for avoiding the `EMFILE - too many open files error`

- [serverless-plugin-common-excludes](https://www.serverless.com/plugins/serverless-plugin-common-excludes)
- [serverless-plugin-include-dependencies](https://www.serverless.com/plugins/serverless-plugin-include-dependencies)

It filters out extra unnecessary files as determined on the above pages. Read through the docs for more info on what is filtered out.

The above thread makes it clear edge cases do exist where functionality will be impacted, so test it locally and in the deployed env to verify.

Despite claiming to not exclude aws-sdk, the plugins still remove it. Whitelisting it individually leads to it having dependencies missed. For now, including the node_modules folder resolves it and its dependencies while still shaking out some of the extra file size bloating the action.

For reference, without these plugins, the createBanner lambda sits at 21MB. With them in this configuration, it shrinks to 9. Unconfigured and failing was between 1-2MB.

This may need continued monitoring, and please try to poke around and break things.

### Related ticket(s)

CMDCT-3045

---
### How to test

Existing behavior should not break on the deployed application

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/app-api/package.json
+++ b/services/app-api/package.json
@@ -26,6 +26,8 @@
     "aws-lambda": "^1.0.7",
     "jest": "^27.4.7",
     "serverless-associate-waf": "^1.2.1",
+    "serverless-plugin-common-excludes": "^4.0.0",
+    "serverless-plugin-include-dependencies": "^5.1.0",
     "serverless-plugin-typescript": "^2.1.4",
     "ts-jest": "^27.1.3",
     "typescript": "^4.5.4"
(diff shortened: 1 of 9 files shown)


gmrabian:
E2E tests failed twice. I don't see how these l libraries would cause it, though just calling it out

karla-vm (approved):

gmrabian (approved):

PR #11527, opened 2023-11-27
User Zustandification™️

karla-vm:
### Description

Converting the `UserContext` state management of MCR from `Context API` to `Zustand` (following the patterns implemented in MFP).

### Related ticket(s)

CMDCT-2794

---
### How to test

Running this locally, you should be able to log in, log out, and manage Banners (as an Admin) as usual.

Optionally: install the [Redux Devtools Chrome extension](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd) or the [Firefox Redux Devtools extension](https://addons.mozilla.org/en-US/firefox/addon/reduxdevtools/)

- Open devtools
- Log in
- You should see the `user` object on the extension ✨

Screenshot 2023-11-27 at 2 31 09 PM

### Important updates

---
### Author checklist

---


benmartin-coforma (approved):
This is a ton of work! Well done.

gmrabian (changes requested):
Requesting changes because: Two E2E cypress tests are failing in GitHub Actions and the same two fail for me locally.

Other notes:
I noticed we use this statement a lot when pulling in the user data to files:
`useStore().user ?? {};`

Is there a way we could make that logic part of the store / do we still need the optional empty object? Just curious since it's used so often

gmrabian (approved):

PR #11533, opened 2023-12-14
Pccm field mods 3136

gmrabian:
### Description

There is a question within the MCPAR report that asks for program type, including PCCM as an option. If a user selects PCCM when creating the report, we want this question to reflect that as well.

I tried to write these functions/pathways so that they could be extensible

The form template creation flow now has an additional step that calls a function to make any necessary modifications to the template for PCCM.
The overall report creation flow now has a branch that makes any necessary modifications to formData for PCCM.

Feedback requested on:
- modifying these two separate things (template and fieldData) in a way that makes sense and keeps them somewhat in sync
- Is this the most logical way to do this?

### Related ticket(s)

CMDCT-3136

---
### How to test

- Log in as a state user
- Create a MCPAR report and select "Yes" in the modal to indicate it's a PCCM report
- Go into the report -> Section C -> I
- Verify that C1.I.3 is disabled and has the PCCM option already selected

---
### Author checklist

---

### DIFF
--- a/services/app-api/utils/formTemplates/formTemplates.ts
+++ b/services/app-api/utils/formTemplates/formTemplates.ts
@@ -316,5 +316,10 @@ export const generatePCCMTemplate = (originalReportTemplate: any) => {
 
 const makePCCMTemplateModifications = (reportTemplate: ReportJson) => {
   // Find Question C1.I.3 Program type in Section C.I and disable it
-  reportTemplate.routes[2].children![0].form!.fields[3].props!.disabled = true;
+  const programTypeQuestion =
+    reportTemplate.routes[2].children![0].form!.fields[3];
+  if (programTypeQuestion.id !== "program_type") {
+    throw new Error("Update PCCM logic!");
+  }
+  programTypeQuestion.props!.disabled = true;
 };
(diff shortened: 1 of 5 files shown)


karla-vm (approved):
This approach makes sense to me.

benmartin-coforma (dismissed):
This is a great approach. The one thing that makes me a little uneasy is that the form template modification feels a little... bashful. Every form template should have a `questionC1I3` - but if `makePCCMTemplateModifications` doesn't find it, it will silently do nothing.

Do you think it would be appropriate to throw an error instead? I won't object either way.

services/app-api/utils/formTemplates/formTemplates.ts
```diff
@@ -308,5 +308,13 @@ export const generatePCCMTemplate = (originalReportTemplate: any) => {
...
 
+  // make additional form modifications as necessary
+  makePCCMTemplateModifications(reportTemplate);
+
   return reportTemplate;
 };
+
+const makePCCMTemplateModifications = (reportTemplate: ReportJson) => {
+  // Find Question C1.I.3 Program type in Section C.I and disable it
+  reportTemplate.routes[2].children![0].form!.fields[3].props!.disabled = true;
```
gmrabian, line 319: at @benmartin-coforma 's suggestion I changed this to attempt to modify the intended field directly. this is more likely to cause issues if the template ever changes, which will bring this to our attention to fix sooner

benmartin-coforma (approved):
Looks good to me!

benmartin-coforma (approved):
Better than ever 🚀

karla-vm (approved):

PR #11557, opened 2024-01-09
[Bug Fix] MCR Reports Access

karla-vm:
### Description

Currently, any user (specifically, the `staterep` user) is able to access reports. The correct behavior is that only admins and state users should have access to reports.

### Related ticket(s)

N/A

---
### How to test

- Log in as `staterep@test.com`
- Verify that you cannot navigate to any MCPAR or MLR routes
- Log in as a state user or admin user
- Verify that you CAN navigate to those routes

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/app/AppRoutes.tsx
+++ b/services/ui-src/src/components/app/AppRoutes.tsx
@@ -48,134 +48,92 @@ export const AppRoutes = () => {
           <Route path="/profile" element={<ProfilePage />} />
           <Route path="*" element={<NotFoundPage />} />
 
-          {/* MCPAR ROUTES */}
-          <Route
-            path="/mcpar"
-            element={
-              userHasAccess ? (
-                <DashboardPage reportType="MCPAR" />
-              ) : (
-                <Navigate to="/" />
-              )
-            }
-          />
-          <Route
-            path="/mcpar/get-started"
-            element={
-              userHasAccess ? (
-                <ReportGetStartedPage reportType="MCPAR" />
-              ) : (
-                <Navigate to="/" />
-              )
-            }
-          />
-          {report?.reportType === ReportType.MCPAR && (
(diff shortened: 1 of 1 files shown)


services/ui-src/src/components/app/AppRoutes.tsx
```diff
@@ -21,12 +21,15 @@ import { Fragment, useContext } from "react";
...
 export const AppRoutes = () => {
-  const { userIsAdmin } = useStore().user ?? {};
+  const { userIsAdmin, userIsEndUser } = useStore().user ?? {};
   const { report, contextIsLoaded } = useContext(ReportContext);
 
   // LaunchDarkly
   const mlrReport = useFlags()?.mlrReport;
 
+  // verify whether a user has the required roles / access to MCPAR, MLR, NAAAR reports
+  const userHasAccess = userIsAdmin || userIsEndUser;
```
BearHanded, line 31: nit: i think it's worth updating this comment mentioning it is possible for a user to be auth'd by idm, but not have either of these, or have the old role that is being removed

benmartin-coforma (commented):
That's a lot of ternaries. I'm not super-familiar with routing, so I'm not sure... but would it be possible to put in an early return? If `!userHasAccess`, return a single wildcard route that redirects to "/"; otherwise, return the routes as usual with no additional checks?

karla-vm:
> That's a lot of ternaries. I'm not super-familiar with routing, so I'm not sure... but would it be possible to put in an early return? If `!userHasAccess`, return a single wildcard route that redirects to "/"; otherwise, return the routes as usual with no additional checks?

great point. I cleaned it up 💯 thanks for the callout.

britt-mo (approved):
lgtm

BearHanded (approved):

PR #11552, opened 2024-01-05
CMDCT-3165: Remove State Rep Role and Reports Attribute from MCR State Users

karla-vm:
### Description

Removing the `State Representative` role from MCR and removing the `custom:reports` attribute from state users.

### Related ticket(s)

CMDCT-3165

---
### How to test

Everything should work as expected 💯 

### Important updates

N/A

---
### Author checklist

- [ ] I have performed a self-review of my code
- [ ] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- [ ] I have updated relevant documentation, if necessary
---

_convert to a different template: [test → val](?expand=1&template=test-to-val-deployment.md)_ | _[val → prod](?expand=1&template=val-to-prod-deployment.md)_

PR #11570, opened 2024-01-29
consolidate state rep to state user role

braxex:
### Description

If a user has the IDM user role "mdctmcr-state-rep" we need to treat them as if they are a state user. This is related to a recent change (#11552) to retire the state rep role, but because there are legacy users who will still have the state rep role in IDM, we need to treat them as state users in the app. 

### Related ticket(s)

[CMDCT-3165](https://jiraent.cms.gov/browse/CMDCT-3165)

---
### How to test

You could try logging in as a state rep user, but we removed that user from the test accounts, so you could just take my word for it I guess?

### Important updates

n/a

---
### Author checklist

- [ ] ~~I have added [thorough](https://shorturl.at/aejkF) tests, if necessary~~
- [ ] ~~I have updated relevant documentation, if necessary~~
---


gmrabian (approved):
I believe this covers the two locations that need to acknowledge the old role

PR #11606, opened 2024-02-28
Update MCPAR PDF Header Tags

keeysnc:
### Description
For Acrobat PDF readers the PDF html headers need to be correct for bookmarking. Some of the headers where also different from what was on the figma mockup. Those have been updated as well.

Note: Per Kim's note on the ticket, Acrobat needs a plugin or additional functionality to even use bookmarks. I'm not completely sure if users have that set up or not but, if they do the html headers are set up for it. 

UPDATES: The ticket has been updated to only display H2 tags (section headers) once to reduce redundancy in screenreaders. 

```Evan: I’m going with Approach 1 - while it may not be ideal for understanding exactly which section a user is in due to <h3> text repeating in multiple sections, I still feel like a user will have a better experience than hearing long <h2> text every time. Users will assume they’re in Section A when hearing the subsequent headings until they get to Section B, so it shouldn’t stump seasoned SR users. Also, SR users tend to navigate via headings so they should hear the new section headings when going through the doc.```

UPDATES 2:
On the PDF page only the `h4` tags from 'Quality Measures, Access Measures, and Sanctions' cards had to be changed to `p` tags because the PDF reader was treating the cards as header sections for bookmarking.

### Related ticket(s)
CMDCT- https://jiraent.cms.gov/secure/RapidBoard.jspa?rapidView=5874&projectKey=CMDCT&view=detail&selectedIssue=CMDCT-3274

---
### How to test
Fill out MCPAR report and view PDF - Inspect the headers and checl them against the tags from the mockup

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/export/ExportedSectionHeading.tsx
+++ b/services/ui-src/src/components/export/ExportedSectionHeading.tsx
@@ -71,7 +71,7 @@ const sx = {
   heading: {
     fontWeight: "bold",
     h2: {
-      fontSize: "xl",
+      fontSize: "2xl",
       margin: "1.5rem 0",
     },
     h3: {
--- a/services/ui-src/src/components/export/ExportedSectionHeading.tsx
+++ b/services/ui-src/src/components/export/ExportedSectionHeading.tsx
@@ -13,7 +13,8 @@ export const ExportedSectionHeading = ({
 }: Props) => {
   const sectionHeading = verbiage?.intro?.exportSectionHeader
     ? verbiage?.intro?.exportSectionHeader
-    : verbiage?.intro?.subsection || heading;
+    : verbiage?.intro?.section;
+  const sectionSubHeader = verbiage?.intro?.subsection || heading;
   const sectionInfo = verbiage?.intro?.exportSectionHeader
     ? null
     : verbiage?.intro?.info;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by karla-vm — APPROVED** [review]

**Comment by jessabean** [comment]
I'm seeing the `h2` headings duplicated throughout the PDF view...is that expected behavior?

Screenshot 2024-03-01 at 9 10 54 AM

**Review by BearHanded — DISMISSED** [review]

**Comment by keeysnc** [comment]
> I'm seeing the `h2` headings duplicated throughout the PDF view...is that expected behavior?
> 
> Screenshot 2024-03-01 at 9 10 54 AM

@jessabean Good catch, I brought this up to Kim. The section title is being pulled into each row because there's multiple subsections with the same "parent" section title name. Were going to reach out to Evan to see the best way to display these section titles (for screen readers). 

It may be a good idea to keep them on each section so screen reader users know what parent section they're in. I'll hold off on merging this until we come to a conclusion.

File: services/ui-src/src/components/export/ExportedSectionHeading.tsx [hunk 1]
```diff
@@ -11,38 +11,48 @@ export const ExportedSectionHeading = ({
...
+          {sectionIntroType === "heading" &&
+          sectionIntroContent !== "Appeals Overview" &&
+          sectionIntroContent !== "Network Adequacy"
+            ? ""
+            : sectionSubHeader}
+        </Heading>
+      ) : null}
+      <Box data-testid="exportedSectionHeading" sx={sx.container}>
+        {sectionInfo && (
+          <Box sx={sx.info}>
```
BearHanded, line 37: should this be a box in a box directly? I'm unfamiliar with this chunk of code but just curious

braxex:
@keeysnc Where do we stand on this PR? Looks like we have a lot of failing checks. Depending on what's going on there, we might have to close the PR and reopen a new one.

keeysnc:
> @keeysnc Where do we stand on this PR? Looks like we have a lot of failing checks. Depending on what's going on there, we might have to close the PR and reopen a new one.

@braxex I pushed the last few updates so its ready. Not sure why its failing but I'll try making a new PR.

PR #11613, opened 2024-03-05
[SPIKE] CMDCT-3361: Conditionally Rendered Field Based on Dropdown Input

karla-vm:
### Description

This is a SPIKE to check the level of effort of adding a conditionally-rendered text field (`Specify a new program name`) on the MCPAR creation form modal. 

Below is the expected behavior:

https://github.com/Enterprise-CMCS/macpro-mdct-mcr/assets/99458559/68637f5b-8781-454e-9978-245e7cddfce3

Next steps would be:
- Check why the `Reports eligible to copy` dropdown enables on change
- Pull the programs list by state (this is just a simple example under `constants.ts`)
- Update unit and E2E tests
- Double-check form validation on this new pattern

Would love any thoughts, feedback, input.

### Related ticket(s)

CMDCT-3361

---
### How to test

- Log in as a state user
- Create a new MCPAR report
- Pick `OTHER` from the list of names
- Input a program name

That's it! ✨ 

### Important updates

---
### Author checklist

---


services/ui-src/src/forms/addEditMcparReport/addEditMcparReport.ts
```diff
@@ -0,0 +1,105 @@
+import { ProgramList } from "../../constants";
+import { DropdownOptions } from "types";
+
+// create dropdown options
+const dropdownOptions: DropdownOptions[] = Object.keys(ProgramList).map(
+  (value) => {
+    return {
+      label: ProgramList[value as keyof typeof ProgramList],
```
ailZhou, line 8: This feels reminiscent of when QMR tied the labels to the ids. Not sure if it'll be the same issue here but something to note.

services/ui-src/src/forms/addEditMcparReport/addEditMcparReport.ts
```diff
@@ -0,0 +1,105 @@
...
+  },
+  fields: [
+    {
+      id: "programName",
+      type: "dropdown",
+      validation: "dropdown",
+      props: {
+        label: "Program name (for new MCPAR)",
+        hint: 'Select from an existing program name. If you need to add a new program or rename a program, select "Other, specify".',
+        options: dropdownOptions,
```
ailZhou, line 33: Curious why you decided to convert this to a component instead of appending "other, specify" to the bottom of a duplicate "copyEligibleReports" dropdown? Or at least following that pattern to avoid having to convert this.

services/ui-src/src/components/modals/AddEditReportModal.tsx
```diff
@@ -208,6 +238,7 @@ export const AddEditReportModal = ({
         formJson={form}
         formData={selectedReport?.fieldData}
         onSubmit={writeReport}
+        onChange={onChange}
```
ailZhou, line 241: This makes sense and I agree with using onChange as the way to trigger the showing/hiding of the dropdown.

services/ui-src/src/components/modals/AddEditReportModal.tsx
```diff
@@ -76,9 +80,35 @@ export const AddEditReportModal = ({
     setForm(customizedModalForm);
   }, [selectedReport, copyEligibleReportsByState]);
 
+  // handle program name selection from MCPAR Program List
+  const onChange = (event: InputChangeEvent) => {
+    if (reportType === ReportType.MCPAR) {
+      // make deep copy of baseline form for customization
+      let customizedModalForm: FormJson = JSON.parse(
+        JSON.stringify(modalFormJson)
```
ailZhou, line 88: This seems redundant? Isn't it doing object -> string -> object? If it's not acting correctly, i think you might just need to cast it.

ailZhou (commented):
Left some comments about a couple of areas that I would like to understand your thought process behind. I'm not also not too familiar with MCR so forgive me if some of it seemed self explanatory.

Note: It works as shown in the mockups (except hiding the textbox but you had noted that) so its functional!

services/ui-src/src/components/modals/AddEditReportModal.tsx
karla-vm: true, I copy-pasted from above and didn't catch that. thanks!

services/ui-src/src/forms/addEditMcparReport/addEditMcparReport.ts
karla-vm: On this one, I don't have a lot of context of how we're pulling in the `copyEligibleReports` but that's a great point. We can potentially keep the list of programs in the database and pull them in similarly!

PR #11614, opened 2024-03-06
Have entity's in the drawer use the id to save to prevent same names overwriting other entities on save

ntsummers1:
### Description

If you had a multiple plans of the same name then it would overwrite the data of the first one whenever you update the others specifically through the drawer. This fixes it to use the ID of the plan instead of the name of the plan.

### Related ticket(s)

CMDCT-

---
### How to test

Create a MCPAR
Create 2 plans of the same name
Go to the Grievances by Reason page and edit the second one
See that it doesn't submit data for the first one!

---
### Author checklist

---


ailZhou (approved):

BearHanded (approved):
verified, git is hard

PR #11617, opened 2024-03-06
3242 PDF Accessibility Header Updates

keeysnc:
### Description

For Acrobat PDF readers the PDF html headers need to be correct for bookmarking. Some of the headers where also different from what was on the figma mockup. Those have been updated as well.

Note: Per Kim's note on the ticket, Acrobat needs a plugin or additional functionality to even use bookmarks. I'm not completely sure if users have that set up or not but, if they do the html headers are set up for it.

UPDATES: The ticket has been updated to only display H2 tags (section headers) once to reduce redundancy in screenreaders.

```Evan: I’m going with Approach 1 - while it may not be ideal for understanding exactly which section a user is in due to <h3> text repeating in multiple sections, I still feel like a user will have a better experience than hearing long <h2> text every time. Users will assume they’re in Section A when hearing the subsequent headings until they get to Section B, so it shouldn’t stump seasoned SR users. Also, SR users tend to navigate via headings so they should hear the new section headings when going through the doc.```

UPDATES 2:
On the PDF page only the h4 tags from 'Quality Measures, Access Measures, and Sanctions' cards had to be changed to p tags because the PDF reader was treating the cards as header sections for bookmarking.

In Report:
Screenshot 2024-03-06 at 10 16 50 PM

In PDF:
Screenshot 2024-03-06 at 10 17 52 PM

Related ticket(s)
CMDCT- https://jiraent.cms.gov/secure/RapidBoard.jspa?rapidView=5874&projectKey=CMDCT&view=detail&selectedIssue=CMDCT-3274

### Related ticket(s)

CMDCT-

---
### How to test

( To check UPDATE 1 )
1. Fill out mcpar report and view PDF
2. Check Headers based on the outline within the ticket CMDCT- https://jiraent.cms.gov/secure/RapidBoard.jspa?rapidView=5874&projectKey=CMDCT&view=detail&selectedIssue=CMDCT-3274

( To check UPDATE 2 )
1. After filling out report, check the card headers from the following sections 'Quality Measures, Access Measures, and Sanctions' ( The card headers should be wrapped in H4 tags)
2. Then go to the PDF and check the same sections ( The card headers should be wrapped in p tags )

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.test.tsx
+++ b/services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.test.tsx
@@ -1,0 +1,56 @@
+import { render, screen } from "@testing-library/react";
+import { EntityCardTopSection } from "./EntityCardTopSection";
+
+jest.mock("utils/state/useStore");
+
+const formattedEntityData = {
+  category: "LTSS-related standard: provider travels to the enrollee",
+  methodFrequency: "Annually",
+  monitoringMethods: [],
+  population: "MLTSS",
+  provider: "Behavioral health",
+  region: "Small counties",
+  standardDescription: "fdsfds",
+  standardType: "Maximum distance to travel",
+};
+
+const entityCardTopSectionComponent = (
+  <EntityCardTopSection
+    entityType="accessMeasures"
+    formattedEntityData={formattedEntityData}
+    printVersion={true}
+  />
+);
+
+describe("Test EntityCardTopSection renders", () => {
+  beforeEach(() => {
+    window.location.pathname === "/mcpar/export";
+  });
+  test("EntityStepCardTopSection renders correctly", () => {
+    render(entityCardTopSectionComponent);
+    expect(screen.getByText("fdsfds")).toBeVisible();
+  });
+
+  test("EntityStepCardTopSection header renders p tag wrapper in report", () => {
+    window.history.pushState(
+      {},
+      "",
+      new URL("/mcpar/export", window.location.origin)
+    );
+    const { container } = render(entityCardTopSectionComponent);
+    const pTag = container.querySelectorAll("p")[0];
+    expect(pTag).toHaveTextContent(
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by karla-vm — DISMISSED** [review]
This looks good to me!

File: services/ui-src/src/components/cards/EntityCard/EntityCardTopSection.test.tsx [hunk 1]
```diff
@@ -0,0 +1,56 @@
...
+describe("Test EntityCardTopSection renders", () => {
+  beforeEach(() => {
+    window.location.pathname === "/mcpar/export";
+  });
+  test("EntityStepCardTopSection renders correctly", () => {
+    render(entityCardTopSectionComponent);
+    expect(screen.getByText("fdsfds")).toBeVisible();
+  });
+
+  test("EntityStepCardTopSection header renders p tag wrapper in report", () => {
```
jessabean, line 34: Smol suggestion: maybe this could be renamed `EntityStepCardTopSection header renders p tag wrapper in PDF report` so it's more clear what the test does?

services/ui-src/src/components/export/ExportedSectionHeading.tsx
```diff
@@ -11,38 +11,48 @@ export const ExportedSectionHeading = ({
...
-            isPdf={true}
-            reportType={reportType}
-          />
-        </Box>
-      )}
-    </Box>
+    <>
+      {sectionSubHeader ? (
+        <Heading as="h3" sx={sx.heading.h3}>
+          {sectionIntroType === "heading" &&
```
jessabean, line 28: This conditional is _just_ long enough that I might consider pulling it out into a function, especially for the hard-coded `sectionIntroContent` checks...It might be nice to name it so it's clear why we are excluding just those ones.

karla-vm (approved):

jessabean (approved):

PR #11623, opened 2024-03-12
CMDCT-3399: PDF Text Overflow (Bug)

karla-vm:
### Description

Currently, in the production environment, there is a text overflow happening that's causing accessibility and readability issues. Adding a `maxWidth` will hopefully fix that.

Screenshot 2024-03-12 at 11 36 17 AM

### Related ticket(s)

CMDCT-3399

---
### How to test

This issue is not happening in `main` or `val`, unfortunately, so we can't replicate it or validate until the fix is up in `prod`

### Important updates

---
### Author checklist

---


keeysnc (approved):
Seems like this should work.

PR #11628, opened 2024-03-14
MDCT-3297 Selected States Dropdown BugFix

keeysnc:
### Description

For admin and internal users - the landing page "Select states" dropdown is greyed out in certain conditions. This issue seems to only occur in Prod.

If you enter a report there's a `selectedReport` value stored in localStorage. When you logout without leaving the report that value seems to still be there when the user logs back in. The dropdown seems to work when removing that value. 

I added a method to just remove the `selectedReport` value to try and avoid regression conflicts. 

### Related ticket(s)

CMDCT-https://jiraent.cms.gov/browse/CMDCT-3297

---
### How to test
This issue has only been found on Prod as a admin or internal user.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/utils/auth/UserProvider.tsx
+++ b/services/ui-src/src/utils/auth/UserProvider.tsx
@@ -36,14 +36,21 @@ export const UserProvider = ({ children }: Props) => {
   // state management
   const { user, showLocalLogins, setUser, setShowLocalLogins } = useStore();
 
+  // Clear selectedReport from localStorage when logging out of report.
+  const clearSelectedReportCache = () => {
+    const selectReportCacheExists = localStorage.getItem("selectedReportType");
+    selectReportCacheExists && localStorage.removeItem("selectedReportType");
+  };
+
   // initialize the authentication manager that oversees timeouts
   initAuthManager();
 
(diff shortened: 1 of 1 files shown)


karla-vm (approved):
confirmed that this works!

services/ui-src/src/utils/auth/UserProvider.tsx
```diff
@@ -36,12 +36,19 @@ export const UserProvider = ({ children }: Props) => {
...
+
   // initialize the authentication manager that oversees timeouts
   initAuthManager();
 
   const logout = useCallback(async () => {
     try {
       setUser(undefined);
+      clearSelectedReportCache();
       await Auth.signOut();
       localStorage.clear();
```
ailZhou, line 53: There's a localStorage.clear() down here that should remove all the items from localStorage which is confusing to me of why you would need to do an individual remove?

ailZhou: got an explanation from @keeysnc, i now understand why

ailZhou (approved):
approved after getting a breakdown from Nate

PR #11653, opened 2024-04-03
Remove outdated custom types in favor of package provided types

gmrabian:
### Description

- Custom types like `DynamoGet` and `S3Put` can be replaced with the AWS SDK V3 provided equivalents
- `convertToDynamoExpressionVars` is not used anywhere and has been overtaken by the sdk provided `*/lib-dynamodb`

I organized the imports at the top of the files since we usually put external imports up there. should I leave it in the `// types` section? in places like the dynamodb-lib the imports contain clients and types so not sure if we want import organization consistency or intention consistency (if that even makes sense)

### Related ticket(s)

no ticket (continuing cleanup after AWS SDK V3 upgrade)

---
### How to test

- All tests pass
- No degradation to app

---
### Pre-review checklist


karla-vm (approved):

benmartin-coforma (approved):
🚀🚀🚀

PR #11683, opened 2024-05-06
Upgrade MCR to Node 20

gmrabian:
### Description

- Upgrade `.nvmrc` to node 20
- Upgrade services to node 20
- Upgrade dependencies that need it for compatability
- Remove redundant dependencies

Note: there are three lambdas that aren't on Node 20. Two of them are created by serverless to link resources as a workaround for...itself. The other one is created by the warmup plugin. The warmup one will likely be removed when we migrate to serverless compose, and the serverless ones we have no control over. We might be able to narrow down where they come from with an infrastructure audit, but it just runs once on deploy and then sits there. It'll only be a concern if AWS removes support for Node16, and in that case everyone using serverless will need a solution, so they should upgrade it by then (right...?)

### Related ticket(s)

CMDCT-2656

---
### How to test

Deployed url: https://d20ag9emgh3qad.cloudfront.net/
Everything works as expected
Lambdas don't fail

---
### Pre-review checklist


karla-vm (approved):
tested in the deployed instance, everything looks good!

PR #11686, opened 2024-05-10
Alleviate flaky cypress tests

gmrabian:
### Description

- Update Cypress to 12.7.4
  - [Cypress 12.7.4 changelog](https://docs.cypress.io/guides/references/changelog#12-17-4) - updates webpack to v5, which our app uses ([under the hood with react-scripts](https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/package.json#L74)). Minor bumps to other underlying versions.

- Clean up funky code
  - We had duplicate code that used different commands, and had commands that were missing the closing square bracket `]` so I aligned those.

- Config updates
  - Bump retries to 2

- Clear session state on each test
  - Cypress thinks sessions ["more accurately represent real-world scenarios"](https://docs.cypress.io/api/commands/session#Switching-sessions-inside-tests). However, we do not have a scenario where a user would log in as both a state user and an admin, and somehow toggle back and forth. Thus clearing sessions when switching logins is the best way to ensure expected behavior for each user (the right fields are enabled/disabled)

- Rip all the cucumber stuff out
  - Remove .feature files and move .stepdef.js to .cy.js
  - Rework old step definitions to work as tests
  - Consolidate MLR form and release tests
  - Add accessibility tests per standard page type
  - Update config
  - Remove packages and cleanup imports

### Related ticket(s)

CMDCT-3602
CMDCT-3337

---
### How to test

No changes to deployed env: https://d2mpyb45d6mx0c.cloudfront.net/

All cypress tests pass every time

I ran the tests six times in a row on [this commit](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/actions/runs/9067847403/job/24918025806) and they passed every time.

---
### Pre-review checklist


ntsummers1 (approved):
image

PR #11690, opened 2024-05-15
Automate .env process for local development

dwhitestratiform:
### Description
As part of the workspace setup/standardization initiative one thing that was aimed to accomplish was to make the .env maintenance  and use less manual. Currently there is no guarantee that the .env_exmaple checked into source has any level of accuracy as its not used in any capacity. The goal was to automate this and not require engineers to pass around updates by hand as well as pull secrets from a secure place in an automated fashion. 

There were multiple considerations on how to approach this. One approach involved using aws ssm and pulling values at deploy time (locally) to populate into the .env.  The downside of that is for development purposes as engineers onboard they lack access to aws and there are times where individuals credentials to eua are locked out, etc ... 
The other approach that was considered was to just use 1Password which the team uses already. This adds another dependency to the development process but allows the team to manage. For the time being we've gone with the 1Password approach.

**How this works:**

1) Engineers need to have a 1Password account through Coforma IT and need to have 1Password CLI installed. You can install the 1Password CLI by running the new workspace setup script [here](https://github.com/Enterprise-CMCS/macpro-mdct-tools/blob/main/mdct-setup.sh) 
2) Secrets are populated into the `mdct-devs` vault in 1Password.  For each application a secure note is used and individual secrets are populated into that secure note. From there we can copy references to the secret from 1Password. See instructions here ... https://developer.1password.com/docs/cli/secrets-config-files/
3) We've replaced the .env_example with a .env.tpl that directly references 1Password. So, when you run `./run local --update-env` or `./run update-env` the "op" references reach out to 1Password pull the values in and inject them into a .env file which is used. We're doing this for both .`env` and `/services/ui-src/.env `

**Things to note:** 
1) 1Password is not required to run the app. If you are an external team member or for some reason cannot get access to 1Password an individual can reach out to the team for a .env file and appropriate values to run the app locally. The default behavior remains the same. However, to automate this we can either run `./run local --update-env` when running our local environment to hit 1password on every deploy. Or, for running tests we can run `/.run update-env` in which case the tests look at the top level .env file for values previously exported by hand.
2) Making changes to the .env .... If we in the future need to update the .env we should make changes to the .env.tpl, update secrets or add secrets into the appropriate 1Password note and thats it. After its checked in users will simply consume this through source control and automation. 
3) You no longer need to provide test passwords to run tests locally 
4) **For this to work you'll need to run yarn in your repo to pull in dotenv otherwise the translation from the .env will break**
5) **.env files are still git ignored and always should be... no secret values should ever be checked in**

**Things to consider:** 
-we can merge this now as this keeps default behavior ... however, as devs want to run these commands I would suggest that we install 1password cli (and other tools) using workspace setup. 

### Related ticket(s)
https://jiraent.cms.gov/browse/CMDCT-3605

---
### How to test
1) install 1Password CLI through brew or run the worksapce setup script linked above. 
2) run `./run local --update-env`  or run `./run update-env`. To run e2e tests run `./run update-env` then `yarn test` at the top level of the directory 
3) confirm values are outputting to your .env and /services/ui-src/.env as expected 

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


gmrabian (approved):
this looks good and works well on my end! I like the ease up getting .env populated with 1Password, especially using the biometric auth flow

gmrabian (dismissed):

gmrabian (changes requested):
issue with running cypress tests, cypress attempts to input the comment with the password string. looks like:
`passwordstring# pragma: allowlist secret`
and then auth fails
(this is only a problem locally because GHA uses the github secret values)

berryd (approved):

gmrabian (approved):

PR #11691, opened 2024-05-15
Move destroy to run command

gmrabian:
### Description
All of our applications have been leaking infrastructure due to the reliability of the destroy mechanism.  In the past it's been a rather manual process to clean up latent and orphaned infrastructure in order to keep account costs under control.  Luckily, our ephemeral environments and very light weight and cheap to operate, but manually cleaning up orphaned infrastructure doesn't scale.

The purpose of this PR is retrofit an existing solution into MCR. We have already seen this working in QMR.

Notable changes:
- The dev script (executed by ./dev), has been renamed to run (executed by ./run)
  - The previous commands are unaffected so `./run local` is equivalent to prior functionality
- The destroy.sh script has been removed
- Issuing destroy locally is accommodated by executing `./run destroy --stage [stage]`

### Related ticket(s)

CMDCT-3640

---
### How to test

Evidence of this successfully executing can be seen [here](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/actions/runs/9100484932)
(It failed on the first try but then passed on re-run, so something to watch out for. I think QMR, SEDS, and CARTS all passed on the first time. I [ran it again against a different branch](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/actions/runs/9101636747/job/25019308175) just to see and it passed the first time.)

---
### Pre-review checklist

- ~[ ] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary~


berryd (commented):
Yay, git mv....looks good to me...should we update the version of the destroyer elsewhere since I used an older version?

berryd (approved):

gmrabian:
> Yay, git mv....looks good to me...should we update the version of the destroyer elsewhere since I used an older version?

I checked the other apps and they pulled in 2.1.1 because of the ^ anyway so I think we are fine!
https://github.com/Enterprise-CMCS/macpro-mdct-qmr/blob/master/yarn.lock#L3826-L3827

PR #11719, opened 2024-06-03
Types Cleanup

karla-vm:
### Description

Cleaning up the `types` across `ui-src` and `app-api` for better understanding in the future.

### Related ticket(s)

N/A

---
### How to test

There should be no functional changes in this PR.

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

---


services/ui-src/src/types/other.ts
```diff
@@ -1,13 +1,95 @@
...
+  isBase64Encoded: boolean;
+  path: string;
+  pathParameters: Record<string, string | undefined> | null;
+  queryStringParameters: Record<string, string | undefined> | null;
+  multiValueQueryStringParameters: Record<string, string | undefined> | null;
+  stageVariables: Record<string, string | undefined> | null;
+  /** The context is complicated, and we don't (as of 2023) use it at all. */
+  requestContext: any;
+  resource: string;
+}
```
gmrabian, line 43: The frontend shouldn't care about this type I don't think

services/ui-src/src/types/other.ts
```diff
@@ -1,13 +1,95 @@
...
+    };
+    object: {
+      key: string;
+      size: number;
+      eTag: string;
+      versionId?: string | undefined;
+      sequencer: string;
+    };
+  };
+  glacierEventData?: S3EventRecordGlacierEventData | undefined;
```
gmrabian, line 92: frontend should not care about these either

services/ui-src/src/types/other.ts
```diff
@@ -34,8 +116,27 @@ export interface TimeShape {
...
+  as?: string;
+  props?: AnyObject;
+  children?: any[];
+}
+
+export const enum TemplateKeys {
+  MCPAR = "templates/mcpar-reporting-template.xlsx",
+  MLR = "templates/mlr-reporting-template.xlsx",
+  NAAAR = "templates/naaar-reporting-template.xlsx",
+}
```
gmrabian, line 132: hmm what is the use case for these in the frontend?

gmrabian (commented):
I know some of these are unused and may be used in the future, but just want to be careful about what we assume is important to both sides. ex: I think ReportMetadata should absolutely be in both and should match. That is an essential part of the fe/be contract. `S3EventRecordGlacierRestoreEventData` though? If the FE ever needs that one we are doing something wrong.

Do we have a plan for the next phase? Is there a way to tighten up the ones we need in both and the ones we don't?

gmrabian (approved):

PR #11784, opened 2024-07-10
CMDCT-3763: ILOS PDF Update

karla-vm:
### Description

The ILOS section of MCPAR has an edge case of nested entity responses, as seen below:

Screenshot 2024-07-10 at 3 45 44 PM

Because of the additional nesting, this response wasn't showing up on the PDF and this PR addresses that issue.

### Related ticket(s)

CMDCT-3763

---
### How to test

To enable the ILOS sections, set [this variable}(https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/eb2f0a4fd74a7a105509d907764b09f4095665a9/services/ui-src/src/components/modals/AddEditReportModal.tsx#L44) to `true`.

- Log in as a state user
- Create a new MCPAR report
- Add plans
- Add an ILOS (or however many you'd like 💯 , I'm not your boss)
- Navigate to `/mcpar/plan-level-indicators/ilos`
- Select "Yes, at least 1 ILOS is offered by this plan" and enter some data for each of the ILOS you've added
- Navigate to "Review & submit" and click on **Review PDF** 

You should now something like this, depending on your responses:

Screenshot 2024-07-10 at 3 44 12 PM

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/utils/other/export.test.tsx
+++ b/services/ui-src/src/utils/other/export.test.tsx
@@ -5,6 +5,7 @@ import {
   parseFormFieldInfo,
   renderResponseData,
   renderDefaultFieldResponse,
+  getNestedIlosResponses,
 } from "./export";
 import { mockFormField, mockNestedFormField } from "utils/testing/setupJest";
 
@@ -52,6 +53,24 @@ describe("Test rendering methods", () => {
       "standard"
     );
     expect(result.props.children.id).toEqual("email-field-id");
+  });
+
+  test("Correctly renders nested ILOS fields", () => {
+    const mockFieldResponseData = [
+      {
+        key: "123",
+        value: "mock-ilos",
+      },
+    ];
+    const mockPlan = {
+      id: "mock-id",
+      plan_ilosUtilizationByPlan: [...mockFieldResponseData],
+      plan_ilosUtilizationByPlan_123: "N/A",
+    };
+
+    const result = getNestedIlosResponses(mockFieldResponseData, mockPlan);
+    expect(result[0].key).toEqual("mock-ilos");
+    expect(result[0].value).toEqual("N/A");
   });
 });
 
--- a/services/ui-src/src/utils/other/export.test.tsx
+++ b/services/ui-src/src/utils/other/export.test.tsx
@@ -1,3 +1,4 @@
+import { ReactJSXElement } from "@emotion/react/types/jsx-namespace";
 // types
 import { FormField, AnyObject } from "types";
 // utils
@@ -6,6 +7,7 @@ import {
   renderResponseData,
   renderDefaultFieldResponse,
   getNestedIlosResponses,
+  renderDrawerDataCell,
 } from "./export";
 import { mockFormField, mockNestedFormField } from "utils/testing/setupJest";
 
@@ -53,6 +55,38 @@ describe("Test rendering methods", () => {
       "standard"
     );
     expect(result.props.children.id).toEqual("email-field-id");
+  });
+
+  test("renderDrawerDataCell renders ilos responses", () => {
+    const mockFormField: FormField = {
+      id: "plan_ilosUtilizationByPlan",
+      props: {
+        choices: [],
+      },
+      type: "checkbox",
+      validation: "checkbox",
+    };
+
+    const mockPlan: AnyObject = [
+      {
+        id: "mock-id",
+        plan_ilosUtilizationByPlan: [
+          {
+            key: "123",
+            value: "mock-ilos",
+          },
+        ],
+        plan_ilosUtilizationByPlan_123: "N/A",
+      },
+    ];
+
+    const cells = renderDrawerDataCell(mockFormField, mockPlan, "drawer");
+    const ul = cells[0].props.children;
+    const li = Array.from(ul.props.children)
+      .flat()
+      .filter((el) => (el as ReactJSXElement)?.type === "li");
+    expect(cells.length).toBe(1);
+    expect(li.length).toBe(3);
   });
 
   test("Correctly renders nested ILOS fields", () => {
--- a/services/ui-src/src/utils/other/export.test.tsx
+++ b/services/ui-src/src/utils/other/export.test.tsx
@@ -1,4 +1,4 @@
-import { ReactJSXElement } from "@emotion/react/types/jsx-namespace";
+import { render, screen } from "@testing-library/react";
 // types
 import { FormField, AnyObject } from "types";
 // utils
--- a/services/ui-src/src/utils/other/export.test.tsx
+++ b/services/ui-src/src/utils/other/export.test.tsx
@@ -81,12 +81,10 @@ describe("Test rendering methods", () => {
     ];
 
     const cells = renderDrawerDataCell(mockFormField, mockPlan, "drawer");
-    const ul = cells[0].props.children;
-    const li = Array.from(ul.props.children)
-      .flat()
-      .filter((el) => (el as ReactJSXElement)?.type === "li");
-    expect(cells.length).toBe(1);
-    expect(li.length).toBe(3);
+    const Component = () => cells[0];
+    const { container } = render(<Component />);
+    expect(container.querySelectorAll("li").length).toBe(3);
+    expect(container.querySelectorAll("li")[2]).toHaveTextContent("mock-ilos: N/A");
   });
 
   test("Correctly renders nested ILOS fields", () => {
--- a/services/ui-src/src/utils/other/export.test.tsx
+++ b/services/ui-src/src/utils/other/export.test.tsx
@@ -84,7 +84,9 @@ describe("Test rendering methods", () => {
     const Component = () => cells[0];
     const { container } = render(<Component />);
     expect(container.querySelectorAll("li").length).toBe(3);
-    expect(container.querySelectorAll("li")[2]).toHaveTextContent("mock-ilos: N/A");
+    expect(container.querySelectorAll("li")[2]).toHaveTextContent(
+      "mock-ilos: N/A"
+    );
   });
 
   test("renderDrawerDataCell renders without ilos responses", () => {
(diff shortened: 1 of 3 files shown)


services/ui-src/src/utils/other/export.tsx
```diff
@@ -93,6 +93,13 @@ export const renderDrawerDataCell = (
       parentFieldCheckedChoiceIds &&
       !parentFieldCheckedChoiceIds?.includes(entity.id);
     const fieldResponseData = entity[formField.id];
+
+    // check for nested ILOS data
+    let nestedIlosResponse;
+    if (formField.id === "plan_ilosUtilizationByPlan") {
```
bangbay-bluetiger, line 99: Make `plan_ilosUtilizationByPlan` a constant?

services/ui-src/src/utils/other/export.tsx
```diff
@@ -93,6 +93,13 @@ export const renderDrawerDataCell = (
...
+      nestedIlosResponse = getNestedIlosResponses(fieldResponseData, entity);
```
bangbay-bluetiger, line 100: Add unit test for this?

services/ui-src/src/utils/other/export.tsx
```diff
@@ -261,6 +271,22 @@ export const getEntityDetailsMLR = (entity: EntityShape) => {
   };
 };
 
+const getNestedIlosResponses = (
+  fieldResponseData: AnyObject,
+  entity: EntityShape
+) => {
+  if (fieldResponseData) {
```
bangbay-bluetiger, line 278: `fieldResponseData` is always defined. Is this conditional needed?

services/ui-src/src/utils/other/export.tsx
karla-vm: just added one, thanks!

services/ui-src/src/utils/other/export.tsx
karla-vm: If the user doesn't "Yes" for ILOS, the `fieldResponseData` could come back undefined because that nested field is never generated, so I just shifted it around to be a check prior to entering this function.

britt-mo (commented):
Is it not required to fill out each field? Like I was able to successfully save without filling out the second field and in the PDF it remained blank

Screenshot 2024-07-11 at 8 43 56 AM

services/ui-src/src/utils/other/export.tsx
```diff
@@ -109,6 +119,17 @@ export const renderDrawerDataCell = (
               index
             )}
           </li>
+          {/* If there are nested ILOS responses available, render them here */}
+          {nestedIlosResponses &&
+            nestedIlosResponses.map((response: AnyObject) => {
+              return (
+                <li>
```
bangbay-bluetiger, line 126: `<li>` needs a `key` so it won't throw a warning.

services/ui-src/src/utils/other/export.tsx
```diff
@@ -93,6 +93,16 @@ export const renderDrawerDataCell = (
...
       !parentFieldCheckedChoiceIds?.includes(entity.id);
     const fieldResponseData = entity[formField.id];
+
+    // check for nested ILOS data
+    let nestedIlosResponses;
+    if (
+      fieldResponseData?.length &&
+      formField.id === "plan_ilosUtilizationByPlan"
+    ) {
+      nestedIlosResponses = getNestedIlosResponses(fieldResponseData, entity);
```
bangbay-bluetiger, line 103: `renderDrawerDataCell` doesn't have its own tests and it does seem tricky to test. Here's a suggestion to test this scenario:
```
test("renderDrawerDataCell renders ilos responses", () => {
  const mockForm: FormJson = {
    id: "plan_ilosUtilizationByPlan",
    fields: [
      {
        id: "mock-field-id",
        props: {
          choices: [
            {
              id: "mock-choice-id-1",
              label: "mock label 1",
            },
            {
              id: "mock-choice-id-2",
              label: "mock label 2",
            },
          ],
        },
        type: "radio",
        validation: "radio",
      },
    ],
  };

  const mockPlan: AnyObject = [
    {
      id: "mock-id",
      plan_ilosUtilizationByPlan: [
        {
          key: "123",
          value: "mock-ilos",
        },
      ],
      plan_ilosUtilizationByPlan_123: "N/A",
    }      
  ]

  const cells = renderDrawerDataCell(mockForm, mockPlan, "drawer");
  const ul = cells[0].props.children;
  const li = ul.props.children
  expect(cells.length).toBe(1);
  expect(li.length).toBe(3);
});
```

bangbay-bluetiger: `expect(li.length).toBe(3);` is actually giving a false positive.

services/ui-src/src/utils/other/export.tsx
```diff
@@ -93,6 +93,16 @@ export const renderDrawerDataCell = (
       parentFieldCheckedChoiceIds &&
       !parentFieldCheckedChoiceIds?.includes(entity.id);
     const fieldResponseData = entity[formField.id];
+
+    // check for nested ILOS data
+    let nestedIlosResponses;
```
bangbay-bluetiger, line 98: Could do `let nestedIlosResponses = []`;

services/ui-src/src/utils/other/export.tsx
```diff
@@ -109,6 +119,17 @@ export const renderDrawerDataCell = (
               index
             )}
           </li>
+          {/* If there are nested ILOS responses available, render them here */}
+          {nestedIlosResponses &&
```
bangbay-bluetiger, line 123: If you have `let nestedIlosResponses = [];`, you can remove `nestedIlosResponses &&`.

karla-vm:
> Is it not required to fill out each field? Like I was able to successfully save without filling out the second field and in the PDF it displayed `hghg: `
> 
> Screenshot 2024-07-11 at 8 43 56 AM

that shouldn't be happening, I'll take a look! it's supposed to only display checked options.

karla-vm:
@britt-mo I'm not able to reproduce that issue you're describing. what steps did you take?

services/ui-src/src/utils/other/export.tsx
karla-vm: I think this is out of the scope for this one, but I'll create a ticket for increasing test coverage and make a note of this example! thank you, @bangbay-bluetiger 🥇 great call.

karla-vm: @bangbay-bluetiger that's super helpful, I included this unit test in the PR. thank you!!

bangbay-bluetiger (dismissed):

britt-mo (dismissed):

services/ui-src/src/utils/other/export.test.tsx
```diff
@@ -53,6 +56,56 @@ describe("Test rendering methods", () => {
...
+      },
+    ];
+
+    const cells = renderDrawerDataCell(mockFormField, mockPlan, "drawer");
+    const ul = cells[0].props.children;
+    const li = Array.from(ul.props.children)
+      .flat()
+      .filter((el) => (el as ReactJSXElement)?.type === "li");
+    expect(cells.length).toBe(1);
+    expect(li.length).toBe(3);
```
bangbay-bluetiger, line 89: After thinking about it, it seemed there should be a way to use `render` here and I think this is cleaner:

```suggestion
    const cells = renderDrawerDataCell(mockFormField, mockPlan, "drawer");
    const Component = () => cells[0];
    const { container } = render(<Component />);
    expect(container.querySelectorAll("li").length).toBe(3);
    expect(container.querySelectorAll("li")[2]).toHaveTextContent("mock-ilos: N/A");
```

services/ui-src/src/utils/other/export.test.tsx
```diff
@@ -53,6 +56,56 @@ describe("Test rendering methods", () => {
...
+  });
+
```
bangbay-bluetiger, line 91: Adding the opposite scenario would be good too:

```suggestion

    test("renderDrawerDataCell renders without ilos responses", () => {
    const mockFormField: FormField = {
      id: "mock",
      props: {
        choices: [],
      },
      type: "checkbox",
      validation: "checkbox",
    };

    const mockPlan: AnyObject = [
      {
        id: "mock-id",
        mock: [
          {
            key: "123",
            value: "mock-ilos",
          },
        ],
        mock_123: "N/A",
      },
    ];

    const cells = renderDrawerDataCell(mockFormField, mockPlan, "drawer");
    const Component = () => cells[0];
    const { container } = render(<Component />);
    expect(container.querySelectorAll("li").length).toBe(2);
    expect(screen.queryByText("mock-ilos: N/A")).not.toBeInTheDocument();

```

services/ui-src/src/utils/other/export.test.tsx
```diff
@@ -1,10 +1,13 @@
+import { ReactJSXElement } from "@emotion/react/types/jsx-namespace";
```
bangbay-bluetiger, line 1:
```suggestion
import { render, screen } from "@testing-library/react";
```

ntsummers1 (dismissed):
What happens when a user creates an ILOS, and then deletes the parent plan? Currently the data still stays despite there being no plan to associate it with. Is that how we should have it? If we delete a plan, should it delete all the ilos information associated with it?

karla-vm:
> What happens when a user creates an ILOS after, and then deletes the parent plan? Currently the data still stays despite there being no plan to associate it with. Is that how we should have it? If we delete a plan, should it delete all the ilos information associated with it?

that's a great callout. I'll make a ticket for this and address it in a separate PR!

ntsummers1 (approved):
I think I was a fool. I'm sorry. Everything's looking great now. Here 👑, take this and my ✅

bangbay-bluetiger (approved):

PR #11804, opened 2024-07-18
Cmdct 3802 - PDF banner button should not be displayed if the PDF has an error, a spinner if the PDF is not loaded

britt-mo:
### Description
The user always sees an active button to download the PDF version of the report, regardless of if the PDF has been generated. This PR addresses this issue by adding a spinner if the report has not generated, or does not display if there is an error message.

### Related ticket(s)

CMDCT-3802

---
### How to test
Create a program, go to the review and submit, go to pdf, see that the button is a spinner if the PDF isn't generated -- you can set the check on line 38 in `<ExportedReportBanner />` to `false` . To see the loader. Also, Garrett was able to see the spinner after  deleting the mcr-store in local storage.

In `<ExportedSectionHeading />`, remove lines 35 - 41, leaving only `{sectionInfo}` and see that the pdf banner does not display at all.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/app-api/forms/mcpar.json
+++ b/services/app-api/forms/mcpar.json
@@ -1075,7 +1075,11 @@
                     },
                     {
                       "type": "p",
-                      "content": "Revisions to the Medicaid managed care regulations in 2016 and 2020 built on existing requirements that managed care plans maintain provider networks sufficient to ensure adequate access to covered services by: (1) requiring states to develop quantitative network adequacy standards for at least eight specified provider types if covered under the contract, and to make these standards available online; (2) strengthening network adequacy monitoring requirements; and (3) addressing the needs of people with long-term care service needs (42 CFR 438.66; 42 CFR 438.68).</p><p>42 CFR 438.66(e) specifies that the MCPAR must provide information on and an assessment of the availability and accessibility of covered services within the MCO, PHIP, or PAHP contracts, including network adequacy standards for each managed care program."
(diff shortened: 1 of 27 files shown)


gmrabian (commented):
I think having a spinner when the page has an error might be confusing. To me it makes it seem like something might be happening, or it will change eventually to a non-error state.

Wasn't able to see the spinner for normal pdf load because it seems to render the header alongside the report

britt-mo:
@gmrabian I can ask design how they feel about it

gmrabian (commented):
I was able to get the spinner to show briefly by deleting the mcr-store in local storage

The error state looks good with not header.
Is the footer supposed to be floating in the middle there?
Screenshot 2024-07-22 at 9 13 04 AM

karla-vm:
> I was able to get the spinner to show briefly by deleting the mcr-store in local storage
> 
> The error state looks good with not header. Is the footer supposed to be floating in the middle there? Screenshot 2024-07-22 at 9 13 04 AM

I don't think it's supposed to be floating up there, but I'm not really sure 🤔

karla-vm (commented):
LGTM, although we may want to look at the footer display issue @gmrabian pointed out.

britt-mo:
@gmrabian undid the changes in <App /> that had messed up the container styling. New logic in <ExportedReportBanner /> to determine displaying or not.

karla-vm (approved):

gmrabian (approved):

PR #11881, opened 2024-10-21
Refactor amplify methods to single file

bangbay-bluetiger:
### Description

[Amplify](https://github.com/aws-amplify/amplify-js) typically has breaking changes with each version upgrade. To mitigate how much refactoring we have to do with each upgrade, we should use generic helpers and call amplify methods from a single file, as much as possible.

### Related ticket(s)

CMDCT-4037

---
### Notes

---
### Pre-review checklist


karla-vm (approved):

gmrabian (commented):
I did a quick scan of it all and I like this! I appreciate you breaking out all the amplify stuff, whereas in some apps we only have the apis broken out

britt-mo (approved):
Everything worked as expected when running locally

gmrabian (approved):

PR #11932, opened 2024-11-18
Revert "Fix signout maybe"

jessabean:
Reverts Enterprise-CMCS/macpro-mdct-mcr#11931


gmrabian (commented):
deploy failing on branch name length.

since this is a revert it's probably safe to override and merge anyway...but up to you 🤷

PR #11944, opened 2024-11-25
add border to footer on report pages, remove bottom border on last entity row

britt-mo:
### Description
After discussing with Kim, the MCPAR report pages are following MFP styling -- there is a footer border line before the navigation buttons at the bottom of every report page, and on drawer pages the last entity row  does not have a bottom border.

Also -- the update to the form e2e (commented out code block) is a very temporary solution until we have time to agree on what solution we want to use for tracking flag statuses in Cypress e2e tests since we are moving onto Playwright.

### Related ticket(s)

CMDCT-4161

---
### How to test
Create a MCPAR program, see that the pages all have a bottom border line that matches the spec, and see that the last row on a drawer page does not have a bottom border.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/reports/DrawerReportPage.test.tsx
+++ b/services/ui-src/src/components/reports/DrawerReportPage.test.tsx
@@ -298,6 +298,12 @@ describe("Test DrawerReportPage with completed entity", () => {
     expect(screen.queryAllByText("Enter")).toHaveLength(1);
     expect(screen.getAllByAltText("Entity is complete")).toHaveLength(1);
   });
+
+  test("should not render a bottom border on last entity row", () => {
+    const entityRows = screen.getAllByTestId("report-drawer");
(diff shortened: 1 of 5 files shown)


jessabean:
Does the Figma specify that the footer border shouldn't be on drawer page types, or is it more that there should not be a border when the content includes a table that has a bottom border? I kind of want to take this back to design and see if we can come up with a better rule so we don't need to have exceptions like this. Maybe we can force a larger amount of padding at the bottom of report page content, so there is obvious space between content and the footer, or something.

britt-mo:
@jessabean I've reached out to Kim about the border rule

karla-vm (approved):

bangbay-bluetiger (approved):

JonHolman (approved):

PR #11962, opened 2024-12-07
Use intercept to check for Launch Darkly flag

bangbay-bluetiger:
### Description

- Add a `flag` attribute to route
- If it doesn't match a live Launch Darkly flag, the route is skipped in Cypress form test

### Related ticket(s)

---
### How to test

1. `yarn test` should pass

---
### Pre-review checklist

### DIFF
--- a/services/app-api/forms/mcpar.json
+++ b/services/app-api/forms/mcpar.json
@@ -637,6 +637,7 @@
         },
         {
           "name": "XIII: Prior Authorization",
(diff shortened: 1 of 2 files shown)


britt-mo (approved):
Tests passed. Great solution for the flag checks. It's probably worth writing a spike card to look into how we can do the flag check with Launch Darkly SDK.

karla-vm (approved):

PR #11979, opened 2025-01-02
fix some validation that was removing fields, and only add applicable fields

gmrabian:
### Description

- Analysis methods gets stripped from our fieldData unless we add it to the validation schema
  - Check if report is NAAAR and then add to validation
  - Note: still not positive the drawer information will save correctly with this but we can figure it out later
- Move `versionControl` to mlr-only method, since it's only for that report type
- Remove `submissionCount` from fieldData
  - It exists in metadata and is never referenced from fieldData (as far as I could tell)
  - It is stripped from fieldData on create anyway because it's not in the validation schema
- Remove `programName` from MLR and NAAAR
  - It is only used in mcpar and likely just got copied over by accident
  - It is stripped from fieldData on create anyway because it's not the in validation schema (other than mcpar)

### Related ticket(s)

CMDCT-4115 (partial)

---
### How to test

[deployed url](https://d3uzmu4pvkgkrr.cloudfront.net/)
- create a NAAAR and verify that analysis methods exists in fieldData when you enter the report (verify with react plugin or by going to analysis methods page)
- creating and editing reports works as before
- do a double check that removed fields are not used anywhere (or at least not referenced where they were created, like in fieldData)

### Notes

i forgot how to write nice pr descriptions

---
### Pre-review checklist

### DIFF
--- a/services/app-api/handlers/reports/create.test.ts
+++ b/services/app-api/handlers/reports/create.test.ts
@@ -65,13 +65,25 @@ const creationEvent: APIGatewayProxyEvent = {
   }),
 };
 
-const naaarCreationEvent: APIGatewayProxyEvent = {
+const naaarCreationDisabledEvent: APIGatewayProxyEvent = {
   ...proxyEvent,
   headers: { "cognito-identity-id": "test" },
   pathParameters: { reportType: "NAAAR", state: "AL" },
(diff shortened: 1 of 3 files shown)


bangbay-bluetiger (approved):
LGTM

PR #11994, opened 2025-01-06
Add additional Analysis Methods

gmrabian:
### Related ticket(s)

CMDCT-4115

---
### Notes

---
### Pre-review checklist

---
#### Review
---


services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -50,6 +54,8 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
   const { report, selectedEntity, setSelectedEntity } = useStore();
 
   const { entityType, verbiage, drawerForm, form: standardForm } = route;
+  const { addEntityDrawerForm } = route ?? {};
```
bangbay-bluetiger, line 57: Seems like this is causing skipping a few type checks later on. How about doing something like this and avoid using `!`?
```
const addEntityDrawerForm = route.addEntityDrawerForm || {} as FormJson;
const canAddEntities = !!addEntityDrawerForm.id;
```

services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -254,14 +305,29 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
...
       <ReportDrawer
         selectedEntity={selectedEntity!}
         verbiage={{
           drawerTitle: `${verbiage.drawerTitle} ${selectedEntity?.name}`,
           drawerInfo: verbiage.drawerInfo,
         }}
-        form={form}
+        form={
+          isAnalysisMethodsPage &&
+          !getDefaultAnalysisMethodIds().includes(selectedEntity?.id!)
```
bangbay-bluetiger, line 327: Instead of needing to coalesce for `selectedEntity` throughout this file, seems like we could do something at the top like:
```
const selectedEntity = useStore().selectedEntity ?? {};
```

JonHolman:
How about extracting entity row into its own component?

services/ui-src/src/components/reports/DrawerReportPage.tsx
gmrabian: I think this still runs into the issue of `.id` not being defined if it falls back on `{}`.
I added a useEffect so it only has to do the check once. What do you think?

gmrabian:
> How about extracting entity row into its own component?

Funny you mention it...we have one of those! I can take a look and see how well it would fit together.
https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/ui-src/src/components/tables/EntityRow.tsx

services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -254,14 +310,24 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
...
+            onClick={() => openRowDrawer()}
+          >
+            {verbiage.addEntityButtonText}
+          </Button>
+        )}
       </Box>
       <ReportDrawer
         selectedEntity={selectedEntity!}
         verbiage={{
           drawerTitle: `${verbiage.drawerTitle} ${selectedEntity?.name}`,
```
bangbay-bluetiger, line 327: When adding a new analysis method, `selectedEntity?.name` prints out undefined in the title:

Screenshot 2025-01-07 at 12 38 32 PM

Could do something like `${selectedEntity?.name || "add other"}` or check with Kim for verbiage.

services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -254,14 +310,24 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
...
+            sx={sx.bottomAddEntityButton}
+            leftIcon={<Image sx={sx.buttonIcons} src={addIcon} alt="Add" />}
+            onClick={() => openRowDrawer()}
+          >
+            {verbiage.addEntityButtonText}
+          </Button>
+        )}
       </Box>
       <ReportDrawer
         selectedEntity={selectedEntity!}
```
bangbay-bluetiger, line 325: Nit: Would love to get rid of more of these `!` operators...

gmrabian: I hear you. For now I'm only focusing on the changes related to this feature, but I do think this component could use some reworking separately.

services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -186,30 +223,63 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
...
+              <Heading as="h4" sx={sx.customEntityName}>
+                {entity.custom_analysis_method_name}
+              </Heading>
+              {entity.custom_analysis_method_description && (
+                <Text>{entity.custom_analysis_method_description}</Text>
+              )}
+              {entity.analysis_method_frequency &&
+                entity.analysis_method_applicable_plans && (
+                  <Text>
+                    {entity.analysis_method_frequency[0].value}:
```
bangbay-bluetiger, line 257: Nit of nits: Might want to add a `&nbsp;` or other kind of separator after `:`

Screenshot 2025-01-08 at 12 32 25 PM

bangbay-bluetiger (approved):

services/ui-src/src/components/reports/DrawerReportPage.tsx
gmrabian: Good callout! I'm getting a design review now so I'll include this change with any other suggestions they may have

services/app-api/forms/naaar.json
```diff
@@ -284,7 +284,7 @@
                 }
               ]
             },
-            "dashboardTitle": "Analysis method <br/> Frequency and plan utilization",
+            "dashboardTitle": "<div style=\"margin-left: 3rem;\">Analysis method <br/> Frequency and plan utilization</div>",
```
gmrabian, line 287: I did try using the json structured styling that we have elsewhere ([for example](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/app-api/forms/mlr.json#L319-L324)) but it seemed to ignore the sx props for the dashboard title in the component

gmrabian: nvm decided I didn't like this

bangbay-bluetiger (approved):

karla-vm (dismissed):
This looks great and works well, I don't have any major feedback on the code. I did create a separate ticket for us to refactor the `DrawerReportPage` in the future, so this might be a good case for it --- separating out this type of page as its own thing as you mentioned.

karla-vm (approved):

bangbay-bluetiger (approved):

PR #12008, opened 2025-01-13
Add ability to delete custom analysis methods

gmrabian:
### Description

- Add analysis methods page to completion calculation
  - Determines if method is part of default, and then applies calculation with the applicable form
    - The `calculateEntityCompletion` method accepts an array of different forms, but it expects the data to fulfill every form, rather than being able to apply a specific form per dataset, so this was the best way I could see to do it
- Add logic to disable "submit" button on delete modal for non-end users and when form is locked (submitted)
- Add delete entity button and modal to `DrawerReportPage`

### Related ticket(s)

CMDCT-4233

---
### How to test

[Deployed url](https://d31vhlh58sn663.cloudfront.net/)
- Log in as a state user
- Create and enter a NAAAR
- Create a plan
- Go to Analysis Methods
- Add a custom analysis method
- Click the X button on the new method
  - Verify it opens the delete modal
  - Verify the text is relevant to analysis methods
  - Verify when you click the submit button in the modal it deletes that method
- Verify analysis methods completion statusing on the review and submit page when you have completed all methods, including adding custom methods

Navigate to a different `drawer` type page (in MCPAR you can create a plan and go to plan level indicators)
- Verify the delete button does not show up in those rows

### Notes

---
### Pre-review checklist

---
#### Review
---


JonHolman:
When I add multiple, then I hit the delete, from the modal I cannot confirm which one I hit delete on.

JonHolman:
image
The columns are not lining up when I add a bunch with some different frequencies and descriptions.

services/app-api/utils/validation/completionStatus.ts
```diff
@@ -210,6 +211,36 @@ export const calculateCompletionStatus = async (
             route.entityType
           ),
         };
+        // handle Analysis Methods case: this section allows users to add custom methods, which use different form questions
+        if (
+          route.path ===
+            "/naaar/state-and-program-information/analysis-methods" &&
```
britt-mo, line 217: maybe make this url a constant and refer to it here?

karla-vm:
This looks great! Statusing is working as expected, I was able to add and delete custom methods. One thing I noticed is that when you enter a custom method, the `Plans utilizing this method` question is not validating correctly, I'm able to save without clicking on a plan checkbox. 

In terms of the styling, I played around a bit with adding an empty `<Heading />` or a hidden `<Button />` and it feels hacky... Your idea of converting this into a table sounds like the best approach for sure! I'll create a follow-up ticket for that refactoring work.

gmrabian:
> When I add multiple, then I hit the delete, from the modal I cannot confirm which one I hit delete on.

@JonHolman this is a good callout! The Figma also does not specify. Perhaps in the future we could add that as an enhancement but for now I am not going to add it to this work.

services/app-api/utils/validation/completionStatus.ts
gmrabian: @britt-mo the other routes in this switch statement are hardcoded so I will leave it for now, but good callout!

services/app-api/utils/validation/completionStatus.ts
```diff
@@ -210,6 +211,36 @@ export const calculateCompletionStatus = async (
...
+        ) {
+          let areAllFormsComplete = true;
+          for (const entity of fieldData[route.entityType]) {
+            // get completion status for entity, using the correct form template
+            const applicableForm = DEFAULT_ANALYSIS_METHODS.includes(
+              entity.name
+            )
+              ? route.drawerForm
+              : route.addEntityDrawerForm;
+            const isEntityComplete = await calculateFormCompletion(
```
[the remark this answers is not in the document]
gmrabian: thank you! I realized I did the validation wrong for those fields so I made the changes in the naaar.json. if you create a new naaar you will see it working! 🎉 
cc: @karla-vm

services/app-api/utils/validation/completionStatus.ts
```diff
@@ -210,6 +211,36 @@ export const calculateCompletionStatus = async (
...
+              applicableForm,
+              entity
+            );
+            // update combined result
+            areAllFormsComplete &&= isEntityComplete;
```
bangbay-bluetiger, line 233: Little confused why the results need to be combined? Only one `false` is needed to change the value of `areAllFormsComplete`, so we could exit the loop early if we get a `false`.

```
areAllFormsComplete = isEntityComplete;
if (!areAllFormsComplete) break;
```
We could also eliminate one of the variables:

```
areAllFormsComplete = await calculateFormCompletion(...);
if (!areAllFormsComplete) break;
```

services/ui-src/src/components/reports/DrawerReportPage.test.tsx
```diff
@@ -384,44 +384,65 @@ describe("Test DrawerReportPage with custom entities", () => {
...
-      reportsByState: [mockNaaarReportContextWithCustomAnalysisMethods.report],
-    };
-
-    mockedUseStore.mockReturnValue({
-      ...mockStateUserStore,
-      ...mockCustomNaaarReportStore,
-      ...mockAnalysisMethodEntityStore,
+  describe("test analysis methods custom logic", () => {
+    beforeEach(() => {
+      const mockNaaarReportContextWithCustomAnalysisMethods: any =
```
bangbay-bluetiger, line 389: Say `mockNaaarReportContextWithCustomAnalysisMethods` quickly three times.  =)

services/ui-src/src/components/reports/DrawerReportPage.test.tsx
```diff
@@ -384,44 +384,65 @@ describe("Test DrawerReportPage with custom entities", () => {
...
+        mockNaaarReportContext;
+      mockNaaarReportContextWithCustomAnalysisMethods.report.fieldData[
```
bangbay-bluetiger, line 391: I'd add a blank new line here because it was hard to tell this was a new method and not something continuously wrapped from the line before. Also, could destructure out `report` since it's used a few places:

```
const { report } = mockNaaarReportContextWithCustomAnalysisMethods;

report.fieldData["analysisMethods"]  = etc.
```

services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -299,7 +318,18 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
...
-          {enterButton(entity, isEntityCompleted)}
+          <Box sx={buttonBoxStyling(canAddEntities)}>
+            {enterButton(entity, isEntityCompleted)}
+            {canAddEntities && !entity.isRequired && (
+              <Button
+                sx={sx.deleteButton}
+                data-testid="delete-entity"
+                onClick={() => openDeleteEntityModal(entity)}
+              >
+                <Image src={deleteIcon} alt="delete icon" boxSize="2xl" />
```
bangbay-bluetiger, line 329: Since there's no text in this button, wonder if the alt tag should reflect what the button action would be, i.e. `Delete` vs. describing what the image looks like, i.e. `delete icon`. Question for Cypress, perhaps.

services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -424,6 +463,15 @@ function dashboardTitleStyling(canAddEntities: boolean) {
   };
 }
 
+function buttonBoxStyling(canAddEntities: boolean) {
+  return {
+    marginRight: canAddEntities && "1.5rem",
+    "button + button": {
```
bangbay-bluetiger, line 469: Alternatively, putting `mr={-10}` on the delete `Button` or putting `marginRight: "-2.5rem"` with the `deleteButton`  CSS definitions below seem to give the same styling effect.

bangbay-bluetiger (approved):
Looks good to me! Comments are really nits.

gmrabian:
thanks @bangbay-bluetiger! great feedback as always

ntsummers1 (approved):

karla-vm (approved):

PR #12016, opened 2025-01-22
NAAAR — remove deleted plans from analysis methods

gmrabian:
### Description

- Ensure deleted plans are removed from their analysis methods
- Add tests
  - Refactor some reusable mocks

Note: this successfully removes the plans, but does not always correctly update statusing (try deleting all plans and see if statusing changes). Will create a separate ticket for that.

### Related ticket(s)

CMDCT-4250

---
### How to test

- Create a NAAAR report
- Create two or more plans
- Fill out at least one default analysis method and select all plans
- Fill out at least one custom analysis method and select all plans
- Delete one of the plans on the plans page
- Check the analysis methods to ensure they no longer have those plans associated (check frontend and saved data to verify)
  - If verifying locally, the Redux devtools extension is really helpful for checking this data change

---
### Pre-review checklist

---
#### Review
---


services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -141,6 +141,19 @@ export const DynamicField = ({ name, label, ...props }: Props) => {
...
+        (method: EntityShape) => {
+          if (method.analysis_method_applicable_plans?.length) {
+            method.analysis_method_applicable_plans =
+              method.analysis_method_applicable_plans.filter(
+                (plan: AnyObject) => plan.key !== selectedRecord.id
+              );
+          }
+          return method;
+        }
+      );
```
gmrabian, line 155: Let me know if you can think of a cleaner way to make this edit! I got this from combining patterns above here in the file. I think it's tricky because there is double nesting of arrays and objects to get to what we need to edit. I played around with a for...of but still needed some kind of map/filter loop to do the changes.

services/ui-src/src/components/fields/DynamicField.test.tsx
```diff
@@ -327,6 +332,68 @@ describe("<DynamicField />", () => {
...
+        report: mockNaaarReportWithAnalysisMethodsContext.report,
+      };
+
+      await act(async () => {
+        await render(
+          dynamicFieldComponent(mockHydrationPlans, mockedReportContext)
+        );
+      });
+      // delete mock-plan-1
+      const removeButton = screen.queryAllByTestId("removeButton")[0];
```
bangbay-bluetiger, line 354: Looks like the button is accessible by the alt text `Delete mock-ilos-1`, so you could use `screen.getByRole("button", { name: "Delete mock-ilos-1" })` and remove the `data-testid="removeButton"` attribute on the button.

services/ui-src/src/components/fields/DynamicField.test.tsx
```diff
@@ -327,6 +332,68 @@ describe("<DynamicField />", () => {
...
+      await userEvent.click(removeButton);
+      const deleteButton = screen.getByText("Yes, delete plan");
```
bangbay-bluetiger, line 356: I would use `screen.getByRole("button", { name: "Yes, delete ILOS" })` to get the button.

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -141,6 +141,19 @@ export const DynamicField = ({ name, label, ...props }: Props) => {
...
 
+      // filter analysis methods to remove deleted plans
+      const filteredAnalysisMethods = report?.fieldData?.analysisMethods?.map(
+        (method: EntityShape) => {
+          if (method.analysis_method_applicable_plans?.length) {
+            method.analysis_method_applicable_plans =
+              method.analysis_method_applicable_plans.filter(
+                (plan: AnyObject) => plan.key !== selectedRecord.id
+              );
+          }
```
bangbay-bluetiger, line 152: You can shorten this slightly to:
```
method.analysis_method_applicable_plans =
  method.analysis_method_applicable_plans?.filter(
    (plan: AnyObject) => plan.key !== selectedRecord.id
  );
```

britt-mo (dismissed):
Looked correct in the UI and correct updated data in devtools

bangbay-bluetiger (dismissed):
Worked for me. Minor nits.

services/ui-src/src/components/reports/DrawerReportPage.test.tsx
```diff
@@ -384,67 +357,28 @@ describe("<DrawerReportPage />", () => {
...
-
-        // verify that the field is removed
-        const inputBoxLabelAfterRemove = screen.queryAllByTestId("test-label");
-        expect(inputBoxLabelAfterRemove).toHaveLength(0);
-        expect(addCustomMethod).toBeVisible();
-      });
+    test("DrawerReportPage opens the delete modal on remove click", async () => {
+      render(drawerReportPageWithCustomEntities);
+      const addCustomMethod = screen.getByText("Add other analysis method");
+      const removeButton = screen.getByTestId("delete-entity");
```
bangbay-bluetiger, line 369: This is another button where you could remove `data-testid` and get the button with the alt text. In this case, I'd change the alt text to something more meaningful like `` alt={`Delete ${entity.custom_analysis_method_name}`}``.

services/ui-src/src/components/fields/DynamicField.test.tsx
gmrabian: I made the change for this test, but I didn't want to rewrite all the tests for this file in this PR (some of the Delete buttons don't have identifiers because the tests aren't typing anything, so I just wanted to leave it alone for now).

services/ui-src/src/components/fields/DynamicField.tsx
gmrabian: I got this error. I think We need to ensure it as an array. Even with the optional symbol `?` it still seemed to cause an issue (throws error in console and hangs)
`Uncaught (in promise) TypeError: method.analysis_method_applicable_plans.filter is not a function`

gmrabian:
@bangbay-bluetiger was able to implement some of the feedback. See notes! Thanks

bangbay-bluetiger (approved):

jessabean (approved):

services/ui-src/src/components/fields/DynamicField.tsx
karla-vm: I think this could be an opportunity for a refactor in the future, but it looks good to me for now 💯

PR #12018, opened 2025-01-22
NAAAR analysis methods statusing: more custom logic!

gmrabian:
### Description

- Ensure statusing for analysis methods updates correctly when you remove all applicable plans
- Tighten up display logic
  - Only show delete custom analysis method button if there are plans (meaning if they are also editable)
  - Only show custom analysis method frequency and plan text if there are plans in the plan array

Open to improvements :)

I ran into a few issues that got me here:
1. The way we check statusing is to ensure every top level field idea is present in the data. When I changed it to included nested ids, it included the "other, specify" text, which is only required if selected, but caused the methods to fail statusing since it was in the id array. It got complicated 😓
2. Statusing also doesn't run validation against the data, just ensures the id exists. So an empty array, empty string, etc. all counts as complete.
3. I would have considered implementing one of the above more thoroughly, but at the moment we only have the logic for digging up all nested fields and validating them against their field types in the backend. Thus, it would have added a lot of copied code to repeat the same.

The solution I landed on does a simple check against the only offending field: plans. This is the only one that can be edited outside of the drawer (by deleting a plan on the plans page).

### Related ticket(s)

CMDCT-4253

---
### How to test

- Create a NAAAR
- Add three plans
- Fill out multiple default and custom analysis methods, with various plans selected
- See what happens as you delete all plans associated with a certain method
  - Ensure statusing changes to incomplete if there are all applicable plans were deleted
  - Ensure you cannot delete a method if all plans were deleted
  - Ensure methods where you selected "No", meaning not applicable, that it stays complete after deleting plans
  - Ensure frequency and plan text only shows up for custom methods when a plan is applicable (goes away when deleted)

---
### Pre-review checklist

  - Please don't make me 😭  

---
#### Review
---


karla-vm (approved):
works like a charm! ❇️

services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -257,6 +257,18 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
       const isCustomEntity =
         canAddEntities && !getDefaultAnalysisMethodIds().includes(entity.id);
       const calculateEntityCompletion = () => {
+        // logic to ensure analysis methods always have a plan selected
```
bangbay-bluetiger, line 260: I'm a little confused about what `calculateEntityCompletion` is supposed to return. An array of field ids?

services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -257,6 +257,18 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
...
+        if (
+          isAnalysisMethodsPage &&
+          !entity?.analysis_method_applicable_plans?.length
+        ) {
+          if (isCustomEntity) {
+            return false;
```
bangbay-bluetiger, line 266: If it's normally returning an array, having it return `false` here is confusing. I would have an undefined return or an empty array. Consolidating all into one conditional could also make the code slightly more readable:
```
if (
  isAnalysisMethodsPage &&
  entity?.analysis_method_applicable_plans?.length < 1 &&
  (isCustomEntity || entity?.analysis_applicable?.[0]?.value === "Yes")
) {
  return;
}
```

services/ui-src/src/components/reports/DrawerReportPage.tsx
gmrabian: It returns a boolean. The final return call ends with [`.every()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every), so these shortcuts are just returning false early.

This function is called a few lines below its definition as such
```
const isEntityCompleted = reportingOnIlos
        ? calculateEntityCompletion() &&
          isIlosCompleted(reportingOnIlos, entity)
        : calculateEntityCompletion();
```

bangbay-bluetiger: Okay. I would vote for changing `calculateEntityCompletion()` to something more boolean-sounding though `isEntityCompleted` is already taken...

karla-vm (dismissed):

karla-vm (dismissed):

jessabean (approved):
👍🏻

karla-vm (approved):

services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -257,6 +257,15 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
...
         canAddEntities && !getDefaultAnalysisMethodIds().includes(entity.id);
       const calculateEntityCompletion = () => {
+        // logic to ensure analysis methods always have a plan selected
+        if (
+          isAnalysisMethodsPage &&
+          !entity?.analysis_method_applicable_plans?.length &&
+          (isCustomEntity || entity?.analysis_applicable?.[0]?.value === "Yes")
+        ) {
+          return false;
+        }
```
bangbay-bluetiger, line 267: Consolidated from
```
if (
  isAnalysisMethodsPage &&
  !entity?.analysis_method_applicable_plans?.length
) {
  if (isCustomEntity) {
    return false;
  } else if (entity?.analysis_applicable?.[0]?.value === "Yes") {
    return false;
  }
}
```

bangbay-bluetiger (approved):

services/ui-src/src/components/reports/DrawerReportPage.tsx
gmrabian: Thank you!

PR #12019, opened 2025-01-22
Add a longer wait to Cypress test

bangbay-bluetiger:
### Description

Follow up to https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/12015. This test passes locally and in branches, but has been failing when merged to `main`. Adding a longer wait to see if that helps.

Also update `todos` and `flags` logs from latest main updates.

---
### How to test

1. `yarn test` to test locally

### Notes

The failing test was changed from `cy.contains("Archive").should("be.visible");` to `cy.get('button:contains("Archive")').should("be.visible");` in https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/12009/ because the former test was a false positive, since the text is also used in a table header. We want to test that the text is within a button in a table row.

---
### Pre-review checklist

### DIFF
--- a/logs/launchdarkly-flags.log
+++ b/logs/launchdarkly-flags.log
@@ -2,21 +2,21 @@
 
 # LaunchDarkly flags in components
 ./services/ui-src/src/components/app/AppRoutes.tsx:30:useFlags()?.naaarReport;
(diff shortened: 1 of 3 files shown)


gmrabian (approved):
🥲

PR #12022, opened 2025-01-27
Use serverless version installed by package.json

bangbay-bluetiger:
### Description

Our apps are set up to use a global version of `serverless`. Since there can only be one version of a package installed globally, it's causing apps to fail if they're not on the same version. Particularly with `serverless`, there doesn't seem to be a reason to use a global version when the apps install a version of `serverless` locally. This PR:

- Updates `./run local` to use the local `serverless` binary
- Removes global installation of `serverless`

---
### How to test

1. Run `./run local`
2. App should boot normally

---
### Pre-review checklist


JonHolman:
We can do this, but I will be opening a PR very soon (maybe today, at least tomorrow morning to undo this)

JonHolman (approved):

PR #12036, opened 2025-02-03
[CMDCT-4259]: Plans Compliance Page

ntsummers1:
### Description

This page kicks off the third section of the NAAAR Form. It adds a new page type, as well as a quick fix for the mobile view of the entity rows. It also makes the plans created in Section 1 have the `isRequired` boolean. This page is very rough. Not a lot of styling, the error messages don't check for much and the statuses don't work. But! It does kick off the process and let's us begin hooking things up. 

### Related ticket(s)

CMDCT-4259

---
### How to test

Create a NAAR Form and navigate to the new Section 3: Plan Compliance!
See that theres 2 error messages currently about not filling things out
Go to Section 1: Add Plans and add a plan or 2
Go back to Section 3 and see that the plans you added are there with the first error message out of the way!
Check out the mobile view too and see it in all its glory
Click enter in either view and see the beginnings of the overlay for the next ticket :D 

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/app-api/forms/naaar.json
+++ b/services/app-api/forms/naaar.json
@@ -197,7 +197,8 @@
                 "type": "dynamic",
                 "validation": "dynamic",
                 "props": {
-                  "label": "Plan name"
+                  "label": "Plan name",
+                  "isRequired": true
                 }
               }
             ]
@@ -507,6 +508,7 @@
       "name": "III. Plan compliance",
       "path": "/naaar/plan-compliance",
       "pageType": "planOverlay",
+      "entityType": "plans",
       "verbiage": {
         "intro": {
           "section": "",
@@ -518,8 +520,44 @@
             }
           ]
         },
-        "tableHeader": "Plan name",
-        "enterPlanButtonText": "Enter"
+        "missingInformationMessage": [
+          {
+            "type": "p",
+            "children": [
+              {
+                "type": "html",
+                "content": "This program is missing required information. You won’t be able to access this section until you’ve completed all the required questions in section "
+              },
+              {
+                "type": "internalLink",
+                "content": "I. State and program information.",
+                "props": {
+                  "to": "/naaar/state-and-program-information/state-information-and-reporting-scenario"
+                }
+              }
+            ]
+          }
+        ],
+        "missingStandardsMessage": [
+          {
+            "type": "p",
+            "children": [
+              {
+                "type": "html",
+                "content": "This program is missing standards. You won’t be able to access this section until you’ve added standards in section "
+              },
+              {
+                "type": "internalLink",
+                "content": "II. Program-level access and network adequacy standards.",
+                "props": {
+                  "to": "/naaar"
+                }
+              }
+            ]
+          }
+        ],
+        "tableHeader": "Plan Name",
+        "enterEntityDetailsButtonText": "Enter"
       },
       "form": {
         "id": "pc",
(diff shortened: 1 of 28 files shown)


services/ui-src/src/components/reports/PlanOverlayReportPage.tsx
```diff
@@ -0,0 +1,174 @@
...
+          )}
+
+          {missingStandards() && (
+            <Box sx={sx.missingEntityMessage}>
+              {parseCustomHtml(verbiage.missingStandardsMessage || "")}
+            </Box>
+          )}
+
+          <Box sx={sx.dashboardBox}>
+            {planEntities.length !== 0 && (
```
bangbay-bluetiger, line 82: Then if we have `entityGroups` , we could do something like:
```
{entityGroups.map(({ entities, requiredMessage }) => {
  if (entities.length === 0) {
    return requiredMessage;
  }

  return (
    <Table>
       {entities.map((entity: EntityShape) => etc.
    </Table>
  );
})}
```

services/ui-src/src/components/reports/PlanOverlayReportPage.tsx
```diff
@@ -0,0 +1,174 @@
...
+  ReportPageFooter,
+  ReportPageIntro,
+  Table,
+} from "components";
+// types
+import { EntityShape, PlanOverlayReportPageShape } from "types";
+// utils
+import { parseCustomHtml, useBreakpoint, useStore } from "utils";
+
+export const PlanOverlayReportPage = ({ route, setSidebarHidden }: Props) => {
```
karla-vm, line 16: This is great! In terms of the naming, I think we could get ride of `Plan` here in case we end up using this type of page for a different entity type, so either `OverlayReportPage` or `EntityOverlayReportPage` to generalize the page?

karla-vm (dismissed):

services/ui-src/src/components/reports/PlanOverlayReportPage.tsx
ntsummers1: Bet! I'll go with OverlayReportPage :D

services/app-api/forms/naaar.json
```diff
@@ -994,6 +995,62 @@
...
+          {
+            "type": "p",
+            "children": [
+              {
+                "type": "html",
+                "content": "This program is missing required information. You won’t be able to access this section until you’ve completed all the required questions in section "
+              },
+              {
+                "type": "internalLink",
+                "content": "I. State and program information.",
```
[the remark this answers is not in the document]
ntsummers1: Oh good catch. Removed!

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -313,6 +314,7 @@ export const DynamicField = ({ name, label, ...props }: Props) => {
 interface Props {
   name: EntityType;
   label: string;
+  isRequired?: boolean;
```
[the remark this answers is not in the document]
ntsummers1: I used isRequired here since the prop that's being used elsewhere was also isRequired so I felt like it kept parity. Could rename all of them to isRequired, or maybe thats me going overboard. But I do like the idea of keeping everything the same.

jessabean (dismissed):
🚀

karla-vm (approved):

PR #12043, opened 2025-02-12
Add individual plan compliance

bangbay-bluetiger:
### Description

- Add individual plan compliance overlay
- Refactor `OverlayReportPage` to separate concerns for page vs overlay 
- Add `getEntityStatus` method, refactor where `getMlrEntityStatus` is called
- Add `ResponsiveEntityRow` that switches between `EntityRow` and `MobileEntityRow`
- Add `translate` util from MFP

### Related ticket(s)

CMDCT-4261

---
### How to test

1. Create a NAAAR
2. Add a plan
3. Go to III. Plan Compliance
4. Click "Enter" on your plan
5. Select Yes or No, save

### Notes

TODO:
- When "No" is selected, the `Enter` button in the table becomes active.
- `getNaaarEntityStatus` logic

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/components/forms/Form.tsx
+++ b/services/ui-src/src/components/forms/Form.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useEffect } from "react";
+import { ForwardedRef, forwardRef, ReactNode, useEffect } from "react";
 import {
   FieldValues,
   FormProvider,
@@ -29,18 +29,21 @@ import {
   useStore,
 } from "utils";
 
-export const Form = ({
-  id,
-  formJson,
-  onSubmit,
-  onError,
-  formData,
-  validateOnRender,
-  autosave,
-  dontReset,
-  children,
-  ...props
-}: Props) => {
+export const Form = forwardRef(function Form(
+  {
+    id,
+    formJson,
+    onSubmit,
+    onError,
+    formData,
+    validateOnRender,
+    autosave,
+    dontReset,
+    children,
+    ...props
+  }: Props,
+  ref?: ForwardedRef<HTMLFormElement>
+) {
   const { fields, options } = formJson;
 
   // determine if fields should be disabled (based on admin roles )
@@ -104,6 +107,7 @@ export const Form = ({
         id={id}
         autoComplete="off"
         onSubmit={form.handleSubmit(onSubmit as any, onError || onErrorHandler)}
+        ref={ref}
         {...props}
       >
         <Box sx={sx}>{renderFormFields(fields)}</Box>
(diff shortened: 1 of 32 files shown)


services/ui-src/src/components/forms/Form.tsx
```diff
@@ -29,18 +29,21 @@ import {
...
-  onSubmit,
-  onError,
-  formData,
-  validateOnRender,
-  autosave,
-  dontReset,
-  children,
-  ...props
-}: Props) => {
+export const Form = forwardRef(function Form(
```
gmrabian, line 32: I haven't used `forwardRef` before, could you provide a brief summary of what it affords us in this instance?
Also open to this being a 16th minute item in our dev sync

gmrabian (approved):
I really like the component refactoring and organization I've seen (around things like ResponsiveEntityRow)

I have one non-blocking, curiosity question. Otherwise this looks good to me!

services/ui-src/src/components/forms/Form.tsx
bangbay-bluetiger: On pages where there's a single form, we add the `id` of the form to the submit button. Example:
https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/8effa0aac2b3d7030c856d58adfd9724b7f2eeba/services/ui-src/src/components/overlays/EntityDetailsOverlay.tsx#L103-L105
In order to submit multiple forms from a single button, we can't use that method. So, I added a click action to loop through the forms and submit. I could gather the forms with regular JavaScript, but the React way would be to use `ref`. Advantage of doing this is the forms I'll want to get later are stored in a single variable with `useRef`, instead of needing to make JS calls every time I want to get them.

One `ref` can be for a single element or an array of them like I've done here:
https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/f5909b91834394c736abce8a681b96a4b81e189e/services/ui-src/src/components/overlays/EntityDetailsMultiformOverlay.tsx#L52
https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/f5909b91834394c736abce8a681b96a4b81e189e/services/ui-src/src/components/overlays/EntityDetailsMultiformOverlay.tsx#L167

In order to pass a `ref` from one component to another, it doesn't appear as a prop to the receiving component. You have to wrap the receiving component in `forwardRef` and `ref` is passed on as a second argument after props. (This is for functional components.) I think they're doing away with `forwardRef` in React 19, though, and `ref` will be available in props to receiving components.

gmrabian: Thanks for the information! So it allows us to submit multiple forms with one button, and to work with those forms at a higher level.

gmrabian (approved):

PR #12057, opened 2025-02-21
Multi banner

gmrabian:
### Description

- Allow multiple banners to exist at once in database
- Allow admins to view all existing banners, active and inactive
- Add logic to filter and sort banners so that most recent start date is active
- Adjust tests and api logic to fit pattern

### Related ticket(s)

CMDCT-4180

---
### How to test

- Log in as an admin
- Go to `/admin`
- Create a banner
- Create another banner
- Notice they both display in a list at the top
- Experiment with creating and deleting banners
- Visit the homepage to see what banner is active
  - It should always be the banner where today's date falls within the range and with the most recent start date

### Notes

There is still some logic to cleanup for this, like how the delete spinner shows on all the delete buttons, not just the one you clicked. Also multiple banners will have the "Active" text on the admin dashboard but only one will ever show on the homepage.

What do you think? Is this concept worth finalizing? Will admins be able to navigate this? How can we make it more clear? Is this even a use case admins really see themselves needing?

---
### Pre-review checklist

---
#### Review
---

### DIFF
--- a/services/app-api/handlers/banners/fetch.ts
+++ b/services/app-api/handlers/banners/fetch.ts
@@ -1,20 +1,14 @@
 import handler from "../handler-lib";
 import dynamoDb from "../../utils/dynamo/dynamodb-lib";
 // utils
-import { error } from "../../utils/constants/constants";
-import { badRequest, ok } from "../../utils/responses/response-lib";
+import { ok } from "../../utils/responses/response-lib";
 
-export const fetchBanner = handler(async (event, _context) => {
-  if (!event?.pathParameters?.bannerId!) {
-    return badRequest(error.NO_KEY);
-  }
-  const params = {
+export const fetchBanner = handler(async (_event, _context) => {
+  const scanParams = {
     TableName: process.env.BANNER_TABLE_NAME!,
-    Key: {
-      key: event.pathParameters.bannerId,
-    },
   };
-  const response = await dynamoDb.get(params);
+
+  const response = await dynamoDb.scanAll(scanParams);
 
   return ok(response);
 });
--- a/services/app-api/utils/dynamo/dynamodb-lib.ts
+++ b/services/app-api/utils/dynamo/dynamodb-lib.ts
@@ -9,6 +9,8 @@ import {
   QueryCommandInput,
   PutCommand,
   PutCommandInput,
+  ScanCommandInput,
+  ScanCommand,
 } from "@aws-sdk/lib-dynamodb";
 // utils
 import { logger } from "../debugging/debug-lib";
@@ -48,6 +50,19 @@ export default {
 
     do {
       const command = new QueryCommand({ ...params, ExclusiveStartKey });
+      const result = await client.send(command);
+      items = items.concat(result.Items ?? []);
+      ExclusiveStartKey = result.LastEvaluatedKey;
+    } while (ExclusiveStartKey);
+
+    return items;
+  },
+  scanAll: async (params: Omit<ScanCommandInput, "ExclusiveStartKey">) => {
+    let items: AnyObject[] = [];
+    let ExclusiveStartKey: Record<string, any> | undefined;
+
+    do {
+      const command = new ScanCommand({ ...params, ExclusiveStartKey });
       const result = await client.send(command);
       items = items.concat(result.Items ?? []);
       ExclusiveStartKey = result.LastEvaluatedKey;
--- a/services/ui-src/src/components/banners/AdminBannerProvider.tsx
+++ b/services/ui-src/src/components/banners/AdminBannerProvider.tsx
@@ -42,6 +42,10 @@ export const AdminBannerProvider = ({ children }: Props) => {
     setBannerLoading(true);
     try {
       const allBanners = await getBanners();
+      // sort by latest start date (assuming the most recent one should be active)
+      allBanners.sort(
+        (a: AdminBannerData, b: AdminBannerData) => b.startDate - a.startDate
+      );
       setAllBanners(allBanners);
     } catch {
       setBannerErrorMessage(bannerErrors.GET_BANNER_FAILED);
@@ -53,18 +57,15 @@ export const AdminBannerProvider = ({ children }: Props) => {
   const fetchAdminBanner = async () => {
     setBannerLoading(true);
     try {
-      const currentBanners = await getBanners();
+      let currentBanners = await getBanners();
       if (currentBanners.length > 1) {
-        const currentTimeStamp = Date.now();
-        // ensure banner should be active at current time
-        currentBanners.filter(
-          (banner: AdminBannerData) =>
-            banner.startDate < currentTimeStamp &&
-            banner.endDate > currentTimeStamp
+        // limit to banners active at current time
+        currentBanners = currentBanners.filter((banner: AdminBannerData) =>
+          checkDateRangeStatus(banner.startDate, banner.endDate)
         );
-        // sort by earliest start date
+        // sort by latest start date (assuming the most recent one should be active)
         currentBanners.sort(
-          (a: AdminBannerData, b: AdminBannerData) => a.startDate - b.startDate
+          (a: AdminBannerData, b: AdminBannerData) => b.startDate - a.startDate
         );
       }
       const bannerData = currentBanners[0] || {};
@@ -81,6 +82,7 @@ export const AdminBannerProvider = ({ children }: Props) => {
     setBannerDeleting(true);
     try {
       await deleteBanner(bannerKey);
+      await fetchAllBanners();
       await fetchAdminBanner();
     } catch {
       setBannerErrorMessage(bannerErrors.DELETE_BANNER_FAILED);
@@ -94,6 +96,7 @@ export const AdminBannerProvider = ({ children }: Props) => {
     } catch {
       setBannerErrorMessage(bannerErrors.CREATE_BANNER_FAILED);
     }
+    await fetchAllBanners();
     await fetchAdminBanner();
   };
 
--- a/services/ui-src/src/components/banners/AdminBannerProvider.tsx
+++ b/services/ui-src/src/components/banners/AdminBannerProvider.tsx
@@ -1,12 +1,10 @@
 import { createContext, ReactNode, useMemo, useEffect } from "react";
-// constants
-import { bannerId } from "../../constants";
 // types
 import { AdminBannerData, AdminBannerMethods } from "types/banners";
 // utils
 import {
   deleteBanner,
-  getBanner,
+  getBanners,
   writeBanner,
   useStore,
   checkDateRangeStatus,
@@ -14,17 +12,20 @@ import {
 // verbiage
 import { bannerErrors } from "verbiage/errors";
 
-const ADMIN_BANNER_ID = bannerId;
-
 export const AdminBannerContext = createContext<AdminBannerMethods>({
   fetchAdminBanner: Function,
+  fetchAllBanners: Function,
   writeAdminBanner: Function,
   deleteAdminBanner: Function,
 });
 
 export const AdminBannerProvider = ({ children }: Props) => {
+  const { userIsAdmin } = useStore().user ?? {};
+
   // state management
   const {
+    allBanners,
+    setAllBanners,
     bannerData,
     setBannerData,
     bannerActive,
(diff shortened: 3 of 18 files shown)


services/ui-src/src/components/pages/Admin/AdminPage.tsx
```diff
@@ -28,13 +28,40 @@ export const AdminPage = () => {
...
+    useStore();
+
+  const BannerPreview = (banner: AdminBannerData) => {
+    const bannerActive = checkDateRangeStatus(banner.startDate, banner.endDate);
+    return (
+      <React.Fragment key={banner.key}>
+        <Flex sx={sx.currentBannerInfo}>
+          <Text sx={sx.currentBannerStatus}>
+            Status:{" "}
+            <span className={bannerActive ? "active" : "inactive"}>
```
bangbay-bluetiger, line 41: If there are multiple banners, I think we should expand the states. Maybe to:

- active
- scheduled
- inactive

But I think it'd also be good to explain to the user that if they schedule a banner that overlaps with the active banner, it'll never become active. Or disallow creating overlapping banner dates.

Then, in the UI, instead of having all of the banners until Current Banner(s), it'd be better to separate them under three headings, like: Current, Scheduled, Past.

bangbay-bluetiger (commented):
I like where this is going!

services/app-api/utils/dynamo/dynamodb-lib.ts
```diff
@@ -55,6 +57,19 @@ export default {
...
+  scanAll: async (params: Omit<ScanCommandInput, "ExclusiveStartKey">) => {
+    let items: AnyObject[] = [];
+    let ExclusiveStartKey: Record<string, any> | undefined;
+
+    do {
+      const command = new ScanCommand({ ...params, ExclusiveStartKey });
+      const result = await client.send(command);
+      items = items.concat(result.Items ?? []);
+      ExclusiveStartKey = result.LastEvaluatedKey;
+    } while (ExclusiveStartKey);
```
benmartin-coforma, line 69: Non-blocking recommendation: The SDK's builtin `paginateScan` was made for just such an occasion.
```suggestion
    for await (let page of paginateScan({ client }, params)) {
      items = items.concat(page.Items ?? []);
    }
```

That would introduce a stylistic difference between `scanAll` and `queryAll`; if that bugs you then `paginateQuery` is also available.

services/app-api/utils/dynamo/dynamodb-lib.ts
```diff
@@ -9,6 +9,8 @@ import {
   QueryCommandInput,
   PutCommand,
   PutCommandInput,
+  ScanCommandInput,
+  ScanCommand,
```
benmartin-coforma, line 13: Of course, if you choose to do that, you'll need to import it.

```suggestion
  paginateScan,
```

services/ui-src/src/components/banners/AdminBannerProvider.tsx
```diff
@@ -37,12 +38,38 @@ export const AdminBannerProvider = ({ children }: Props) => {
...
   } = useStore();
 
+  const fetchAllBanners = async () => {
+    setBannerLoading(true);
+    try {
+      const allBanners = await getBanners();
+      // sort by latest start date (assuming the most recent one should be active)
+      allBanners.sort(
+        (a: AdminBannerData, b: AdminBannerData) => b.startDate - a.startDate
+      );
```
benmartin-coforma, line 48: Since `allBanners` is already typed as `AdminBannerData[]`, TS already knows the types required to `sort` them, so the type annotations on the lambda aren't doing us any good.
```suggestion
      allBanners.sort((a, b) => b.startDate - a.startDate);
```

The same would apply to the `filter` and `sort` lambdas in `fetchAdminBanner()` below.

services/ui-src/src/components/banners/AdminBannerProvider.tsx
```diff
@@ -37,12 +38,38 @@ export const AdminBannerProvider = ({ children }: Props) => {
...
-      const newBannerData = currentBanner?.Item || {};
-      setBannerData(newBannerData);
+      let currentBanners = await getBanners();
+      if (currentBanners.length > 1) {
+        // limit to banners active at current time
+        currentBanners = currentBanners.filter((banner: AdminBannerData) =>
+          checkDateRangeStatus(banner.startDate, banner.endDate)
+        );
+        // sort by latest start date (assuming the most recent one should be active)
+        currentBanners.sort(
```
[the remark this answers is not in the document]
benmartin-coforma: My two cents is that any code spent managing a new flag would be more trouble than it's worth.

When we talk about having multiple banners with overlapping start and end dates, I feel we're already talking about an unlikely situation (given that the site has only one user creating banners, and that user knows what they're doing). Taking our best guess at what they intended and moving ahead seems like a very much good-enough course of action here.

The only thing I might change would be to sort by create date rather than start date. In this (again, unlikely) scenario, I would guess that the most-recently-saved banner is the one our admin is most likely to want to see. Especially given that the banner start dates could easily be identical, down to the millisecond, due to how we calculate them. Unfortunately, while we do have `createdAt` in the banner data, we don't have it in the type definition, so this isn't a trivial change.

services/ui-src/src/components/banners/AdminBannerProvider.tsx
```diff
@@ -37,12 +38,38 @@ export const AdminBannerProvider = ({ children }: Props) => {
...
+          (a: AdminBannerData, b: AdminBannerData) => b.startDate - a.startDate
+        );
+      }
+      const bannerData = currentBanners[0] || {};
+      setBannerData(bannerData);
```
benmartin-coforma, line 72: Non-blocking suggestion: This is definitely a style consideration and not a correctness one, but the `if length > 0` check isn't doing much, and I like the look of `.find(...)` over `.filter(...)[0]`. So I might have written something more like this:
```suggestion
      // Find the most recent currently-active banner
      const currentBanner = currentBanners
        .sort((a, b) => b.startDate - a.startDate)
        .find((banner) => checkDateRangeStatus(bannner.startDate, banner.endDate));
      setBannerData(currentBanner ?? {});
```

Note that if you apply this suggestion as-is, Typescript will begin to complain that `{}` isn't a real banner. I think the right move might be to remove the `?? {}` altogether, since `setBannerData()` claims to accept `undefined`, but I haven't tested that or looking into possible consequences.

(The reason that TS isn't complaining right now is because it assumes that `currentBanners[0]` will always be defined, which is not true)

services/ui-src/src/components/pages/Admin/AdminPage.tsx
```diff
@@ -46,50 +73,17 @@ export const AdminPage = () => {
...
-                      )}
-                    </Button>
-                  </Flex>
-                </>
-              )}
+            <Collapse in={allBanners && allBanners?.length > 0}>
+              {allBanners?.map((banner) => BannerPreview(banner))}
             </Collapse>
-            {!bannerData?.key && <Text>There is no current banner</Text>}
+            {!allBanners?.length && <Text>There is no current banner</Text>}
```
benmartin-coforma, line 86: Suggestion: just a language tweak here, to make sure the user understands we're looking for past & future banners as well.

```suggestion
            {!allBanners?.length && <Text>There are no existing banners</Text>}
```

benmartin-coforma (commented):
I know this PR is still in draft, but I like where you're heading with it!

services/ui-src/src/components/pages/Admin/AdminPage.tsx
gmrabian: Good point. I updated the display text to have three different states. I will see about organizing the display to match!

services/ui-src/src/components/banners/AdminBannerProvider.tsx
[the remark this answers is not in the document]
gmrabian: I did play around with setting a boolean on the banner for active state, but I couldn't come up with a clean way to have the admin user select the one to make active. the start/end date already limit it, and it felt clunky to have a button on each for "make banner active"

gmrabian:
The playwright test setup for the banner page is tricky because we need to iterate over delete buttons and click all of them. However, I think this is a similar issue we had with the admin archive setup. I will see what we can do, but it may depend on what we discover with that test.

karla-vm:
This is looking great, Garrett!! ✨ in terms of admins using this, the client had asked for this based on a specific release where we had a banner up until a Friday and then another banner that needed to be showing that following Monday morning. We _could_ use validation to disallow users from creating overlapping banners, but showing the latest is a good enough solution in my opinion. This is likely not a feature that will be used extensively, so I wouldn't be too concerned about users being able to navigate it.

gmrabian:
Note to self: make it so an error pops up (and cannot submit) if banner overlaps with existing banner

gmrabian:
Still need to fix playwright tests but otherwise this is ready for review!

ntsummers1 (approved):
Loaded in as an admin, made banners, got told I couldn't make em when they were between other active banner dates. 10/10 would error out again

services/app-api/handlers/banners/fetch.ts
```diff
@@ -1,20 +1,14 @@
...
-import { error } from "../../utils/constants/constants";
-import { badRequest, ok } from "../../utils/responses/response-lib";
+import { ok } from "../../utils/responses/response-lib";
 
-export const fetchBanner = handler(async (event, _context) => {
-  if (!event?.pathParameters?.bannerId!) {
-    return badRequest(error.NO_KEY);
-  }
-  const params = {
+export const fetchBanner = handler(async (_event, _context) => {
```
bangbay-bluetiger, line 6: Nit: Since neither `event` or `context` are used, do we need to include them? Usually would only use `_` if there's an arg we're using after one we're not.

services/app-api/utils/dynamo/dynamodb-lib.ts
```diff
@@ -55,6 +57,14 @@ export default {
 
     return items;
   },
+  scanAll: async (params: Omit<ScanCommandInput, "ExclusiveStartKey">) => {
+    let items: AnyObject[] = [];
+    for await (let page of paginateScan({ client }, params)) {
```
bangbay-bluetiger, line 62: Nit: `const page`

services/ui-src/src/components/banners/AdminBannerProvider.tsx
```diff
@@ -37,12 +38,31 @@ export const AdminBannerProvider = ({ children }: Props) => {
...
+    }
+  };
+
   const fetchAdminBanner = async () => {
     setBannerLoading(true);
     try {
-      const currentBanner = await getBanner(ADMIN_BANNER_ID);
-      const newBannerData = currentBanner?.Item || {};
-      setBannerData(newBannerData);
+      let currentBanners = await getBanners();
```
bangbay-bluetiger, line 58: Nit: `const currentBanners`

services/ui-src/src/components/pages/Admin/AdminPage.tsx
```diff
@@ -43,53 +87,20 @@ export const AdminPage = () => {
...
-                      {bannerDeleting ? (
-                        <Spinner size="md" />
-                      ) : (
-                        "Delete Current Banner"
-                      )}
-                    </Button>
-                  </Flex>
-                </>
-              )}
+            <Collapse in={allBanners && allBanners?.length > 0}>
```
bangbay-bluetiger, line 100: Are you able to do this?
```
<Collapse in={allBanners?.length > 0}>
  {allBanners.map((banner) => BannerPreview(banner))}
```

services/ui-src/src/verbiage/pages/admin.ts
```diff
@@ -1,6 +1,6 @@
 export default {
   intro: {
     header: "Banner Admin",
-    body: "Manage the announcement banner below.",
+    body: "<p>To create an announcement banner for the top of the MCR portal page, enter the information below.</p><br/><p>Multiple banners can be created and scheduled as long as the dates do not overlap with other banners. Only one Active banner can appear at a time. For example, if your Active banner has an end date of 03/17/2025, the start date of the next banner must be 03/18/25 or after.</p>",
```
bangbay-bluetiger, line 4: This is putting the paragraphs inside a `span`, which isn't valid semantic HTML. Looks like `parseCustomHtml` automatically does that unless you add use `type` and `content`. Also would be in favor of removing the `<br>` and adding a style to the parent `<Box />.

bangbay-bluetiger (commented):
Functionality and style looks good to me. Some comments, mostly nits.

karla-vm (dismissed):
🚀

ntsummers1 (dismissed):

services/ui-src/src/components/pages/Admin/AdminPage.tsx
gmrabian: This gives a Typescript error:
`'allBanners.length' is possibly 'undefined'.`

bangbay-bluetiger: With the coalescer `allBanners?.length`?

karla-vm (approved):

PR #12071, opened 2025-03-11
Cmdct 4298 - add captions to tables in exported reports

britt-mo:
### Description
 Tables have empty <caption> tags, thus not getting labelled. This makes it harder for screen readers to be able to navigate to a specific table that they want

Every report now has captions (visually hidden) on the review& submit page table along with the exported tables

### Related ticket(s)

CMDCT-4298

---
### How to test
Create a report (MLR and MCR), use the inspector to verify that tables have captions in the review & submit page and in the exported report. NAAR was also updated on the review & submit page.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/export/ExportedEntityDetailsOverlaySection.tsx
+++ b/services/ui-src/src/components/export/ExportedEntityDetailsOverlaySection.tsx
@@ -30,8 +30,6 @@ export const ExportedEntityDetailsOverlaySection = ({
   const { report } = useStore();
   const entityType = section.entityType;
 
-  // console.log(section.verbiage.intro);
-
   return (
     <Box sx={sx.sectionHeading} {...props}>
       <ExportedSectionHeading
--- a/services/ui-src/src/components/export/ExportedEntityDetailsOverlaySection.tsx
+++ b/services/ui-src/src/components/export/ExportedEntityDetailsOverlaySection.tsx
@@ -30,6 +30,8 @@ export const ExportedEntityDetailsOverlaySection = ({
   const { report } = useStore();
   const entityType = section.entityType;
 
+  // console.log(section.verbiage.intro);
+
   return (
     <Box sx={sx.sectionHeading} {...props}>
       <ExportedSectionHeading
@@ -146,6 +148,7 @@ export function getEntityTableComponents(
                 fields={filteredFields.slice(1) as FormField[]}
                 entity={entity}
                 showHintText={false}
+                caption={header.props?.content}
               />
             </Fragment>
           );
--- a/services/ui-src/src/components/export/ExportedEntityDetailsTable.tsx
+++ b/services/ui-src/src/components/export/ExportedEntityDetailsTable.tsx
@@ -44,7 +44,7 @@ export const ExportedEntityDetailsTable = ({
       {...props}
       sx={sx.root}
       content={{
-        caption: caption,
+        caption,
         headRow: threeColumnHeaderItems,
       }}
     >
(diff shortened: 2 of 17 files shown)


services/ui-src/src/components/export/ExportedEntityDetailsOverlaySection.tsx
```diff
@@ -30,6 +30,8 @@ export const ExportedEntityDetailsOverlaySection = ({
   const { report } = useStore();
   const entityType = section.entityType;
 
+  // console.log(section.verbiage.intro);
```
gmrabian, line 33: reminder to remove this

gmrabian (commented):
Some of the table names are duplicates and the roman numerals are applied inconsistently. It occurs to me that this is how the names display on the screen as well, just calling this out in case it's a particular concern for a list of items in a voiceover menu.

Otherwise looks good! Impressed with how little code it took to get all the captions :)


file

services/ui-src/src/components/export/ExportedEntityDetailsTable.tsx
```diff
@@ -43,6 +44,7 @@ export const ExportedEntityDetailsTable = ({
       {...props}
       sx={sx.root}
       content={{
+        caption: caption,
```
bangbay-bluetiger, line 47: Nit: Could be just `caption,` instead of `caption: caption,` if we want to be less redundant.

cy-by:
> Some of the table names are duplicates and the roman numerals are applied inconsistently. It occurs to me that this is how the names display on the screen as well, just calling this out in case it's a particular concern for a list of items in a voiceover menu.
> 
> Otherwise looks good! Impressed with how little code it took to get all the captions :)
> 
> file

---
Good point @gmrabian, the headers being duplicated and the numerals are a problem as far as content/design/understand-ability.

From this PR's perspective, the `<caption>` should reflect the same content as the headings, as that is the content the tables hold.

So I think this PR is good to go 👍🏻 , and we'll need to discuss the content issue with design/product.

cy-by (approved):

services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
```diff
@@ -303,8 +303,12 @@ const entityTable = (
 ) => {
   const { isTablet, isMobile } = useBreakpoint();
   const tableHeaders = () => {
-    if (isTablet || isMobile) return { headRow: ["", ""] };
-    return { headRow: ["", verbiage.tableHeader!, ""] };
+    if (isTablet || isMobile)
+      return { caption: verbiage.tableHeader, headRow: ["", ""] };
```
bangbay-bluetiger, line 307: Instead of empty head rows, I think these should have use `{ hiddenName: something }` instead. What are actually in those rows?

services/ui-src/src/components/reports/ModalOverlayReportPage.tsx
```diff
@@ -64,8 +64,12 @@ export const ModalOverlayReportPage = ({
   const reportFieldDataEntities = report?.fieldData[entityType] || [];
   const dashTitle = `${verbiage.dashboardTitle} ${reportFieldDataEntities.length}`;
   const tableHeaders = () => {
-    if (isTablet || isMobile) return { headRow: ["", ""] };
-    return { headRow: ["", verbiage.tableHeader, ""] };
+    if (isTablet || isMobile)
+      return { caption: verbiage.tableHeader, headRow: ["", ""] };
```
bangbay-bluetiger, line 68: Same thing using `hiddenName`.

services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
britt-mo: Just made that change

bangbay-bluetiger (approved):

gmrabian (approved):

PR #12079, opened 2025-03-18
Add compareText util

bangbay-bluetiger:
### Description

- Add `otherSpecify` util because it's a pattern we use often
- Add `compareText` util as what `otherSpecify` uses under the hood
- Limiting this PR to `otherSpecify` uses and a couple of nearby cases where `compareText` seems useful

---
### How to test

1. `yarn test` in `ui-src`
2. All previous tests plus new `compareText` tests should pass

### Notes

---
### Pre-review checklist


gmrabian (approved):
Nice helper! 

Figuring out which value returns felt like a bit of a logic teaser at first 😆

britt-mo (approved):
Looks good, tests passed locally

PR #12102, opened 2025-04-04
Render NAAAR standards in a table in export view

gmrabian:
### Description

I added comments to explain the key code changes. If you review by commit you'll see the core changes in the first commit, and some type updates and tests in the rest.

- Add custom branch to render standards page using modal overlay export page type (it's the closest to what we wanted)
  - The standards page is a `drawer` page type. What do you think of this? Is it ok to render another page type for the export page logically? I thought it would save us lines of code and be more reusable. Perhaps our export types don't perfectly align with our form page types?
- Add NAAAR standards logic to `ModalOverlay` export page
- Add display name to screen reader header type and rename
- Add tests

### Related ticket(s)

CMDCT-4464

---
### How to test

Open the [deployed environment](https://d6dq6fzvsiy1v.cloudfront.net/)
- Log in as `stateuser@test.com`
- Go to the existing NAAAR titled "standards"
- Go to the PDF view (Review & Submit -> Review PDF)
- View the standards table and verify it matches the [Figma](https://www.figma.com/design/QcoVlfZ2adUBnzcHiI0G0S/MDCT-MCR-NAAAR?node-id=1356-27373&t=3x4VxLNRMgWtuzKX-4)
- Verify the counter indicates the correct number of standards
- Leave this report and open the NAAAR report titled "no standards"
- Go to the PDF view (Review & Submit -> Review PDF)
- Verify the table indicates its empty, and the counter is set to 0 with a custom message

Bonus:
- Turn on VoiceOver and verify the `Pop.` column header reads out "Population" 

### Notes

I'm not sure what the best way is to color this text mid-line so I left it all black.
Figma:
Screenshot 2025-04-04 at 11 27 43 AM
Deployed instance:
Screenshot 2025-04-04 at 11 56 21 AM

---
### Pre-review checklist

---
#### Review
---

### DIFF
--- a/services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
+++ b/services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from "@testing-library/react";
+import { render, screen, within } from "@testing-library/react";
 // components
 import {
   ExportedModalOverlayReportSection,
@@ -208,38 +208,54 @@ describe("<ExportedModalOverlayReportSection />", () => {
 
     test("Should render data correctly for naaar", async () => {
       mockedUseStore.mockReturnValue(mockNaaarReportStore);
-      const { container, findByText } = render(exportedNaaarStandardsComponent);
+      render(exportedNaaarStandardsComponent);
 
       // All table headers are present
-      expect(container.querySelectorAll("th").length).toBe(7);
+      expect(screen.getAllByRole("columnheader").length).toBe(7);
 
       // Every entity has a row (+1 for header)
-      expect(container.querySelectorAll("tr").length).toBe(
-        mockNaaarReportStore.report?.fieldData.standards.length + 1
+      const tbody = screen.getAllByRole("rowgroup")[1];
+      const rows = within(tbody).getAllByRole("row");
+      expect(rows.length).toBe(
+        mockNaaarReportStore.report?.fieldData.standards.length
       );
 
       // index
-      expect(await findByText("1")).toBeVisible();
+      expect(screen.getByRole("gridcell", { name: "1" })).toBeVisible();
 
       // provider type
       expect(
-        await findByText("Mock Provider; mock provider details")
+        screen.getByRole("gridcell", {
+          name: "Mock Provider; mock provider details",
+        })
       ).toBeVisible();
 
       // standard type
-      expect(await findByText("mock standard type")).toBeVisible();
+      expect(
+        screen.getByRole("gridcell", { name: "mock standard type" })
+      ).toBeVisible();
 
       // description
-      expect(await findByText("description of standard")).toBeVisible();
+      expect(
+        screen.getByRole("gridcell", { name: "description of standard" })
+      ).toBeVisible();
 
       // analysis methods, joined with a comma
-      expect(await findByText("Mock am 1, Mock am 2, Mock am 3")).toBeVisible();
+      expect(
+        screen.getByRole("gridcell", {
+          name: "Mock am 1, Mock am 2, Mock am 3",
+        })
+      ).toBeVisible();
 
       // population
-      expect(await findByText("Mock population")).toBeVisible();
+      expect(
+        screen.getByRole("gridcell", { name: "Mock population" })
+      ).toBeVisible();
 
       // region
-      expect(await findByText("Mock region")).toBeVisible();
+      expect(
+        screen.getByRole("gridcell", { name: "Mock region" })
+      ).toBeVisible();
     });
 
     test("Should render message for naaar with no standards", async () => {
@@ -248,17 +264,17 @@ describe("<ExportedModalOverlayReportSection />", () => {
       };
       mockEmptyStandardsNaaarStore.report!.fieldData.standards = undefined;
       mockedUseStore.mockReturnValue(mockEmptyStandardsNaaarStore);
-      const { container, findByText } = render(exportedNaaarStandardsComponent);
+      render(exportedNaaarStandardsComponent);
 
       // No table renders
-      expect(container.querySelectorAll("table").length).toBe(0);
+      expect(screen.queryByRole("table")).toBeNull();
 
       // region
       expect(
-        await findByText(
-          "Standard total count: 0 - No access standards entered"
-        )
-      ).toBeVisible();
+        screen.getByTestId("exportedModalOverlayReportSection")
+      ).toHaveTextContent(
+        "Standard total count: 0 - No access standards entered"
+      );
     });
 
     test('Should render "other" explanations if they are filled.', async () => {
--- a/services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
+++ b/services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
@@ -242,6 +242,25 @@ describe("<ExportedModalOverlayReportSection />", () => {
       expect(await findByText("Mock region")).toBeVisible();
     });
 
+    test("Should render message for naaar with no standards", async () => {
+      const mockEmptyStandardsNaaarStore = {
+        ...mockNaaarReportStore,
+      };
+      mockEmptyStandardsNaaarStore.report!.fieldData.standards = undefined;
+      mockedUseStore.mockReturnValue(mockEmptyStandardsNaaarStore);
+      const { container, findByText } = render(exportedNaaarStandardsComponent);
+
+      // No table renders
+      expect(container.querySelectorAll("table").length).toBe(0);
+
+      // region
+      expect(
+        await findByText(
+          "Standard total count: 0 - No access standards entered"
+        )
+      ).toBeVisible();
+    });
+
     test('Should render "other" explanations if they are filled.', async () => {
       mockReportContextOther.report.fieldData.program = [mockMlrProgramOther];
 
--- a/services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
+++ b/services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
@@ -5,12 +5,13 @@ import {
   renderModalOverlayTableBody,
 } from "./ExportedModalOverlayReportSection";
 // types
-import { ModalOverlayReportPageShape, ReportType } from "types";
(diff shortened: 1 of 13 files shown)


services/ui-src/src/components/tables/Table.tsx
```diff
@@ -52,9 +52,12 @@ export const Table = ({
...
                   {typeof headerCell === "object" ? (
-                    <VisuallyHidden>
-                      {sanitizeAndParseHtml(headerCell.hiddenName)}
-                    </VisuallyHidden>
+                    <>
+                      <VisuallyHidden>
+                        {sanitizeAndParseHtml(headerCell.hiddenName)}
+                      </VisuallyHidden>
+                      {headerCell?.name ? headerCell.name : null}
+                    </>
```
gmrabian, line 60: I added the `name` attribute so that you could still have an aria label and have the visual name be different.

services/ui-src/src/components/export/ExportedReportWrapper.tsx
```diff
@@ -27,6 +28,16 @@ export const ExportedReportWrapper = ({ section }: Props) => {
...
+      // standards uses the table pattern from the modal overlay component
+      if (section.entityType === EntityType.STANDARDS) {
+        return (
+          <>
+            <ExportedModalOverlayReportSection
+              section={section as ModalOverlayReportPageShape}
+            />
+          </>
+        );
+      }
```
gmrabian, line 40: Modal Overlay was really close to the pdf layout for the standards data so I thought we should use that, even though it's a drawer page type in the report.
What do you think?

services/ui-src/src/components/export/ExportedModalOverlayReportSection.tsx
```diff
@@ -114,13 +130,37 @@ export function renderModalOverlayTableBody(
           </Tr>
         );
       });
-    case ReportType.MCPAR:
     case ReportType.NAAAR:
-      throw new Error(
-        `The modal overlay table headers for report type '${reportType}' have not been implemented.`
-      );
+      return entities.map((entity, idx) => {
```
gmrabian, line 134: retrieve and map standard data to table data within row. Great way to reuse `mapNaaarStandardsData`!

services/ui-src/src/components/export/ExportedModalOverlayReportSection.tsx
```diff
@@ -31,24 +40,35 @@ export const ExportedModalOverlayReportSection = ({ section }: Props) => {
...
+    <Box data-testid="exportedModalOverlayReportSection">
+      {entityType === EntityType.STANDARDS && (
+        <Text sx={sx.standardCount}>
+          Standard total count:{" "}
+          {entityCount > 0
+            ? entityCount
+            : verbiage.emptyEntityMessage[entityType]}
+        </Text>
+      )}
+      {entityType === EntityType.STANDARDS && !entityCount ? null : (
```
gmrabian, line 52: per design: only render table if there are standards to show

services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
```diff
@@ -196,6 +206,61 @@ describe("<ExportedModalOverlayReportSection />", () => {
       expect(await findByText(mockMlrProgram.report_miscellaneousNotes));
     });
 
+    test("Should render data correctly for naaar", async () => {
+      mockedUseStore.mockReturnValue(mockNaaarReportStore);
+      const { container, findByText } = render(exportedNaaarStandardsComponent);
+
+      // All table headers are present
+      expect(container.querySelectorAll("th").length).toBe(7);
```
bangbay-bluetiger, line 214: You can use `screen.getAllByRole("columnheader").length` .

services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
```diff
@@ -196,6 +206,61 @@ describe("<ExportedModalOverlayReportSection />", () => {
...
+
+      // Every entity has a row (+1 for header)
+      expect(container.querySelectorAll("tr").length).toBe(
```
bangbay-bluetiger, line 217: If we want to count just the body rows, I'd do something like this:
```
const tbody = screen.getAllByRole("rowgroup")[1];
const rows = within(tbody).getAllByRole("row");
expect(rows.length).toBe(mockNaaarReportStore.report?.fieldData.standards.length);
```

services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
```diff
@@ -196,6 +206,61 @@ describe("<ExportedModalOverlayReportSection />", () => {
...
+        mockNaaarReportStore.report?.fieldData.standards.length + 1
+      );
+
+      // index
+      expect(await findByText("1")).toBeVisible();
```
bangbay-bluetiger, line 222: For these text finds, you could replace with `screen.getByRole("gridcell", { name: textToFind })`.

services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
```diff
@@ -196,6 +206,61 @@ describe("<ExportedModalOverlayReportSection />", () => {
...
+    test("Should render message for naaar with no standards", async () => {
+      const mockEmptyStandardsNaaarStore = {
+        ...mockNaaarReportStore,
+      };
+      mockEmptyStandardsNaaarStore.report!.fieldData.standards = undefined;
+      mockedUseStore.mockReturnValue(mockEmptyStandardsNaaarStore);
+      const { container, findByText } = render(exportedNaaarStandardsComponent);
+
+      // No table renders
+      expect(container.querySelectorAll("table").length).toBe(0);
```
bangbay-bluetiger, line 254: I'd use `expect(screen.queryByRole("table")).toBeNull()`.

services/ui-src/src/components/export/ExportedModalOverlayReportSection.test.tsx
```diff
@@ -196,6 +206,61 @@ describe("<ExportedModalOverlayReportSection />", () => {
...
+
+      // region
+      expect(
```
bangbay-bluetiger, line 257: I'd use:
```
expect(screen.getByTestId("exportedModalOverlayReportSection"))
  .toHaveTextContent("Standard total count: 0 - No access standards entered");
```

services/ui-src/src/components/tables/Table.tsx
bangbay-bluetiger: It seems like if there's visible text we shouldn't be using `VisuallyHidden`. What if we did something like this instead:
```
{!headerCell.name && (
  <VisuallyHidden>
    {sanitizeAndParseHtml(headerCell.hiddenName)}
  </VisuallyHidden>
)}

{headerCell.name && (
  <Text as="span" aria-label={headerCell.hiddenName}>
    {headerCell.name}
  </Text>
)}

```

gmrabian: I tried the suggestion locally and was having trouble with voiceover where it would announce:
"Population group, pop., end of population group"

I think when we have the Text element it changes how the voiceover reads it? I tried both as span and as p.

As a result I left the change as is.

services/ui-src/src/components/export/ExportedReportWrapper.tsx
```diff
@@ -27,6 +28,16 @@ export const ExportedReportWrapper = ({ section }: Props) => {
         </Box>
       );
     case PageTypes.DRAWER:
+      // standards uses the table pattern from the modal overlay component
+      if (section.entityType === EntityType.STANDARDS) {
+        return (
+          <>
```
bangbay-bluetiger, line 34: Nit: Extra `<>` wrapper not needed.

services/ui-src/src/components/export/ExportedReportWrapper.tsx
bangbay-bluetiger: This seems fine to me.

bangbay-bluetiger (approved):
Looks good to me.

bangbay-bluetiger (approved):

karla-vm (dismissed):

ntsummers1:
```
Add custom branch to render standards page using modal overlay export page type (it's the
closest to what we wanted)
The standards page is a drawer page type. What do you think of this? Is it ok to render another 
page type for the export page logically? I thought it would save us lines of code and be more 
reusable. Perhaps our export types don't perfectly align with our form page types?
```
Personally I don't like it since its a hard break in the pattern. However, deadline is coming up and I'm not one to stop progress over something like that. But in a world where these forms are getting more and more complex and we're passing and injecting questions, any semblance of constancy and clear seperation would help. I'd love to circle back to this in the future when we're in a lul period. Like the ticket could chill for 6 months as a nice to have backlog ticket for a rainy day type thing. As of right now, I'm totally cool with this approach. Save lines, save time, meet deadline, and circle back when the time feels right.

services/ui-src/src/components/export/ExportedModalOverlayReportSection.tsx
```diff
@@ -114,13 +134,38 @@ export function renderModalOverlayTableBody(
...
+      // render pattern for NAAAR Standards
+      return entities.map((entity, idx) => {
+        const {
+          provider,
+          standardType,
+          description,
+          analysisMethods,
+          population,
+          region,
+        } = mapNaaarStandardsData<NaaarStandardsTableShape>([entity])[0];
```
ntsummers1, line 147: Is it ever possible that this [0] can not be there? Double checking cause that'll hard error the user out.

ntsummers1 (approved):
A question and comment to note, but nothing blocking here. Looks great overall!

services/ui-src/src/components/export/ExportedModalOverlayReportSection.tsx
gmrabian: Hmm good callout 🤔 

since we are within `entities.map()` I believe entity must be defined, and in that case this function will return an array with an object of the relevant info. I also think the entity (Standards) will have to be complete because it's a drawer form, which can't submit without all fields.

I think if `entity` here were undefined or incomplete it might be a problem with the report as whole.

bangbay-bluetiger (approved):

PR #12103, opened 2025-04-04
Cmdct 4264

britt-mo:
### Description

 Provide details about plan non-compliance and exceptions for this standard for 42 C.F.R. § 438.68:*

When the user checks the checkbox labeled Standard is non-compliant for this plan, they should see a nested form with a section for analysis methods:

`Label: Plan deficiencies: 42 C.F.R. § 438.68 analyses used to identify deficiencies`

`Hint: Indicate which analyses uncovered the deficiencies. If the analysis method does not appear here, go back to “Standards” and add it.`

`ChoiceList options (checkboxes): The list of analysis methods that apply to the plan (selected during the Add Analysis Method flow in Section I)`

### Related ticket(s)

CMDCT-4264

---
### How to test
Go into a plan and select 'no', then go to the non-compliance section and select the first checkbox, the analysis methods that were previously selected should display as checkbox options

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/app-api/forms/naaar.json
+++ b/services/app-api/forms/naaar.json
@@ -1308,8 +1308,24 @@
                               "parentOptionId": "l8yWzYniMtH5HQrBIjwApO"
                             },
                             "props": {
-                              "hint": "Indicate which analyses uncovered the deficiencies. If the analysis method does not appear here, go back to “Standards” and add it.",
                               "label": "Plan deficiencies: 42 C.F.R. § 438.68 analyses used to identify deficiencies",
+                              "hint": [
+                                {
+                                  "type": "html",
+                                  "content": "Indicate which analyses uncovered the deficiencies. If the analysis method does not appear here, go back to "
+                                },
+                                {
+                                  "type": "internalLink",
+                                  "content": "Standards",
+                                  "props": {
+                                    "to": "/naaar/program-level-access-and-network-adequacy-standards"
+                                  }
+                                },
+                                {
+                                  "type": "html",
+                                  "content": " and add it."
+                                }
+                              ],
                               "choices": []
                             }
                           },
--- a/services/app-api/forms/naaar.json
+++ b/services/app-api/forms/naaar.json
@@ -1255,7 +1255,7 @@
                 "description": { "header": "Standard description" },
                 "population": { "header": "Population" },
                 "region": { "header": "Region" },
-                "actions": { "header": "Actions", "hidden": true }  
+                "actions": { "header": "Actions", "hidden": true }
               },
               "verbiage": {
                 "backButton": "Return to plan compliance data for {{planName}} dashboard",
@@ -1296,7 +1296,22 @@
                             "props": {
                               "hint": "Describe plan deficiencies identified during the reporting period.",
                               "label": "Plan deficiencies: 42 C.F.R. § 438.68 description"
-                            }                          
+                            }
+                          },
+                          {
+                            "id": "analysis_method_applicable_plans",
+                            "type": "checkbox",
+                            "validation": {
+                              "type": "checkbox",
+                              "nested": true,
+                              "parentFieldName": "planCompliance43868_standard",
+                              "parentOptionId": "l8yWzYniMtH5HQrBIjwApO"
+                            },
+                            "props": {
+                              "hint": "Indicate which analyses uncovered the deficiencies. If the analysis method does not appear here, go back to “Standards” and add it.",
+                              "label": "Plan deficiencies: 42 C.F.R. § 438.68 analyses used to identify deficiencies",
+                              "choices": []
+                            }
                           },
                           {
                             "id": "planCompliance43868_standard-nonCompliancePlanToAchieveCompliance",
--- a/services/app-api/forms/naaar.json
+++ b/services/app-api/forms/naaar.json
@@ -1299,7 +1299,7 @@
                             }
                           },
                           {
-                            "id": "analysis_method_applicable_standard-plans_nonComplianceAnalyses",
+                            "id": "planCompliance43868_standard-nonComplianceAnalyses",
                             "type": "checkbox",
                             "validation": {
                               "type": "checkbox",
--- a/services/app-api/forms/naaar.json
+++ b/services/app-api/forms/naaar.json
@@ -1299,7 +1299,7 @@
                             }
                           },
                           {
-                            "id": "analysis_method_applicable_plans",
+                            "id": "analysis_method_applicable_standard-plans_nonComplianceAnalyses",
                             "type": "checkbox",
                             "validation": {
                               "type": "checkbox",
--- a/services/ui-src/src/utils/forms/dynamicItemFields.ts
+++ b/services/ui-src/src/utils/forms/dynamicItemFields.ts
@@ -176,14 +176,11 @@ const updatedItemChoiceList = (
 // dynamically filter by partialId to find the analysis methods
 export const availableAnalysisMethods = (
   analysisMethodsFieldId: string,
-  items: string[]
+  items: AnyObject[]
 ) => {
-  const updatedItemChoices: AnyObject[] = [];
-  items.forEach((item) => {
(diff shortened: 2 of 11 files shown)


services/ui-src/src/utils/forms/dynamicItemFields.ts
```diff
@@ -172,3 +172,27 @@ const updatedItemChoiceList = (
...
 };
+
+// dynamically generate fields for choices
+export const generateAnalysisMethodFields = (
+  form: FormJson,
+  items: AnyObject[]
+) => {
+  const choicesToInject = availableAnalysisMethods(items);
+  form!.fields[0]!.props!.choices[0].children[1].props.choices =
+    choicesToInject;
```
gmrabian, line 183: this isn't pretty but we have done it this way before.
An alternative is to filter by the keys for the arrays, so that at least if the order changes we still end up at the right one. That said, it'd be more verbose and not necessarily nicer better to read.

At the very least I'd add a comment here indicating which fieldId it's targeting so that a developer can more easily reference it back in the naaar.json

gmrabian (commented):
I did notice a bug that when you delete a custom analysis method it doesn't remove it from the list, and it still shows up in the saved data (actually the one that's in the saved data is one I deleted previous to the video!).
I think this is a known issue other places as well, so maybe something we can ticket and work out at bug bash, unless you see a quick way to do it here. (would useEffect help us 🤔 or would it end up resetting it too much?)

https://github.com/user-attachments/assets/54bc9bdb-c897-4cee-964c-94591040d407

services/app-api/forms/naaar.json
```diff
@@ -1296,7 +1296,22 @@
                             "props": {
                               "hint": "Describe plan deficiencies identified during the reporting period.",
                               "label": "Plan deficiencies: 42 C.F.R. § 438.68 description"
-                            }                          
+                            }
+                          },
+                          {
+                            "id": "analysis_method_applicable_standard-plans_nonComplianceAnalyses",
```
bangbay-bluetiger, line 1302: I'd use `id: planCompliance43868_standard-nonComplianceAnalyses` here because the convention is to name every field starting with `planCompliance43868_standard`. Then, `addStandardId` will change this id to `planCompliance43868_standard-{{standardId}}-nonComplianceAnalyses` on render.

services/ui-src/src/utils/forms/dynamicItemFields.ts
```diff
@@ -172,3 +172,27 @@ const updatedItemChoiceList = (
...
   return updatedChoiceList;
 };
+
+// dynamically generate fields for choices
+export const generateAnalysisMethodFields = (
+  form: FormJson,
+  items: AnyObject[]
+) => {
+  const choicesToInject = availableAnalysisMethods(items);
+  form!.fields[0]!.props!.choices[0].children[1].props.choices =
```
bangbay-bluetiger, line 182: Instead of passing in `form`, could it just be the object we need modified? Then it would be `whateverThatObjectIs.props.choices = choicesToInject`;

services/ui-src/src/utils/forms/dynamicItemFields.ts
```diff
@@ -172,3 +172,27 @@ const updatedItemChoiceList = (
...
+    choicesToInject;
+  return form;
+};
+
+const availableAnalysisMethods = (items: AnyObject[]) => {
+  const analysisMethodsFieldName =
```
bangbay-bluetiger, line 188: Instead of hardcoding `analysis_method_applicable_standard-plans_nonComplianceAnalyses`, pass it to `availableAnalysisMethods` as an arg.

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
```diff
@@ -16,9 +17,27 @@ export const hasComplianceDetails = (
...
+        return item[analysisMethodsKey].map(
+          (analysisMethod) => analysisMethod.value
+        );
+      }
+
+      return [];
+    })
+    .flat();
+
+  return generateAnalysisMethodFields(updatedForm, analysisMethodsInPlan);
```
bangbay-bluetiger, line 40: Instead of passing in the entire form, filter for the object we need modified by `id` and pass just that in.

services/ui-src/src/utils/forms/dynamicItemFields.ts
```diff
@@ -172,3 +172,18 @@ const updatedItemChoiceList = (
...
   return updatedChoiceList;
 };
+
+// dynamically filter by partialId to find the analysis methods
+export const availableAnalysisMethods = (
+  analysisMethodsFieldId: string,
+  items: string[]
+) => {
+  const updatedItemChoices: AnyObject[] = [];
+  items.forEach((item) => {
```
bangbay-bluetiger, line 182: Can use `map` here:
```
const updatedItemChoices = items.map((item) => ({
  id: `${analysisMethodsFieldId}_${item}`,
  label: item,
}));
```
Although, now I'm seeing why you might've gone with the analysis method `name` instead of the `id`, since you need both if you go with my suggestion to use `id` in the `key` instead of `name`. I think `id` is better to use so `items` here would need to be analysis method objects instead of strings.

bangbay-bluetiger (commented):
It's looking really good and works as expected! Some comments for more typing and robustness.

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
```diff
@@ -16,9 +17,94 @@ export const hasComplianceDetails = (
...
+    });
+
+  //Step 2: Grab all the Plans associated with the selected Standard
+  /*
+   * This is going to grab all of the Analysis Methods associated with the standard as selected on
+   * Part 2: Program-level access and network adequacy standards.
+   * Creates an array that looks like ['Geomapping', 'Plan Provider Directory Review']
+   */
+  const utilizedAnalysisMethodsWithSelectedStandard = standards?.flatMap(
+    (item: { [x: string]: any[] }) => {
```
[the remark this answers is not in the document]
britt-mo: set it to `EntityType`

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
```diff
@@ -16,9 +17,94 @@ export const hasComplianceDetails = (
...
 
-// TODO: Add analysis methods checkboxes used by standard
-export const addAnalysisMethods = (formJson: FormJson) => {
-  return formJson;
+// Add analysis methods checkboxes used by standard
+export const addAnalysisMethods = (
+  form: FormJson,
+  standardKeyPrefix: string,
+  entityId: string,
+  standards: any,
```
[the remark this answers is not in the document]
britt-mo: used EntityShape[]

karla-vm (dismissed):
This is looking great! I found a couple of issues related to this section, but I think we can address those in separate tickets (I'll write those in a second)

The first issue is that we need to disable the "Enter" button if a user hasn't entered data for the Analysis Methods. The error message is showing correctly, but the button lets me enter: 

Screenshot 2025-04-08 at 2 25 44 PM

The second issue is that custom Analysis Methods aren't populating in the checkboxes: 

Screenshot 2025-04-08 at 2 27 04 PM

karla-vm:
Got the tickets for the issues mentioned above here: [CMDCT-4504](https://jiraent.cms.gov/browse/CMDCT-4504) and [CMDCT-4505](https://jiraent.cms.gov/browse/CMDCT-4505)

britt-mo:
> Got the tickets for the issues mentioned above here: [CMDCT-4504](https://jiraent.cms.gov/browse/CMDCT-4504) and [CMDCT-4505](https://jiraent.cms.gov/browse/CMDCT-4505)

@karla-vm pushed up a fix for the custom analysis methods option

bangbay-bluetiger (approved):

gmrabian (approved):
looked good in deployed env

PR #12119, opened 2025-04-16
[CMDCT-4504] Make EntityRow's Enter Button Disabled If No Standards

ntsummers1:
### Description

If you make a plan but don't add a standard, you'll now no longer be able to enter the plan compliance section for that plan until you add a standard.

https://github.com/user-attachments/assets/e6f4108e-7e7e-4d3c-88b4-8cde54b2bcc1

### Related ticket(s)

CMDCT-4504

---
### How to test

Create a NAAAR Report
Add a plan name
Navigate to the Section 3 Plan Compliance
See that the enter button is disabled on both mobile and desktop
Go through the process of creating a Standard
Check back on Section 3's Plan Compliance page and see that the enter button is enabled on both mobile and desktop

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/components/reports/OverlayReportPage.tsx
+++ b/services/ui-src/src/components/reports/OverlayReportPage.tsx
@@ -67,6 +67,7 @@ export const OverlayReportPage = ({
   const TablePage = () => {
     const entityData = report?.fieldData[entityType] || [];
     const standardEntities = report?.fieldData["standards"] || [];
+    const hasStandards = standardEntities.length > 0;
 
     const tableHeaders = () => {
       if (isTablet || isMobile) {
@@ -98,7 +99,7 @@ export const OverlayReportPage = ({
             {parseCustomHtml(errorMessage || "")}
           </Box>
         );
-      } else if (standardEntities.length === 0) {
+      } else if (!hasStandards) {
         return (
           <Box sx={sx.missingEntityMessage}>
             {parseCustomHtml(verbiage.requiredMessages.standards || "")}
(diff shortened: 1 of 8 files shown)


services/ui-src/src/components/reports/OverlayReportPage.tsx
```diff
@@ -134,6 +134,7 @@ export const OverlayReportPage = ({
                   entityType={entityType as EntityType}
                   key={entity.id}
                   locked={undefined}
+                  requiresMoreData={!standardEntities.length}
```
gmrabian, line 137: suggestion: I think I'd prefer to call this parameter `disabled`. That's the pattern we've used elsewhere too when using data or user type, for example, to determine functionality. If only because then when you look at this parent component you know exactly what the parameter being passed to the sub-component does.
[ReportDrawer](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/ui-src/src/components/drawers/ReportDrawer.tsx#L68)
[ModalOverlayReportPage](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/ui-src/src/components/reports/ModalOverlayReportPage.tsx#L195)

gmrabian (commented):
Works!

I think the only thing that's a little funny to me is the text indicating you need to do more changes. 
| With no plans it requires section I | once you add a plan it changes to require section II |
| -- | -- |
| Screenshot 2025-04-17 at 9 32 15 AM | Screenshot 2025-04-17 at 9 32 45 AM |

I don't think it's an issue, just something I noticed.

Would it make more sense to require sections I and II from the start?

services/ui-src/src/components/reports/OverlayReportPage.tsx
ntsummers1: Chatted with Garrett about this - fell on the name hasStandards thats going to be passed. I didn't want to use locked here because locked can be used in other scenarios, and targets other buttons while I need something that happens specifically when theres no standards.

britt-mo (approved):
Looked good in testing

gmrabian (approved):

karla-vm (approved):

PR #12120, opened 2025-04-16
Plan compliance PDF part 5: Calculate exceptions and non-compliance counts

gmrabian:
### Description

- Calculate exceptions and non-compliance counts for display on PDF page
- Move two functions from `PlanComplianceTableOverlay` to shared `naaarPlanCompliance` util for reusability

### Related ticket(s)

CMDCT-4467 part 5

---
### How to test

Already set up in [deployed env](https://d9x1dau963fsg.cloudfront.net/)
- Log in as a state user
- Open the created NAAAR and view the plan compliance section of the PDF
- There are four plans created, each representing a different level of completion and answer path
- Verify the display matches [Figma](https://www.figma.com/design/QcoVlfZ2adUBnzcHiI0G0S/MDCT-MCR-NAAAR?node-id=1356-27098&t=PbV1aWE2f3Kn15Ng-4) for each case

---
### Notes
There is a console error for unique keys, specifically noticed it after I had selected "No" on a 438.68 assurance with two plans.
I have `plan.id` set as the key on the top-level `Box` when mapping plans, so I'm not sure where else to look. However, I don't see that key coming up in the browser inspect either. I tried `Fragment` as well and the key didn't show.

---
### Pre-review checklist

---
#### Review
---


britt-mo (dismissed):
Looked good in deployed branch. 
Not coming up with anything helpful for the response column styling. Only thing that came to mind was to try to set a min-width that matches the figma?

services/ui-src/src/components/export/ExportedPlanOverlayReportSection.tsx
```diff
@@ -119,7 +130,11 @@ const complianceTable = (
             <Text sx={sx.fieldLabel}>{heading}</Text>
             <Text sx={sx.fieldHint}>{parseCustomHtml(hint)}</Text>
           </Td>
-          <Td>{answer}</Td>
+          <Td>
```
jessabean, line 133: Idea for column width: can we add either an `sx` prop or inline style here to fix the width of this column? It looks like the width of the response column is about `50%`

gmrabian: thank you!

jessabean (approved):
🙌🏻

PR #12129, opened 2025-04-23
Add qlty.toml

bangbay-bluetiger:
### Description

QLTY is very noisy. This PR adds a `qlty.toml` file with settings that mirror what we have in `.codeclimate.yml`. This should only enable `duplication` and `function_complexity` checks. This PR also adds the `ripgrep` plugin which will report `TODO`, so the `track-todos` shell script can be removed.

Without a `qlty.toml`, QLTY installs linting plugins with versions that are different than what we use in `.pre-commit-config.yaml`. `.pre-commit-config.yaml` handles all of that for us, so we don't need QLTY to do a second pass.

---
### How to test

Locally:
1. Install [qlty](https://docs.qlty.sh/cli/quickstart)
2. Edit a file that has previously triggered a `duplication` or `function_complexity` warning
    - Examples: [services/app-api/utils/constants/constants.ts](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/app-api/utils/constants/constants.ts), [services/app-api/utils/formTemplates/formTemplates.ts](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/app-api/utils/formTemplates/formTemplates.ts)
3. Run `qlty fmt` and confirm no formatting changed
4. Run `qlty smells` and confirm warnings are reported
5. Run `qlty check --all` and confirm `TODO`s are reported

In GitHub:
1. Create a branch off this PR
2. Edit a file that has previously triggered a `duplication` or `function_complexity` warning
3. Open PR, confirm warnings are reported as comments

---
### Pre-review checklist


britt-mo (approved):
confirmed with a PR that the `duplicate code` alerts showed up as comments

gmrabian (approved):

PR #12132, opened 2025-04-24
mock structured clone in jest globals

gmrabian:
### Description

the version of jest we use does not recognize the `structuredClone` function. upgrading jest is a big lift, and we have been mocking `structuredClone` with its dated equivalent `JSON.parse(JSON.stringify(obj))`

- Include `structuredClone` mock in jest setup such that we need only define it once

---
### How to test

- all ui-src/ tests pass
- all app-api/ tests pass


bangbay-bluetiger (approved):
🙌

PR #12138, opened 2025-04-29
[CMDCT-4441] Authentication Timeout Fix

ntsummers1:
### Description

Its been reported in MFP and MCR that users are not seeing fields be saved when entering data. I've noticed this too while working locally. During my investigation, I found that the CognitoJwtVerifier was sending a call but timing out after 1.5 seconds. CognitoJwtVerifier is part of the aws-jwt-verify package, which actually has received this complaint a few times and has increased the timeout from 1.5 to 3 seconds in the latest 5.0.0 update. They also added the ability to increase the timeout further, noting that you might need to do this based on cold server instances. I've updated the timeout to 6 seconds. This is high, intentionally high. I've run through the application and the longest time for a response back I've had was 4.15. This doesn't seem like a good solution overall, but if it helps users get their information in when saving, I think, if nothing else, this can be a nice band-aid fix.

### Related ticket(s)

CMDCT-4441

---
### How to test

Create a NAAAR report
Right click, inspect, and open the network tab.
Add a plan (It might fail here!)
Update an analysis method (Like Geomapping) (It might fail here!).
Hopefully the above works. Note how long the time it takes for your actions to send and save in the report. You should see 90% of instances taking less than 900ms, and more closer to 300 if not 150ms. 

If the above worked, wait 2 minutes. Yes, 2 minutes. The longer you wait, the more likely the issue can trigger
Try updating a new anaylsis method (Secret shopper) (It might fail here!).
Note how long that took. You're looking to try and trigger a connection that takes longer than 2 seconds but still passes. If it doesn't pass and times out after 6 seconds, this band-aid won't fix the problem.

If the above notes don't make a whole lot of sense, I'm happy to chat about whats happening here and how to trigger it :).

### Notes

---
### Pre-review checklist

### DIFF
--- a/services/app-api/package.json
+++ b/services/app-api/package.json
@@ -33,7 +33,7 @@
     "@aws-sdk/client-ssm": "^3.758.0",
     "@aws-sdk/lib-dynamodb": "^3.758.0",
     "@aws-sdk/util-dynamodb": "^3.758.0",
-    "aws-jwt-verify": "^3.1.0",
+    "aws-jwt-verify": "^5.0.0",
     "cross-spawn": "^7.0.6",
     "date-fns": "^2.26.0",
     "date-fns-tz": "^1.2.2",
--- a/services/app-api/utils/auth/authorization.ts
+++ b/services/app-api/utils/auth/authorization.ts
@@ -4,6 +4,8 @@ import { CognitoJwtVerifier } from "aws-jwt-verify";
 // types
 import { APIGatewayProxyEvent, UserRoles } from "../types";
 import { logger } from "../debugging/debug-lib";
+import { SimpleJwksCache } from "aws-jwt-verify/jwk";
+import { SimpleFetcher } from "aws-jwt-verify/https";
 
 interface DecodedToken {
   "custom:cms_roles": UserRoles;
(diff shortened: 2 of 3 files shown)


services/app-api/utils/auth/authorization.test.ts
```diff
@@ -19,6 +19,14 @@ jest.mock("aws-jwt-verify", () => ({
   },
 }));
 
+jest.mock("aws-jwt-verify/jwk", () => ({ SimpleJwksCache: jest.fn() }), {
+  virtual: true,
```
benmartin-coforma, line 23: A quick comment for posterity.

1. Jest's concept of "virtual modules" (modules that are imported as if they came from a file or NPM, when no such file or package exists) doesn't exactly match up with what we're doing in the code under test. When the code say `import { SimpleJwksCache } from "aws-jwt-verify/jwk"`, it's importing code from an actual file that _does_ exist, within the `aws-jwt-verify` package (which you can confirm, by digging in to your node_modules folder).
2. However, Jest doesn't seem... great? at that type of file-within-a-real-package import. It will go looking for a full package by that name, fail to find one, and throw an error. When we use this mock option to tell Jest the module is virtual, it doesn't go looking for any actual package, and the mock succeeds. I'm not sure how robust this is as a solution - for example, if we needed to spy on `SimpleJwksCache`, or change its mocked behavior, I don't know how well this pattern would work. But it _does_ work well enough for our tests to pass.
3. Within SEDS, when I made this same timeout change, I didn't need to do anything fancy with the mocks; they just worked. I _suspect_ this is because SEDS is using `vitest` instead of `jest`, and I suspect Vite/Vitest is just better at handling this type of import. So if/when we switch MCR to vitest, we may be able to simplify the mocks in this file.

This comment is based on my understanding of the situation, which is fuzzy and incomplete.

benmartin-coforma (approved):

services/app-api/utils/auth/authorization.test.ts
benmartin-coforma: [SEDS' unit tests, for reference](https://github.com/Enterprise-CMCS/macpro-mdct-seds/blob/06a7526678ce90521c791b8f3dae2aa048ba7a19/services/app-api/libs/authorization.test.js)

gmrabian (approved):

PR #12141, opened 2025-04-30
Refactor verbiage

gmrabian:
### Description

We regularly create an enum mapping report types to verbiage files. This change refactors those into one utility that can pull any verbiage file for a given report type.

See notes for exceptions. Open to changes!

---
### How to test

- Log in as a state user
- Create a MCPAR, MLR, and NAAAR report
- Page through each with the console open and verify no new errors show up
- Verify review and submit and PDF pages for all reports have no new errors
- Verify verbiage on key pages like dashboard and review and submit match the report they are for

### Notes

There are a few files that only rely on one verbiage import
- [ExporedEntityDetailsTable](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/ui-src/src/components/export/ExportedEntityDetailsTable.tsx#L18) uses MLR verbiage, but only needs tableHeaders, which are the same for all reports. I didn't see value in refactoring this, but I can
- [ExportedReportFieldTable](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/ui-src/src/components/export/ExportedReportFieldTable.tsx#L22) uses MCPAR verbiage, but also only for tableHeaders
  - maybe we could make tableHeaders universal in a constant? either way I didn't see a reason to refactor these for the same reason
- [ExportedModalDrawerReportSection](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx#L9) uses MCPAR verbiage, but the `modalDrawer` page type is only used in MCPAR at the moment, and we can easily switch to using the verbiage function here later if we need
- [ReportGetStartedPage](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/ui-src/src/components/pages/GetStarted/ReportGetStartedPage.tsx#L19) uses MCPAR verbiage, but it's the only report that has that page and verbiage file
- [StatusTable](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/ui-src/src/components/statusing/StatusTable.tsx#L16) uses MCPAR verbiage, but it only needs the [table caption and headrow](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/ui-src/src/verbiage/pages/mcpar/mcpar-review-and-submit.ts#L37-L40), which are the same for all reports

---
### Pre-review checklist

### DIFF
--- a/services/ui-src/src/utils/other/export.tsx
+++ b/services/ui-src/src/utils/other/export.tsx
@@ -8,6 +8,7 @@ import {
   EntityType,
   FieldChoice,
   FormField,
+  ReportType,
 } from "types";
 // utils
 import {
@@ -17,7 +18,22 @@ import {
   otherSpecify,
 } from "utils";
 // verbiage
-import verbiage from "verbiage/pages/mcpar/mcpar-export";
+import McparVerbiage from "verbiage/pages/mcpar/mcpar-export";
+import MlrVerbiage from "verbiage/pages/mlr/mlr-export";
+import NaaarVerbiage from "verbiage/pages/naaar/naaar-export";
+
+export const getVerbiage = () => {
+  const reportType = localStorage.getItem("selectedReportType");
+  switch (reportType) {
+    case ReportType.MLR:
+      return MlrVerbiage;
+    case ReportType.NAAAR:
+      return NaaarVerbiage;
+    case ReportType.MCPAR:
+    default:
+      return McparVerbiage;
+  }
+};
 
 // checks for type of data cell to be render and calls the appropriate renderer
 export const renderDataCell = (
@@ -39,7 +55,7 @@ export const renderDataCell = (
       entityResponseData = [
         {
           id: uuid(),
-          name: verbiage.missingEntry.missingPlans,
+          name: McparVerbiage.missingEntry.missingPlans,
         },
       ];
     } else {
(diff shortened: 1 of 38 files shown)


services/ui-src/src/utils/other/export.tsx
```diff
@@ -151,97 +140,111 @@ export const renderDrawerDataCell = (
...
 
 export const renderDynamicDataCell = (fieldResponseData: AnyObject) =>
   fieldResponseData?.map((entity: EntityShape) => (
     <Text key={entity.id} sx={sx.dynamicItem}>
       {entity.name}
     </Text>
-  )) ?? <Text sx={sx.noResponse}>{getVerbiage().missingEntry.noResponse}</Text>;
+  )) ?? (
+    <Text sx={sx.noResponse}>
+      {getReportVerbiage().exportVerbiage.missingEntry.noResponse}
```
bangbay-bluetiger, line 245: Nit: Prefer consistency of using a var for this:
```
const { exportVerbiage } = getReportVerbiage();
```

services/ui-src/src/utils/other/export.tsx
```diff
@@ -151,97 +140,111 @@ export const renderDrawerDataCell = (
...
-    let nestedResponses = [];
-    if (
-      formField.id === "plan_ilosUtilizationByPlan" &&
-      fieldResponseData?.length
-    ) {
-      nestedResponses = getNestedIlosResponses(fieldResponseData, entity);
-    }
+) => {
+  const { exportVerbiage } = getReportVerbiage();
+  const { exportVerbiage: mcparExportVerbiage } = getReportVerbiage(
```
bangbay-bluetiger, line 145: These seem a bit confusing to me, but I think it's also because of how the whole method is structured. Could we separate the returns, something like:

```
if (!entityResponseData) {
  const { exportVerbiage: genericVerbiage } = getReportVerbiage();
  return ( <Text sx={sx.noResponse}>{genericVerbiage.missingEntry.noResponse}</Text>);
}

const { exportVerbiage } = getReportVerbiage(ReportType.MCPAR);
return(entityResponseData.map(etc...);

```

bangbay-bluetiger (approved):
Looks great! Non-blocking comments.

britt-mo (approved):
Everything looked good in the browser. Per the notes in the PR description, all of the points made sense to me.

bangbay-bluetiger (approved):

PR #12142, opened 2025-04-30
[CMDCT-4555]: Update dateOptional schema to handle untrue dates

ntsummers1:
### Description

Optional date fields were allowing strings like "42/42/4242" in them, which isn't a real date. This fixes that!

https://github.com/user-attachments/assets/62ddaf93-4f42-4b6e-9a32-19fe68486a65

### Related ticket(s)

CMDCT-4555

---
### How to test

You'll need to fillout the form all the way to plan compliance so you can view the dateoptional field. Inside there you'll be able to play around with it.

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/utils/validation/schemas.test.ts
+++ b/services/ui-src/src/utils/validation/schemas.test.ts
@@ -83,6 +83,7 @@ describe("Schemas", () => {
     "1/1/2",
     "0/0/99",
     "0/01/2023",
+    "42/42/4242",
   ];
 
   const goodValidNumberTestCases = [1, "1", "100000", "1,000,000"];
--- a/services/ui-src/src/utils/validation/schemas.ts
+++ b/services/ui-src/src/utils/validation/schemas.ts
@@ -145,16 +145,7 @@ export const date = () =>
     });
 
 export const dateOptional = () =>
-  string()
-    .nullable()
-    .test({
-      message: error.INVALID_DATE,
-      test: (value) =>
-        value === null ||
-        value === undefined ||
-        isWhitespaceString(value) ||
-        dateFormatRegex.test(value),
-    });
+  string().matches(optionalDateFormatRegex, error.INVALID_DATE).notRequired();
 
 export const endDate = (startDateField: string) =>
   date().test(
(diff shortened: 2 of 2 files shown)


services/ui-src/src/utils/validation/schemas.ts
```diff
@@ -235,3 +226,5 @@ export const nested = (
 // REGEX
 export const dateFormatRegex =
   /^((0[1-9]|1[0-2])\/(0[1-9]|1\d|2\d|3[01])\/(19|20)\d{2})|((0[1-9]|1[0-2])(0[1-9]|1\d|2\d|3[01])(19|20)\d{2})$/;
+export const optionalDateFormatRegex =
```
bangbay-bluetiger, line 229: We could use the same regex for both like this:

```
const datePattern =
  "((0[1-9]|1[0-2])\\/(0[1-9]|1\\d|2\\d|3[01])\\/(19|20)\\d{2})|((0[1-9]|1[0-2])(0[1-9]|1\\d|2\\d|3[01])(19|20)\\d{2})";

export const dateFormatRegex = new RegExp(`^${datePattern}$`);
export const optionalDateFormatRegex = new RegExp(`^(${datePattern})?$`);

```

services/ui-src/src/utils/validation/schemas.test.ts
```diff
@@ -69,20 +69,15 @@ describe("Schemas", () => {
...
   ];
 
-  const goodDateOptionalTestCases = [
-    "",
-    null,
-    undefined,
-    "01/01/2023",
-    "05/15/2023",
-  ];
+  const goodDateOptionalTestCases = ["", "01/01/2023", "05/15/2023"];
```
bangbay-bluetiger, line 72: Why remove the `null` and `undefined` test cases?

ntsummers1: `.nullable()` was previously being used in all of our validations when working with optional fields. We later learned that this was unnecessary to do, and was causing some weird edge cases. Dateoptional never received the update, and thus I've removed it from the validation.

bangbay-bluetiger (approved):

gmrabian (approved):

PR #12154, opened 2025-05-05
Cmdct 4597 - remove filtering of provider types in the standards drawer

britt-mo:
### Description
It was decided that users should still have all provider types available to them in the drawer when they're creating a standard. The provider types and child text fields are now hardcoded in the json file.

### Related ticket(s)

CMDCT-4597

---
### How to test
Create a standard, see that all provider types are available options and the child text field appears upon selection. 
See that there are no changes to drawer functionality in the NAAAR report

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/app-api/forms/naaar.json
+++ b/services/app-api/forms/naaar.json
@@ -588,7 +588,7 @@
         "id": "danas",
         "fields": [
           {
-            "id": "standard_coreProviderTypeCoveredByStandard",
+            "id": "standard_coreProviderType",
             "type": "radio",
             "validation": "radio",
             "props": {
@@ -600,12 +600,12 @@
                   "label": "Primary Care",
                   "children": [
                     {
-                      "id": "standard_coreProviderTypeCoveredByStandard-UZK4hxPVnuYGcIgNzYFHCk",
+                      "id": "standard_coreProviderType_details-UZK4hxPVnuYGcIgNzYFHCk",
                       "type": "text",
                       "validation": {
                         "type": "text",
                         "nested": true,
-                        "parentFieldName": "standard_coreProviderTypeCoveredByStandard",
+                        "parentFieldName": "standard_coreProviderType",
                         "parentOptionId": "standard_UZK4hxPVnuYGcIgNzYFHCk"
                       },
                       "props": {
@@ -619,12 +619,12 @@
                   "label": "Specialist",
                   "children": [
                     {
-                      "id": "standard_coreProviderTypeCoveredByStandard-uITThePQiXntwGGViPTD62",
+                      "id": "standard_coreProviderType_details-uITThePQiXntwGGViPTD62",
                       "type": "text",
                       "validation": {
                         "type": "text",
                         "nested": true,
-                        "parentFieldName": "standard_coreProviderTypeCoveredByStandard",
+                        "parentFieldName": "standard_coreProviderType",
                         "parentOptionId": "standard_uITThePQiXntwGGViPTD62"
                       },
                       "props": {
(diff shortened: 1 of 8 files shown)


services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -75,7 +75,7 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
   const hasPlans = plans?.length > 0;
   const isReportingOnStandards =
     route.path === "/naaar/program-level-access-and-network-adequacy-standards";
-  const hasProviderTypes = report?.fieldData?.["providerTypes"]?.length > 0;
+  // const hasProviderTypes = report?.fieldData?.["providerTypes"]?.length > 0;
```
gmrabian, line 78: remove?

services/app-api/forms/naaar.json
```diff
@@ -594,7 +594,179 @@
...
               "label": "Core provider type covered by standard",
               "hint": "Enter the core provider type the standard applies to. If you wish to specify the provider type, select the core type most suitable and then add the more specific provider type under “Specialist details”.",
-              "choices": []
+              "choices": [
+                {
+                  "id": "standard_UZK4hxPVnuYGcIgNzYFHCk",
+                  "label": "Primary Care",
+                  "children": [
+                    {
+                      "id": "standard_standardDescription-standard_UZK4hxPVnuYGcIgNzYFHCk",
```
gmrabian, line 603: in mcr dev this field is called, for example: 
`standard_coreProviderTypeCoveredByStandard-uITThePQiXntwGGViPTD62-otherText`

and here it is:
`standard_standardDescription-standard_tlzAMYfH4I7iIl5pjqAm7R`

I think we should stick to the `coreProviderTypeCoveredByStandard` key, as it is in the parent. we can leave otherText off I think. So something like:
`standard_coreProviderTypeCoveredByStandard-uITThePQiXntwGGViPTD62`

What do you think?

gmrabian (commented):
works well! a comment on naming scheme

services/app-api/forms/naaar.json
britt-mo: thanks for this, I got really cross-eyed with the ids after a while haha

jessabean: I think that definitely seems cleaner. I wonder a few things (for discussion, non-blocking):

- could we get even more terse with the parent field, like `standard_coreProviderType-{uuid}`?
- I find the key `standard_standardDescription-standard_{uuid}` confusing...both because it contains the term "standard" 3x and because there is another field that contains the term "Description" and it has nothing to do with this field (line 785)

Possible rename?

Screenshot 2025-05-06 at 12 23 44 PM

jessabean (approved):
I have some discussion comments, but nothing that blocks this PR

services/app-api/forms/naaar.json
britt-mo: @jessabean I think that's a good suggestion. One question I have about the `ids` in general: 
I'm using the same UUIDs for the provider types here and in the `provider type coverage` section because I figured they should be the same. Is that correct, or incorrect?

services/ui-src/src/components/reports/DrawerReportPage.tsx
```diff
@@ -259,8 +258,7 @@ export const DrawerReportPage = ({ route, validateOnRender }: Props) => {
     : "";
   const dashTitle = `${verbiage.dashboardTitle}${entityCount}`;
 
-  const isAddStandardsDisabled =
-    !hasProviderTypes || !completedAnalysisMethods();
+  const isAddStandardsDisabled = !completedAnalysisMethods();
```
bangbay-bluetiger, line 261: Nit: Since `isAddStandardsDisabled` doesn't have a conditional anymore, could just use `completedAnalysisMethods()` directly.

PR #12172, opened 2025-05-09
Cmdct 4589 - fix analysis methods logic when plans are deleted

britt-mo:
### Description
As a state user, when I delete a plan, in the plan section, that has previously been selected within the Analysis Methods AND there is no other plan selected for that Analysis method then the analysis method should revert to its unanswered default state 
As a state user, when I delete ALL my plans, then the analysis methods page should revert to its default state

### Related ticket(s)

CMDCT-4589

---
### How to test

https://d3rfe4bkvdu570.cloudfront.net/

Create at least two plans and then create the following analysis methods:

1. one that is not utilized
2. one that is utilized by the first plan only
3. one that is utilized by the second plan only
4. one that is utilized by both plans
5. a custom analysis method utilized by one plan

Verify that:
 when there are multiple plans and only plan 1 is deleted: 
     analysis method #1 is unchanged
     analysis method #2 reverts to unanswered state
     analysis method #3 is unchanged
     analysis method #4 only shows plan 1 being utilized
     the custom method should behave the same way as the required methods

 when there is 1 plan and that plan is deleted: 
     all analysis methods revert to unanswered state

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/fields/DynamicField.test.tsx
+++ b/services/ui-src/src/components/fields/DynamicField.test.tsx
@@ -22,6 +22,7 @@ import {
   mockNaaarReportWithAnalysisMethodsContext,
   mockNaaarAnalysisMethodsReportStore,
   mockAnalysisMethodEntityStore,
+  mockNaaarReportWithAnalysisMethods,
 } from "utils/testing/setupJest";
 import { testA11y } from "utils/testing/commonTests";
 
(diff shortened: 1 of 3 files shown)


gmrabian (commented):
noticed different behavior between default and custom methods after deleting the last plan
1. frequency and plans are not removed from custom methods
  - it probably makes sense to keep frequency and just remove plans? not sure
2. (after adding a new plan) custom methods show up as complete despite not having a plan selected

Screenshot 2025-05-14 at 12 12 02 PM
Screenshot 2025-05-14 at 12 13 59 PM

karla-vm (approved):
Great work, Britt!

PR #12173, opened 2025-05-09
Styling updates for NAAAR plan compliance and MLR reporting pages

gmrabian:
### Description

NAAAR Plan Compliance and MLR Reporting share the `<ResponsiveEntityRow />` component inside a table.

The ask was to make Plan Compliance more like Analysis Methods, which uses a flex to format a pseudo "table".

I've done my best here to bring them closer together, without heavily modifying the MLR Reporting page behavior.

For one, I made mobile the styling breakpoint instead of tablet and mobile. This allows the resizing to behave a little more like Analysis Methods without causing any serious linebreaks or overflows. As such:
- the table top border only goes away at mobile size
- we only use the MobileEntityRow component at mobile size

Other styling changes:
- match table header size, weight, and color to analysis methods (DrawerReportPage) heading
- update EntityRow text to be slightly larger
- update EntityRow Enter button sizing to match DrawerReportPage sizing

| Before | After | Analysis Methods for reference |
| ------ | ----- | -- |
|  Screenshot 2025-05-09 at 2 12 13 PM |  Screenshot 2025-05-09 at 2 12 50 PM |  Screenshot 2025-05-09 at 2 22 26 PM |
| MLR collateral | -- | -- |
| Screenshot 2025-05-09 at 2 14 48 PM | Screenshot 2025-05-09 at 2 14 42 PM | -- |

### Related ticket(s)

CMDCT-4634

---
### How to test

[Deployed env](https://d1rtuzb7qr6rhe.cloudfront.net/)
- Log in as a state user
- Create a NAAAR report
- Complete sections 1 and 2
- Compare the Plan Compliance page at desktop and mobile sizes with Analysis methods. They should behave and look similarly
- At mobile size the Plan Compliance page will break into its own pattern

### Notes

These changes did not touch `MobileEntityRow` as that component differs more signifcantly from the Drawer rows at mobile size. If we think this is worth pursuing, I would make it a separate ticket. As it is, I think the mobile view differing for these two pages is acceptable, as they are both still accessible and performant.

---
### Pre-review checklist

---
#### Review
---


britt-mo (approved):
Looked at this on the deployed site, I think it looked fine

PR #12290, opened 2025-07-02
fix focus indicator

gmrabian:
### Description

Focus indicator needs to be switched to our new mdct styling prefix

How to test:
tab through app and verify pink-ish focus indicator shows!


cy-by (approved):
Approved


https://github.com/user-attachments/assets/ab2c1386-db17-4b29-b569-935350634cb1

PR #12291, opened 2025-07-03
CMDCT-4844: Update MCPAR home accordion text

gmrabian:
### Description

- Update MCPAR state home accordion content
- Remove unused conditionals from TemplateCardAccordion
- Add html parsing for text

This has a nice side effect of eliminating an empty text block at the top of the NAAAR accordion, tightening up the spacing a bit!

### Related ticket(s)

CMDCT-4844

---

### How to test

- Log in as a state user
- Open the MCPAR accordion on the homepage
- Verify the content matches the ticket and the link works

---

### Pre-review checklist

---

#### Review

---

### DIFF
--- a/services/ui-src/src/verbiage/pages/home.ts
+++ b/services/ui-src/src/verbiage/pages/home.ts
@@ -24,19 +24,31 @@ export default {
       },
       accordion: {
         buttonLabel: "When is the MCPAR due?",
-        text: "Due dates vary based on contract year of the managed care program and contract period for the first report.",
-        table: {
-          caption: "MCPAR Due Dates by Contract Year",
-          headRow: ["Contract Year", "Contract Period", "Due Date"],
-          bodyRows: [
-            ["Jan to Dec", "1/1/23 to 12/31/23", "Jun 28, 2024"],
-            ["Apr to Mar", "4/1/23 to 3/31/24", "Sep 27, 2024"],
-            ["Jul to Jun", "7/1/23 to 6/30/24", "Dec 27, 2024"],
-            ["Sep to Aug", "9/1/23 to 8/31/24", "Feb 27, 2025"],
-            ["Oct to Sep", "10/1/23 to 9/30/24", "Mar 29, 2025"],
-            ["Jan to Dec", "1/1/24 to 12/31/24", "Jun 29, 2025"],
-          ],
-        },
+        text: [
+          {
+            type: "p",
+            children: [
+              {
+                type: "html",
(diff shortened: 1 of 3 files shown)


services/ui-src/src/verbiage/pages/home.ts
```diff
@@ -24,19 +24,31 @@ export default {
...
+                type: "html",
+                content: "To view MCPAR due dates select ",
+              },
+              {
+                type: "externalLink",
+                content: "this link",
+                props: {
+                  href: "https://www.medicaid.gov/medicaid/managed-care/guidance/medicaid-and-chip-managed-care-reporting#AMCPR",
+                  target: "_blank",
+                  "aria-label": "MCPAR due dates (link opens in new tab)",
```
cy-by, line 41: ## Problem

This link fails [WCAG 2.5.3: Label in Name (level A)](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html) because the visible text content does not match the accessible name (accName). More info: [Technique F96: Failure due to the accessible name not containing the visible label text](https://www.w3.org/WAI/WCAG21/Techniques/failures/F96).

Content: `This link`
accName: `MCPAR due dates (link opens in new tab)`

This causes issue for folks using voice control. They would see the text as `this link` and try to click it using those words. And because the accName does not contain those words, the link may not be clickable for them.

**You can use Chrome's accessibility tab to view the accessibility tree:**
MCR label in name failure

---

## Solution

If you want to fix this violation having the `aria-label` be: `This link (opens in a new tab)` would do it. However, it would then violate [WCAG 2.4.4: Link Purpose (In Context) (Level A)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html).

To resolve that issue, we need the link text to describe the link destination. Something like:

```html
<p>To view MCPAR due dates, navigate to the <a href="https://www.medicaid.gov/medicaid/managed-care/guidance/medicaid-and-chip-managed-care-reporting#AMCPR">Due Dates sections beneath Managed Care Program Annual Report (MCPAR)</a>. Note, due dates vary based on contract year of the managed care program and contract period for the first report.</p>
```

We don't have to worry about the `(opens in a new tab)` because I'm going to write tickets for MCR to fix all links that open in a new tab with an icon and hidden text. So we can make that improvement with that other ticket.

Apologies that I didn't catch this during planning.

cy-by (changes requested):
Need to finalize content change before this can move on.

cy-by (approved):

PR #12292, opened 2025-07-03
CMDCT-4364 - cdk transition sls side DO NOT MERGE

JonHolman:
### Description
This is the changes we need to make to the serverless setup before we convert to cdk.

### Related ticket(s)

CMDCT-4364

---

### How to test
You can double check that we haven't left out any resources that could or should be retained. At a minimum we need Cognito Userpool, most s3 buckets, all Dynamo Tables, and Cloudfront Distribution.
The tests don't pass because this change breaks the app. That's fine when we are in the moment of transition but that's one reason we should never merge this :)

### Notes

---

### Pre-review checklist

- [X] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- [X] I have updated relevant documentation, if necessary
- [X] I have performed a self-review of my code
- [X] I have manually tested this PR in the deployed cloud environment

---

#### Review

#### Security

---


services/ui/serverless.yml
```diff
@@ -86,7 +84,7 @@ custom:
         firehoseStreamArn=`aws firehose describe-delivery-stream --delivery-stream-name ${self:custom.firehoseStreamName} | jq -r '.DeliveryStreamDescription.DeliveryStreamARN'`
         aws wafv2 put-logging-configuration \
           --logging-configuration ResourceArn=$wafAclArn,LogDestinationConfigs=$firehoseStreamArn \
-          --region ${self:provider.region}
+          --region "us-east-1"
```
peoplespete, line 87: So much more sane! 🙏

src/run.ts
```diff
@@ -1,25 +1,27 @@
+/* eslint-disable multiline-comment-style */
```
peoplespete, line 1: I hate this one too
It's so much easier to deal with comments as `//`

src/run.ts
```diff
@@ -234,13 +261,92 @@ async function destroy_stage(options: {
...
+      `database-${options.stage}-mcpar`,
+      `database-${options.stage}-mlr`,
+      `database-${options.stage}-naaar`,
+      `ui-${options.stage}-cloudfront-logs-${accountId}`,
+      `${accountId}-ui-${options.stage}-waflogs`,
+    ],
   });
 
-  await delete_topics(options);
+  // await delete_topics(options);
```
peoplespete, line 343: topics will be the same in cdk-land right?
This looks right just being sure I understand everything

src/run.ts
```diff
@@ -285,313 +391,27 @@ async function list_topics(options: { stage: string | undefined }) {
...
-  } else {
-    console.log(
-      `Stack ${stackName} delete initiated. Not waiting for completion as --wait is set to false.`
-    );
-  }
-}
-
 /*
- * The command definitions in yargs
+ * The command definitons in yargs
```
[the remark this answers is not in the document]
peoplespete: Or who cares about this one...

peoplespete (dismissed):
LGTM 🔈

src/run.ts
JonHolman: It would be nice if we could flip it in all the `.eslintrc.yaml` files.

PR #12303, opened 2025-07-16
Fix colors (mdct prefix)

gmrabian:
### Description

When we switched our theme prefix to `mdct` the chakra colors stopped working. Switching them over to `mdct` fixes that!

### How to test

- Compare Analysis methods table in this branch to main (borders do not show on main)


cy-by (approved):
Border colors have been corrected.

Inspecting MCR border colors with the web inspector panel open showing the border color value to be --mdct-colors-palette-gray_lighter

PR #12309, opened 2025-07-21
CMDCT-4931 cdk mop up

peoplespete:
### Description
We've convered MCR to CDK. Now we remove all vestiges of it.

### Related ticket(s)
CMDCT-4931
---

### How to test
App should not be affected by these changes. Feel free to test drive it and check that that's true:
https://dwpn1ur0n1277.cloudfront.net/

### Notes
NA

---

### Pre-review checklist

---

#### Review

#### Security

---

### DIFF
--- a/.env.tpl
+++ b/.env.tpl
@@ -2,7 +2,6 @@ COGNITO_USER_POOL_CLIENT_ID=op://mdct_devs/mcr_secrets/COGNITO_USER_POOL_CLIENT_
 COGNITO_USER_POOL_ID=op://mdct_devs/mcr_secrets/COGNITO_USER_POOL_ID
 COGNITO_USER_POOL_CLIENT_DOMAIN=op://mdct_devs/mcr_secrets/COGNITO_USER_POOL_CLIENT_DOMAIN
 COGNITO_IDENTITY_POOL_ID=op://mdct_devs/mcr_secrets/COGNITO_IDENTITY_POOL_ID
-POST_SIGNOUT_REDIRECT=http://localhost:3000/
 REACT_APP_LD_SDK_CLIENT=op://mdct_devs/mcr_secrets/REACT_APP_LD_SDK_CLIENT
 
 # needed for e2e tests


.env.tpl
```diff
@@ -1,23 +1,9 @@
-BannerTable=local-banners
 COGNITO_USER_POOL_CLIENT_ID=op://mdct_devs/mcr_secrets/COGNITO_USER_POOL_CLIENT_ID
 COGNITO_USER_POOL_ID=op://mdct_devs/mcr_secrets/COGNITO_USER_POOL_ID
 COGNITO_USER_POOL_CLIENT_DOMAIN=op://mdct_devs/mcr_secrets/COGNITO_USER_POOL_CLIENT_DOMAIN
 COGNITO_IDENTITY_POOL_ID=op://mdct_devs/mcr_secrets/COGNITO_IDENTITY_POOL_ID
 POST_SIGNOUT_REDIRECT=http://localhost:3000/
```
bangbay-bluetiger, line 5: I think this can be removed, since it's in `services/ui-src/public/env-config.js` now.

JonHolman (approved):

gmrabian (approved):
lgtm. deployed app works.

deployment/stacks/api.ts
JonHolman, line 1: I believe we need to put this stuff back so the authorization.ts can check roles

JonHolman (approved):

.env.tpl
bangbay-bluetiger: I was referring to just the `POST_SIGNOUT_REDIRECT=http://localhost:3000/` line. `POST_SIGNOUT_REDIRECT` sets that value in `services/ui-src/public/env-config.js` from another config?

peoplespete: Ah great point. In that case that value is hardcoded in the `writeLocalUiEnvFile` function, so we don't ever use the env variable. Thanks for noticing! Pushing fix up now.

JonHolman (approved):

PR #12349, opened 2025-08-06
cmdct-4945 - clear console logs

britt-mo:
### Related ticket(s)

CMDCT-4945

---

### How to test
Open the developer console
Login as stateuser
see that there is no console error on the first page (original error was `<p> cannot be a descendant of <p>`)
Create a MCPAR, add a plan, Navigate to /mcpar/plan-level-indicators/program-characteristics
See that there is no console error (original error was `<ul> cannot be a descendant of <p>`

### Notes

---

### Pre-review checklist

---

#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/accordions/TemplateCardAccordion.tsx
+++ b/services/ui-src/src/components/accordions/TemplateCardAccordion.tsx
@@ -11,11 +11,7 @@ export const TemplateCardAccordion = ({ verbiage, ...props }: Props) => {
   return (
     <Accordion sx={sx.root} allowToggle={true} {...props}>
       <AccordionItem label={buttonLabel}>
-        {text && (
-          <Text as="div" sx={sx.text}>
-            {parseCustomHtml(text)}
-          </Text>
-        )}
+        {text && <Text sx={sx.text}>{parseCustomHtml(text)}</Text>}
         {orderedList?.length > 0 && (
           <OrderedList sx={sx.orderedList}>
             {orderedList.map((item: string, index: number) => (


gmrabian (dismissed):

services/app-api/forms/mcpar.json
```diff
@@ -2160,7 +2160,7 @@
                 "validation": "numberSuppressible",
                 "props": {
                   "label": "D1.I.2 Plan share of Medicaid",
-                  "hint": "<p>What is the plan enrollment (within the specific program) as a percentage of the state’s total Medicaid enrollment?</p><ul><li>Numerator: Plan enrollment (D1.I.1)</li><li>Denominator: Statewide Medicaid enrollment (B.I.1)</li></ul>",
+                  "hint": "<span>What is the plan enrollment (regardless of program) as a percentage of total Medicaid enrollment in any type of managed care?</span><span class='fake-list-item'>Numerator: Plan enrollment (D1.I.1)</span><span class='fake-list-item'>Denominator: Statewide Medicaid managed care enrollment (B.I.2)</span>",
```
gmrabian, line 2163: did you mean for this content to change?

jessabean, line 2163: This issue is tricky; I am not sure the ideal solution for users is to "fake" semantic HTML behavior. I think we have 2 paths ahead of us to really solve these console errors (when the error originates from hint text):

1. Work with design and product to reduce hint text to content that would appear in a single `<p>` tag (which probably requires business owner signoff)
2. Work with the CMS design system team to discuss reverting [this decision](https://github.com/CMSgov/design-system/pull/2857) so we can have the flexibility to nest block HTML elements in hint text

Either way, it's a discussion to get to a real solution, rather than band-aid fixes. My gut says we could start with option 2 and see where that gets us...

bangbay-bluetiger: @cy-by looked at how the hint text works for screen readers, and it's flattened into plain text into an `aria-describedby`. So, basically it loses the semantic meaning of using a list. In that case, having a list is really more for visual design than it is for a11y.

But, it would be great to have the `<Hint />` wrapper reverted to `<div />` or have a way to specify what type of wrapper we want to use. Even if we change this particular hint, with a quick regex search, I found 23 instances in MCR where we use `<p>` in our `hint` definitions, so we're probably setting off these nesting errors more than we realize.

gmrabian (dismissed):

services/ui-src/src/components/accordions/TemplateCardAccordion.tsx
```diff
@@ -11,7 +11,11 @@ export const TemplateCardAccordion = ({ verbiage, ...props }: Props) => {
   return (
     <Accordion sx={sx.root} allowToggle={true} {...props}>
       <AccordionItem label={buttonLabel}>
-        {text && <Text sx={sx.text}>{parseCustomHtml(text)}</Text>}
+        {text && (
+          <Text as="div" sx={sx.text}>
```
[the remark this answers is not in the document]
britt-mo: good catch. I'll put in a new PR because there's a separate card for this console error.

PR #12352, opened 2025-08-07
Make upload report actions queue

gmrabian:
### Description

Move upload reports to its own job and give it concurrency condition on its own workflow

### How to test

- Upload report action worked
- Report URL still published in deploy action summary

Concurrency for this is hard to test because it requires a brief action that comes at the end of multiple lengthy actions to trigger at the same time on multiple branches. Though we may not be able to see it happen, we can hope we stop seeing the conflict issue (where the upload job fails because the `gh-pages` branch changed in the time since checkout)

### DIFF
--- a/.github/workflows/upload-e2e-test-reports.yml
+++ b/.github/workflows/upload-e2e-test-reports.yml
@@ -50,4 +50,5 @@ jobs:
       - name: Write URL in Summary
         id: reporturl
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/upload-e2e-test-reports.yml
```diff
@@ -0,0 +1,43 @@
...
+
+on:
+  workflow_call:
+    outputs:
+      result_url:
+        description: "URL to see e2e test results"
+        value: ${{ jobs.upload-reports.outputs.result_url }}
+
+concurrency:
+  group: ${{ github.workflow }}
```
gmrabian, line 11: this is what causes the action to queue if one is already running. we do this with our deploy as well, where its more common to see deploys queue when multiple pushes happen to the same branch

PR #12357, opened 2025-08-11
[CMDCT-4950] Add analysis methods guard for plans autosave

bangbay-bluetiger:
### Description

While working on https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/12337, I came across this bug that could cause analysis methods to be deleted if autosave is implemented improperly on other NAAAR sections.

### Related ticket(s)

CMDCT-4950

---

### How to test

1. Run MCR locally
2. Create a filled NAAAR with yarn db:seed
3. Create a couple more plans and add them to any analysis methods
4. Delete a plan
5. Check that the plan was removed from the analysis method it was added to

---

### Pre-review checklist

---

#### Review

### DIFF
--- a/services/ui-src/src/utils/autosave/autosave.ts
+++ b/services/ui-src/src/utils/autosave/autosave.ts
@@ -122,14 +122,17 @@ export const autosaveFieldData = async ({
   // if there are fields to save, create and send payload
   if (fieldsToSave.length) {
     const reportKeys = { reportType, id, state };
-    let dataToWrite = {} as AnyObject;
+    let dataToWrite = {
+      metadata: { status: ReportStatus.IN_PROGRESS, lastAlteredBy: userName },
+    } as AnyObject;
+
     if (
       entityContext &&
       entityContext.selectedEntity &&
       entityContext.entityType
     ) {
       dataToWrite = {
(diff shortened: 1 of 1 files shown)


services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -161,20 +164,26 @@ export const autosaveFieldData = async ({
       }
     }
 
+    const hasEditedPlans = dataToWrite.fieldData.plans;
+    const hasAnalysisMethods =
```
bangbay-bluetiger, line 168: Since `reportFieldData` is potentially `undefined`, in the previous code:

```
const analysisMethods = reportFieldData?.analysisMethods || [];
```

This would assign an empty array to `analysisMethods`, and later in the code:

```
dataToWrite = {
  ...dataToWrite,
  fieldData: {
    ...dataToWrite.fieldData,
    analysisMethods,
  },
};
```

The empty array would be assigned to `analysisMethods`. But, this shouldn't ever happen because we have default analysis methods assigned to every NAAAR. `reportFieldData` can be `undefined` by not having the current `report`, which seems like it could happen either through lag or improper context set up.

britt-mo (approved):

PR #12364, opened 2025-08-15
Refactor reusable dashboard components and improve accessibility

gmrabian:
### Description

We had some buttons defined for mobile and desktop dashboards that were identical (and already in a utility file!). 
- Make use of those buttons in the mobile view
- Improve screen reader names for the buttons

---

### How to test

[deployed env](https://docm8syadocuj.cloudfront.net/)
- Unit and e2e tests pass
- Test out the dashboard for any report as a state user and an admin
  - Verify all buttons have accessible names
  - Verify desktop and mobile views are the same
  - Verify all buttons work and do what is expected

---

### Pre-review checklist

---


britt-mo (commented):
Does this PR also include the add/copy button? Just wondering because I don't think that button has an accessible name

PR #12374, opened 2025-08-20
CMDCT-5028: Add button theme file

gmrabian:
### Description

- Add theme file for button components that matches [design system](https://www.figma.com/design/lTHypaCnpF0zAHKylxAyxy/MDCT-DesignSystem?node-id=1829-13071)

These definitions do not get imported into our `theme.ts` file because we use variants for some of our buttons and these need to be regression tested. For now we just want these definitions that we can pull in later.

### Related ticket(s)

CMDCT-5028

---

### How to test

- nothing to test in app
- Can scan through and verify colors and sizes line up with design system 🤷 

---

### Pre-review checklist

---

### DIFF
--- a/services/ui-src/src/styles/components/button.ts
+++ b/services/ui-src/src/styles/components/button.ts
@@ -1,0 +1,152 @@
+import { ComponentStyleConfig } from "@chakra-ui/react";
+import { svgFilters } from "styles/theme";
+
+const baseStyles = {
+  transition: "all 0.3s ease",
+  borderRadius: "0.25rem",
+  px: "1.5em",
+  py: "0.5em",
+  fontWeight: "bold",
+  width: "fit-content",
+  "&:disabled, &:disabled:hover": {
+    color: "gray",
+    backgroundColor: "gray_lighter",
+    opacity: "1",
+  },
+  ".mobile &": {
+    fontSize: "sm",
+  },
+};
+
+const solidVariant = {
+  backgroundColor: "primary",
+  color: "white",
+  "&:hover": {
+    backgroundColor: "primary_darker",
+  },
+  "&:disabled, &:disabled:hover": {
+    color: "gray",
+    backgroundColor: "gray_lighter",
+    opacity: 1,
+  },
+};
+
+const solidInverseVariant = {
+  backgroundColor: "white",
+  color: "primary",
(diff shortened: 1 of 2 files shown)


services/ui-src/src/styles/components/button.ts
```diff
@@ -0,0 +1,152 @@
...
+  "&:disabled, &:disabled:hover": {
+    color: "base",
+    backgroundColor: "gray_darker",
+    opacity: 1,
+  },
+};
+
+const outlineVariant = {
+  backgroundColor: "transparent",
+  border: "1px solid",
```
bangbay-bluetiger, line 46: `border` is already a shorthand that includes `borderColor`, so we could combine these two lines. I think it would also avoid potential issues later, such as if these properties don't get defined in the right order. Without a color, `border` would use the browser default color.

In order to pass in the theme colors, would need to write it like this:

```
border: (theme) => `1px solid ${theme.colors.primary}`
```

Alternatively, we could define each part separately:
```
borderColor: "primary",
borderStyle: "solid",
borderWidth: 1,
```

services/ui-src/src/styles/components/button.ts
```diff
@@ -0,0 +1,152 @@
...
+    backgroundColor: "transparent",
+    opacity: 1,
+  },
+};
+
+const ghostVariant = {
+  backgroundColor: "transparent",
+  color: "primary",
+  textDecoration: "underline",
+  padding: "0",
```
bangbay-bluetiger, line 82: Can be `0` without quotes.

services/ui-src/src/styles/components/button.ts
```diff
@@ -0,0 +1,152 @@
...
+  },
+  lg: {
+    fontSize: "lg",
+    px: "1.5em",
+    py: "1em",
+  },
+};
+
+const buttonTheme: ComponentStyleConfig = {
+  baseStyle: baseStyles,
```
bangbay-bluetiger, line 143: Could rename `baseStyles` to `baseStyle` and then use shorthand for the props in here.

bangbay-bluetiger (approved):

bangbay-bluetiger (approved):

britt-mo (approved):
lgtm

PR #12407, opened 2025-09-19
Cmdct 5203 - replace AnyObject with accordion interface

britt-mo:
### Description

From the card:
The Accordion component currently uses AnyObject for the accordionItems prop. Replace the AnyObject type with a specific interface that represents the structure of FAQ accordion items.

### Related ticket(s)

CMDCT-5203

---

### How to test

There shouldn't be any visual difference in the accordions and no errors.

### Notes

---

### Pre-review checklist

---

#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/accordions/FaqAccordion.tsx
+++ b/services/ui-src/src/components/accordions/FaqAccordion.tsx
@@ -2,14 +2,14 @@
 import { Accordion, Box } from "@chakra-ui/react";
 import { AccordionItem } from "components";
 // types
-import { AnyObject, FaqItem } from "types";
+import { FaqItem } from "types";
 // utils
 import { parseCustomHtml } from "utils";
 
 export const FaqAccordion = ({ accordionItems, ...props }: Props) => {
   return (
     <Accordion allowToggle={true} allowMultiple={true} {...props}>
-      {accordionItems.map((item: AnyObject, index: number) => (
+      {accordionItems.map((item: FaqItem, index: number) => (
         <AccordionItem key={index} label={item.question} sx={sx.item}>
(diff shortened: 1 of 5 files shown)


gmrabian (dismissed):
🤔 I wonder if we'd ever want to apply these types to our verbiage files too

services/ui-src/src/components/accordions/FaqAccordion.tsx
```diff
@@ -2,7 +2,7 @@
 import { Accordion, Box } from "@chakra-ui/react";
 import { AccordionItem } from "components";
 // types
-import { AnyObject } from "types";
+import { AnyObject, FaqItem } from "types";
```
jessabean, line 5: We could also replace the AnyObject in the map function on line 12 with FaqItem and then we can remove this AnyObject import altogether! :)

services/ui-src/src/types/other.ts
```diff
@@ -185,3 +185,21 @@ export const assertExhaustive = (_: never): void => {};
 export interface SxObject {
   [key: string]: SystemStyleObject;
 }
+
+export interface FaqItem {
+  question: string;
+  answer: AnswerBlock[];
```
jessabean, line 191: I think instead of creating a new interface, we could reuse the existing CustomHtmlElement interface? So this could be 
```typescript
export interface FaqItem {
  question: string;
  answer: CustomHtmlElement[];
}
```

jessabean (commented):
Couple of suggestions here..

jessabean (approved):
🙌🏻 Approved, with the intention of a fast follow to broaden this into an interface that can be used for all of our custom accordions!

PR #12595, opened 2025-12-22
Report Navigation by URL

BearHanded:
### Description

# Converted back to draft
This PR needs a large ETL for the best implementation, that I have not had time for.
- An ETL will need to remove all report name prefixes from routes in the existing templates like`naaar/myPath`
- This ETL not only needs to address the route per page, but look deep into anywhere we link around the report, which is not a single path. We need to look at warning messages for example and handle those links.
- At that point flatRoutes will only contain `myPath` and half of the translations can be removed, we only ever need to get the path from the route

Deep linking has been an ongoing request in MCR / MFP and I tried to wrap my head around what was stopping us from tackling it.

The goal here was to attempt to minimally change the behavior of the form, while supporting any edge cases I find. Notably I avoided changing the page lifecycle. A component sits in front of loading the pages, and fetches the report before loading.

This to me seems like the shortest path to success that doesn't involve a migration to modify flatPaths and working with dataconnect to change their parsing

## Existing State and Roadblocks
- AppRoutes itemizes all possible routes by building out flatRoutes (we can use variables or *)
- When clicking enter report, we load the report from the API before navigating to the page
- All paths are currently `myReport/myPath/childPath` without reportId, or state, or any other identifying information, so we can't just hit the url.

### URL Structure
AppRoutes just takes `/report/:reportType/:state/:reportId/:pageId?*` and hands it off to a new middle layer that loads the report and hands it off to ReportPageWrapper or ExportPageWrapper
- This logic could be folded into ReportPageWrapper and ExportPageWrapper if you wanted to further flatten the structure, but for the purpose of isolating the logic in this proof of concept, I went with a loader component. This has a spinner while retrieving.

### Functionality Changes (ReportLoader, and translating urls to flatRoutes)
- The report dashboard only navigates you, no pre-loading.
- The new route structure means we need some form of shim for translating between the path in template `mcpar/myPath/other` and the url path `/report/:reportType/:state/:reportId/:pageId?*`
  - pathFormatter is an added util file (feel free to move or rewrite) that has a function for turning the template path into just `myPath/other`, and a function for trimming the first chunk of the url for the same result. With the same format, we can do comparisons and build new routes and links. I would prefer to just ditch the extra `mcpar/` prefix in all the paths but that's more work to update existing forms. 
  - Sidebar, Footer, PrintPDF, and ReportLoader now all use this lookup for their logic
  - Internal Links also get a new component to wrap Router Link and do the same transformation
- Header checked if it was on a report page by checking if we matched a flatRoute. It's simpler now.
- Report Loader bails into 404s if it gets bad params

## Shortcomings and taking it further
- This gets us links to the pages but not to modals or, more noticeably, overlay pages. I think this can still be managed in those components if the links to them change to navigate, and some of the url fragment matching changes to .includes(). The pages then check the end of the url for any specific ids they care about on load. However, my goal was to as much as possible avoid changing any lifecycle behaviors in this proof of concept.
- I don't love the matching logic, of the fragments and url structure. Theoretically the `:reportType` could be the last url param before the template path, and it would match, short of casing. This would imply `:reportId/:reportType` though and that feels like heresy to restful URLs, where a reportId could have multiple types. This is a concession I would leave to the devs of MCR.

The changes themselves aren't massive but it broke a lot of tests.

### Related ticket(s)

CMDCT-

---
### How to test

https://d1ab1m3sa7b3y1.cloudfront.net/
- Look for edge cases that break
- Loading any report type works
- Navigating to any report type by URL alone works
- Continue/Previous buttons work in the footer
- Export to PDF works
- PDF is navigable by URL alone
- Internal Links to other sections like in NAAAR work
- Sidebar Links Work
- Sidebar still knows where you are. If you go to a page, it opens the nested dropdown for you, and shows your current page highlighted
- Header still appears with the Leave button

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


gmrabian (commented):
like what I'm seeing so far! the frequent url mapping is a readability downside, but it's relatively low profile (and not the fault of this PR)

services/ui-src/src/components/app/ReportLoader.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { ReportPageWrapper } from "components/reports/ReportPageWrapper";
+import { ReportContext } from "components/reports/ReportProvider";
+import { useContext, useEffect, useState } from "react";
+import { useNavigate, useParams } from "react-router";
+import { ReportKeys, ReportShape } from "types";
+import { removeReportSpecificPath } from "utils/reports/pathFormatter";
+
+export const ReportLoader = ({ exportView = false }) => {
+  const { reportType, state, reportId, pageId } = useParams();
+  const [isLoading, setIsLoading] = useState<boolean>(true);
```
benmartin-coforma: Nit: I almost always prefer implicit type annotations.

```suggestion
  const [isLoading, setIsLoading] = useState(true);
```

services/ui-src/src/components/reports/ReportPageFooter.tsx
```diff
@@ -1,8 +1,6 @@
-import { useNavigate } from "react-router";
+import { useNavigate, useParams } from "react-router";
 // components
 import { Box, Button, Flex, Image, Spinner } from "@chakra-ui/react";
-// constants
-import { INITIAL_REPORT_ROUTES } from "../../constants";
```
benmartin-coforma: I think this constant is unused now; you may want to go delete it.

ntsummers1 (approved):
Gave this a spin in the deployed env - Worked great! Was a little worried about the overlay pages but they seemed to play nice with this change. Nice update to make deep linking happen!

PR #12715, opened 2026-04-03
[CMDCT-5446]: Disabling fields if user checks not reporting

ntsummers1:
### Description

This PR adds the ability for the user to click on Not Report, then click a reason for not reporting, to have all the child fields after it become disabled. Heres what that looks like in practice:

https://github.com/user-attachments/assets/a9cfc5b9-34b2-49b5-8b26-50535c3e0318

#### Implementation Notes:
A unique approach we did was get as low to the code as possible. This means that the form renderer is responsible for handling this action. There are pros and cons to this, however.
**The Pros:**
- All in 1 file (Moved helper functions out to helper function file)
- Clear line of responsibility
- Watching for data changes and manipulation everything is easy

**The Cons:**
- Reusable? Sure, but can quickly get out of hand if not careful.
- This _is_ low to the code. Business functions likely shouldn't be handled here, but rather in higher up files. That would create a clear separation of responsibilities. 

If we had gone with implementing a json code we would have had to track that variable across _a lot_ of files. That would make this PR quite huge. However, it could make things quite a bit more sustainable and readable for the developer. 

I recognize there are trade offs happening in this PR. Feel free to discuss this implementation and leave your thoughts!

### Related ticket(s)

CMDCT-5446

---
### How to test

- Create a MCPAR Report
- Add a couple plans in A: Add Plans
- Navigate to D: Plan Level Indicators; VII: Quality Measures; Measures and Results
- Add a Measure!   

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/components/fields/ChoiceListField.tsx
+++ b/services/ui-src/src/components/fields/ChoiceListField.tsx
@@ -2,7 +2,7 @@ import { ReactNode, useContext, useEffect, useState } from "react";
 import { FieldValues, useFormContext, UseFormReturn } from "react-hook-form";
 // components
 import { ChoiceList as CmsdsChoiceList } from "@cmsgov/design-system";
-import { Box, SystemStyleObject } from "@chakra-ui/react";
+import { Box, SystemStyleObject, Text } from "@chakra-ui/react";
 import { ReportContext, EntityContext } from "components";
 // types
 import {
@@ -35,6 +35,9 @@ export const ChoiceListField = ({
   sxOverride,
   styleAsOptional,
   clear,
+  disabledStateMessage,
+  disabledStateMessageId,
+  showDisabledStateMessage,
   ...props
 }: Props) => {
   const defaultValue: Choice[] = [];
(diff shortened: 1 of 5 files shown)


britt-mo:
Do we want the fields to clear after selecting from the radio button options?
Screenshot 2026-04-07 at 10 55 25 AM

cy-by (changes requested):
## ❌ `aria-describedby` is on the disabled `<fieldset>`

Image

**Problem:**
Because disabled fields are not keyboard focusable, the described by message is not read out.

**Fix:**
Move the `aria-describedby` from the disabled field to the `<fieldset>` that controls the disabling of other fields.

## ➕ Add `aria-controls` to controlling field

**Problem:**
My bad, this is something that I didn't specify in the Acceptance criteria (AC). 

**Solution:**
Add `aria-control="<id> <id> <etc>"` (the [syntax](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls#:~:text=A%20space%2Dseparated%20list%20of%20one%20or%20more%20ID%20values%20referencing%20the%20elements%20being%20controlled%20by%20the%20current%20element) is a space separated list of `id`s) to the trigger field. Add the id of each field that gets disabled by the controller field.

More info: [aria-controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls#:~:text=When,element,-%2E%20The) can allow Assistive Tech (AT) to clue folks in on which fields the first field controls (e.g. turns disabled).

jessabean (approved):
Pending the aria-disabled changes suggested by Cypress, I think this is good to ship! :shipit:

ntsummers1:
We had an update in requirements! You can find the new PR here: https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/12721

PR #12753, opened 2026-05-27
Handle undefined validation and object validations in print view

ntsummers1:
### Description

**Background:** 

Puerto Rico is reporting for the report 'MLR Platino' in Prod that the print page is throwing an error when they click “Review and Submit” on the print page.

I’ve confirmed that this is true by loading up prod and opening that MLR Report.

The line throwing this error is line 282 in the export.tsx file, specifically the following:
`const validationType = formField.validation.toString();`

**The root issue**

This field here:
```
{
  id: "report_reportingPeriodStartDate",
  type: ReportFormFieldType.DATE,
  props: {
    label: "N. MLR reporting period start date",
    hint: "Auto-populates from program reporting information.",
    disabled: true,
    clear: false,
  },
},
```
does not have a validation field! Regardless of whether of not we want to do an ETL for that (Context, its a pre-filled field that will always have something in it), we should account for this scenario.

**The fix:**

The easiest fix is to simply `??` check for that field and return a string if not found. Based on Bangbay's recommendation, I brought over a function from MFP called isFieldValidationOptional. This has the added benefit of checking for fieldElements on the page. I then updated that code to handle there being no validation object on the field being passed and added appropriate tests to ensure this can't happen again. 

**Video of it working**

https://github.com/user-attachments/assets/b40f9d53-c5d9-4453-9053-3e6f87346812

### Related ticket(s)

[CS5485860](https://cmsitsm.servicenowservices.com/nav_to.do?uri=sn_customerservice_case.do%3Fsys_id=7d9afeda1b4d03508c6687f4604bcbd9%26sysparm_stack=sn_customerservice_case_list.do%3Fsysparm_query=active=true)

---
### How to test

1. Create an MLR Report
2. Navigate to the 'MLR Reporting Page'
3. Click 'Add program reporting information'
4. Add your info
5. Navigate to 'Review and Submit'
6. Click 'Print PDF'
7. See that the page rendered :D 

### Notes

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/utils/other/export.tsx
+++ b/services/ui-src/src/utils/other/export.tsx
@@ -279,7 +279,7 @@ export const renderResponseData = (
     : fieldResponseData;
 
   // check if optional
-  const validationType = formField.validation.toString();
+  const validationType = formField.validation?.toString() ?? "";
   const isOptional = validationType.includes("Optional") || formField?.groupId;
 
   const missingEntryVerbiage = notApplicable
--- a/services/ui-src/src/utils/other/export.tsx
+++ b/services/ui-src/src/utils/other/export.tsx
@@ -15,6 +15,7 @@ import {
 import {
   compareText,
   getReportVerbiage,
+  isFieldValidationOptional,
   maskResponseData,
   otherSpecify,
 } from "utils";
@@ -107,12 +108,8 @@ export const renderOverlayEntityDataCell = (
 
   if (!entity || !entity[fieldId]) {
     const { exportVerbiage } = getReportVerbiage();
-    const validationType =
-      typeof formField.validation === "object"
-        ? formField.validation.type
-        : formField.validation;
 
-    if (validationType.includes("Optional") || formField?.groupId) {
+    if (isFieldValidationOptional(formField) || formField?.groupId) {
       return <Text>{exportVerbiage.missingEntry.noResponseOptional}</Text>;
     } else {
       return (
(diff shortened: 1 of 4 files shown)


services/ui-src/src/utils/other/export.tsx
```diff
@@ -279,7 +279,7 @@ export const renderResponseData = (
     : fieldResponseData;
 
   // check if optional
-  const validationType = formField.validation.toString();
+  const validationType = formField.validation?.toString() ?? "";
```
bangbay-bluetiger: `validation` can also be an object. In MFP, I added this helper because we were doing similar optional validation checks:

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/f2a9039be7a429a5581676f22d48dcb3c43bc8db/services/ui-src/src/utils/forms/forms.ts#L552-L563

```
export const isFieldValidationOptional = (
  formField: FormField | FormLayoutElement
) => {
  if (!isFieldElement(formField)) return false;


  const validationType =
    typeof formField.validation === "object"
      ? formField.validation.type
      : formField.validation;


  return validationType.toLowerCase().includes("optional");
};
```

bangbay-bluetiger: Looking at the helper again, it would probably be better to have:
```
if (isLayoutElement(formField)) return false;
```
instead of:
```
if (!isFieldElement(formField)) return false;
```

Because `isLayoutElement` checks for if `validation` is undefined. If we add this `isFieldValidationOptional` helper to MCR, we could change `isOptional` to:

```
const isOptional = isFieldValidationOptional(formField) || Boolean(formField?.groupId);
```

ntsummers1: Good call outs! I brought over the function and tweaked it a little. I couldn't use isLayoutElement as that made Typescript throw, so I kept isFieldElement and checked for the validation field itself. I then incorporated that into our code!

britt-mo (approved):
Tested and worked as expected

karla-vm (approved):

bangbay-bluetiger (approved):

PR #12761, opened 2026-06-04
[CMDCT-6125] MLR: Update program name form fields

karla-vm:
### Description

Updating the "Program Name" text field to a list of checkbox choices with an optional dynamic field for "Not listed" or "Other" program names within the MLR form.

### Related ticket(s)

CMDCT-6125
CMDCT-6126

---
### How to test

- Log in as a state user
- Create a new MLR report
- Navigate to `/mlr/mlr-reporting`
- Click on the "Add program reporting information"  button

Verify that you see the following fields:

1. A program list (varies by state) with a "Not listed / Other"  checkbox
Screenshot 2026-06-08 at 11 47 51 AM

2. An optional dynamic field where users may add unavailable program names
Screenshot 2026-06-08 at 1 51 25 PM

**NOTE:** This PR does not cover the functionality for displaying these new fields, noted on [this line](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/7387a5fcc866e9e311047ecfe388be768dbf6555/services/ui-src/src/components/tables/EntityRow.tsx#L75).

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


services/app-api/forms/routes/mlr/flags/summer2026SansQm/mlr-reporting.ts
```diff
@@ -73,13 +73,28 @@ export const mlrReportingRoute: ModalOverlayRoute = {
       },
       {
         id: "report_programName",
```
jessabean: For data integrity purposes with DataConnect, I think we should put these changes under a new id. That way `report_programName` can continue to map to a string (legacy reports), and the new id (`report_programNameList`?? `report_programNameArray`??) can just be the new data.

karla-vm: Great call, I did not take that into consideration 🙏🏽

services/app-api/forms/routes/mlr/flags/summer2026SansQm/mlr-reporting.ts
```diff
@@ -72,14 +72,29 @@ export const mlrReportingRoute: ModalOverlayRoute = {
...
+        validation: ValidationType.CHECKBOX,
         props: {
           label: "J. Program name",
-          hint: "Enter the name of the program(s) for this MLR report. A program is defined by a specified set of benefits and eligibility criteria that are articulated in a contract between the state and managed care plans. States providing plan-level MLR results by program should enter the individual program name in the field below. States providing plan-level MLR results across programs should list all relevant program names in the field below, separated by commas.",
+          hint: "Select an existing program for this MLR report. A program is defined by a specified set of benefits and eligibility criteria that are articulated in a contract between the state and managed care plans. States providing plan-level MLR results by program should enter the individual program name in the field below. States providing plan-level MLR results across programs should select all relevant program names.",
+          choices: [],
+        },
+      },
+      {
+        id: "report_otherProgramNameQuestion",
```
bangbay-bluetiger: Although this works visually, I think it causes a problem with accessibility as it isn't tied as a label to the dynamic fields. That's partly also why we refactored [DynamicField](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/components/fields/DynamicField.tsx) in MFP to use `fieldset` and `legend`. Thoughts, @cy-by ?

Image

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -285,6 +285,16 @@ export const DynamicField = ({
     setSelectedRecord(undefined);
   };
 
+  // get verbiage for "Add" button
+  const getButtonVerbiage = () => {
+    if (name.includes("measure_rates")) {
+      return "Add another rate";
+    } else if (name.includes("report_otherProgramName")) {
+      return "Add another program name";
```
bangbay-bluetiger, line 296: This can be a verbiage prop in the field JSON, so we can avoid growing this conditional too much.

services/ui-src/src/utils/forms/forms.ts
```diff
@@ -123,11 +127,33 @@ export const hydrateFormFields = (
...
+      if (
+        field.id === "report_programNameList" &&
+        field.props?.choices.length < 1
+      ) {
+        field.props = {
+          ...field.props,
+          choices: [
+            ...generateProgramListChoices(),
+            {
+              id: uuid(),
```
bangbay-bluetiger, line 151: This will make a different `uuid` each time this is loaded. Should it be the same always?

services/ui-src/src/utils/forms/forms.ts
```diff
@@ -123,11 +127,33 @@ export const hydrateFormFields = (
...
+  };
+
   const fieldsWithChoices = fields.filter(
     (field: FormField | FormLayoutElement) => field.props?.choices
   );
   fieldsWithChoices.forEach((field: FormField | FormLayoutElement) => {
     if (isFieldElement(field)) {
+      if (
+        field.id === "report_programNameList" &&
+        field.props?.choices.length < 1
```
bangbay-bluetiger: I think our style linter prefers `=== 0` statements over `< 1`.

services/ui-src/src/utils/forms/forms.ts
karla-vm: I think we'd like a new one each time because users can add multiple program entities in a single report.

services/app-api/forms/routes/mlr/flags/summer2026SansQm/mlr-reporting.ts
karla-vm: Good call, I just brought those changes over from MFP.

ajaitasaini (dismissed):
been testing this as a result of working on tangential tickets, so functionality wise this lgtm!

services/app-api/forms/routes/mlr/flags/summer2026SansQm/mlr-reporting.ts
bangbay-bluetiger: I see that you changed the container to `fieldset` but aren't using `legend`. Using `fieldset` alone won't address the a11y issue. Rather, we should move the question content to the legend with something like this:

```
<Box as="fieldset" sx={sx.fieldset}>
  <Box
    as="legend"
    className="ds-c-label"
  >
    // groupLabel = "J.a If you don’t see your program listed, please detail the program name as it was during the reporting period."
    {styleAsOptional ? labelTextWithOptional(groupLabel) : groupLabel}
  </Box>
  {displayValues.map((field: EntityShape, index: number) => {
    return (
      <Flex key={field.id} sx={sx.dynamicField}>
        <TextField
          id={field.id}
          name={`${name}[${index}]`}
          label={label}
          errorMessage={fieldErrorState?.[index]?.name?.message}
          onChange={onChangeHandler}
          onBlur={onBlurHandler}
          value={field.name}
          sxOverride={sx.textField}
          {...props}
        />
        ...
```

I'm fine if we address this in another PR and just want to present the visual changes to the BOs, but I don't think using the `ReportFormFieldType.QUESTION` type here is the right way to go.

bangbay-bluetiger (dismissed):

services/app-api/forms/routes/mlr/flags/summer2026SansQm/mlr-reporting.ts
karla-vm: I misunderstood your comment about the `legend` part, I fixed that in the latest push! Apologies.

ajaitasaini (approved):

bangbay-bluetiger (approved):

PR #12799, opened 2026-06-25
CMDCT-6276: [MLR] Nested "Other" program name field

karla-vm:
### Description

Updating `J.a.` in the MLR report to be a nested question under a "Not listed / Other" checkbox option:

Screenshot 2026-06-25 at 11 20 31 AM

### Related ticket(s)

CMDCT-6276

---
### How to test

- Log in as a state user
- Create an MLR report
- Navigate to `/mlr/mlr-reporting`
- Click "Add program reporting information"
- Select "Not listed / Other" under question `J. Program name`

Verify you are able to add and delete other program names with the dynamic field.

### Notes

This is behind the `summer2026SansQm` feature flag.

---
### Pre-review checklist

---
#### Review

#### Security


services/ui-src/src/components/fields/ChoiceListField.tsx
```diff
@@ -170,6 +170,11 @@ export const ChoiceListField = ({
                 form.unregister(child.id);
               }
               break;
+            case "dynamic":
```
bangbay-bluetiger: Would be good to use enum here instead of string: `ReportFormFieldType.DYNAMIC`

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -370,6 +378,7 @@ export const DynamicField = ({
           onClick={appendNewRecord}
         >
           {getButtonVerbiage()}
+          <Image sx={sx.addImage} src={addIcon} alt={getButtonVerbiage()} />
```
bangbay-bluetiger: I feel like this is the first time I'm seeing us put the plus icon after the text. Usually we put it before.

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -370,6 +378,7 @@ export const DynamicField = ({
...
         </Button>
```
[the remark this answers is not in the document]
jessabean: +1 to this suggestion

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -389,16 +398,24 @@ interface Props {
...
+  clear?: boolean;
   [key: string]: any;
 }
 
 const sx = {
   fieldset: {
     marginTop: "spacer3",
   },
+  nestedFieldset: {
+    margin: "1.25rem 0 0 1rem",
```
bangbay-bluetiger: Looks like we could add just values for the top and left margins so we can use at least one of the spacer values:
```
marginLeft: "spacer2",
marginTop: "1.25rem",
paddingLeft: "spacer2",
```

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -301,6 +304,11 @@ export const DynamicField = ({
   // set initial value to form field value or hydration value
   const hydrationValue = props?.hydrate;
   useEffect(() => {
+    if (clear) {
+      form.setValue(name, [], { shouldValidate: true });
+      appendNewRecord();
+      return;
+    }
```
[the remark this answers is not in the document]
bangbay-bluetiger: Agree that this needs a unit test.

bangbay-bluetiger (dismissed):
Functionally, the changes in this PR look good to me. Please follow up on the drop in test coverage in another PR.

Image

services/ui-src/src/components/fields/DynamicField.tsx
karla-vm: That's true! In this case, Jenny requested we put it after.

britt-mo (commented):
Worked as expected, just looks like there are still two codepilot suggestions that haven't been addressed

karla-vm:
> Worked as expected, just looks like there are still two codepilot suggestions that haven't been addressed

Thank you, I missed Copilot's suggestion, just pushed it up 🙏🏽

britt-mo (dismissed):

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -368,8 +376,10 @@ export const DynamicField = ({
           variant="outline"
           sx={sx.appendButton}
           onClick={appendNewRecord}
+          aria-hidden
```
ntsummers1: Was this meant to be added? I believe that this will make the button hidden to screen users

services/ui-src/src/programLists/mlrProgramList.ts
```diff
@@ -823,4 +823,19 @@ export const mlrProgramList: ProgramList = {
...
   id: "6m3NoHsr5qoHrZ2WnkgHfXj7",
   label: "Not listed / Other",
+  children: [
+    {
+      id: "report_otherProgramName",
+      type: ReportFormFieldType.DYNAMIC,
+      validation: {
+        type: ValidationType.DYNAMIC,
+        nested: true,
+        parentFieldName: "report_programNameList",
```
ntsummers1, line 833:
```suggestion
        parentFieldName: "report_programNameList",
        parentOptionId: "6m3NoHsr5qoHrZ2WnkgHfXj7",
```

services/ui-src/src/components/fields/DynamicField.test.tsx
```diff
@@ -732,5 +738,13 @@ describe("<DynamicField />", () => {
     });
   });
 
+  test("Hydrates dynamic fields on clear", async () => {
```
ntsummers1:
```suggestion
  test("Clear existing values and show 1 empty row if clear is true", async () => {
```

services/ui-src/src/components/fields/DynamicField.tsx
karla-vm: @ntsummers1 I was trying to address this comment from co-pilot: https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/12799#discussion_r3477209722, would this not be a good approach?

ntsummers1: Ohhhh, it means adding the aria-hidden label to the addIcon image in the button, not the button itself. So simply moving it over to the image should be the fix we're looking for

services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -301,6 +304,11 @@ export const DynamicField = ({
   // set initial value to form field value or hydration value
   const hydrationValue = props?.hydrate;
   useEffect(() => {
+    if (clear) {
+      form.setValue(name, [], { shouldValidate: true });
```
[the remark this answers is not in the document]
ntsummers1: You might be able to get away with a fix looking more closely like this: 
```
useEffect(() => {
  const currentValues = form.getValues(name);
  if (currentValues?.length) {
    setDisplayValues(currentValues);
    return;
  }
  // Parent option was just (un)checked: start fresh with one empty row.
  if (clear) {
    form.setValue(name, [], { shouldValidate: true });
    appendNewRecord();
    return;
  }
  if (hydrationValue?.length) {
  ...
```

Please give that a test first though. I'm not 100% positive it'll work but it's worth a shot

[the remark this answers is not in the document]
ntsummers1: Stole this from DateField if it helps understand where this set of code is coming from

bangbay-bluetiger (approved):

ajaitasaini (approved):

PR #12800, opened 2026-06-26
[CMDCT-6277] Validate MLR dates are in the past

bangbay-bluetiger:
### Description

- Change the MLR period start date and end date at creation to validate as past dates
- Ensure end date is after the start date
- Change all other MLR dates to validate as past dates (optional)
- Fix a bug in `mapValidationTypesToSchema` that allowed fields to have no validation
    - Added `dropdownOptional` to schema after bug fix
- Reduce overhead by moving schema maps out of separate files and into the same files where the schemas are defined

### Related ticket(s)

CMDCT-6277

---
### How to test

1. https://d45bfyrcukmpl.cloudfront.net/
2. Create an MLR
3. Confirm period start date and end dates must be past dates
4. Confirm end date must after start date
5. After creation, Go to `Enter MLR`
6. Verify all dates need to be past dates

### Notes

Screenshot 2026-06-26 at 2 39 43 PM
Screenshot 2026-06-26 at 2 39 56 PM

---
### Pre-review checklist

---
#### Review


britt-mo:
I was able to create a report with a date in the past: 

Screenshot 2026-06-29 at 12 32 11 PM

karla-vm (dismissed):
Thank you, this looks great!

services/ui-src/src/utils/validation/schemaMap.ts
```diff
@@ -1,34 +1,35 @@
...
-  text: schema.text(),
-  textOptional: schema.textOptional(),
+  checkbox: schema.checkbox(),
+  checkboxOneOptional: schema.checkboxOneOptional(),
+  checkboxOptional: schema.checkboxOptional(),
+  checkboxSingle: schema.checkboxSingle(),
+  date: schema.date(),
+  dateMonthYear: schema.dateMonthYear(),
+  dateOptional: schema.dateOptional(),
+  dropdown: schema.dropdown(),
```
ntsummers1: I think dropdownOptinal is missing from this list, right?

services/app-api/utils/validation/schemaMap.ts
```diff
@@ -241,6 +242,18 @@ export const futureDate = () =>
     }
   );
 
+export const pastDate = () =>
+  date().test(
+    "is-before-current-date",
+    error.INVALID_FUTURE_DATE,
```
ntsummers1: This should be INVALID_PAST_DATE if i'm not mistaken

services/app-api/utils/validation/validation.ts
ntsummers1, line 99: If older MLR Form fields will break because their fieldVAlidation is undefined, shouldn't we add that guard here too?

The guard I'm talking about:
```
/**
       * Legacy: These MLR form fields were created initially without validation
       * because they are auto-populated. They now have validation, but older
       * forms will fail because fieldValidation is undefined. Set validation
       * manually here if it's missing.
       */
      if (
        [
          "report_reportingPeriodStartDate",
          "report_reportingPeriodEndDate",
        ].includes(key) &&
        !fieldValidation
      ) {
        validationSchema[key] = schemaMap[ValidationType.DATE_OPTIONAL];
        return;
      }
```

Or maybe a simpler option is to make the if else section look like this:
```
else if (fieldValidation && fieldValidation.nested) {
      ...
      } else if (fieldValidation?.type === ValidationType.END_DATE) {
      ...
      } else if (fieldValidation?.type === ValidationType.PAST_END_DATE) {
```

bangbay-bluetiger: Yes, I need to add it to the guard here, too.

I wanted to avoid `fieldValidation?` because I think it's a bug to allow it to be undefined. I think `fieldValidation?` was used to fix the breakage without addressing the underlying problem (the fields should not be missing validation).

karla-vm (approved):

ntsummers1 (approved):
Looking good! Nice job Bangbay :)

PR #12802, opened 2026-06-29
cmdct-6203 - add test state ZZ

britt-mo:
### Description

A test state 'ZZ' has been added to all MCR reports.
Also updated README

### Related ticket(s)

CMDCT-6203

---
### How to test
https://d31e9k8x4kxnw3.cloudfront.net/
Confirm that you can login as the new user, and create MCPAR, MLR, and NAAAR reports.
Confirm that when you login as an admin you can see ZZ in the admin state dropdown selector

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,12 @@ MCR pipes updates from fieldData and the report object tables to BigMac for down
 - `services/app-api/handlers/kafka/post/postKafkaData.ts` - Add the bucket and table names into the appropriate arrays. They will be parsed with their event types accordingly.
 - `services/topics/createTopics.js` - Declare the new topic names. Both the stream name for the bucket and table should be added here.
 
+### Test State (ZZ)
+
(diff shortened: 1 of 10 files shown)


karla-vm (dismissed):

ntsummers1 (changes requested):
So this does add the test state ZZ and I can confirm I see it, but theres a couple things left we should do here:

Kafka is going to get notices about this fake state still. Theres 2 places I _think_ you need to add some code to to ensure you don't pass it though kafka:

services/app-api/utils/kafka/kafka-source-lib.ts
```
if (key.split("/")?.[1] === "ZZ") {
  continue;
}
```
Add that code right after `const key: string = s3Record.s3.object.key;`. Should be line 185.

I think you'll also need to update around like 209 that Dynamo shouldn't receive updates on ZZ state reports? I don't know how to add something there though :/. Maybe Jon/Brendan would know?


The other thing I think is missing is the README update is missing from this change. Should see something like:
'Test State (ZZ): ZZ is a non-real state used for testing without IDM. Any new data export, BigMac topic, email, or notification integration must filter out ZZ so test data never reaches downstream consumers. See kafka-source-lib.ts for the reference filter.'

ntsummers1 (dismissed):

karla-vm (dismissed):

ajaitasaini (dismissed):

ajaitasaini (approved):

ntsummers1 (approved):
Heard back from Data Connect that they haven't received ZZ state info! Should be good to merge :D

PR #12835, opened 2026-07-21
CMDCT-6350: Connect dropdown error messages to field

amanda-forrester:
### Description

Error messages for the dropdown field are not connected to the input, therefore Assistive Tech (AT) will not announce the error message. This PR programmatically connects error message to input field via aria-describedby and aria-invalid.

### Related ticket(s)

CMDCT-6350

---
### How to test

- Run this PR locally
- Navigate to NAAAR -> Add/copy NAAAR
- With the Add/copy modal open, [turn on VoiceOver](https://support.apple.com/guide/voiceover/turn-voiceover-on-or-off-vo2682/mac) (if you're on macOS)
- Use the Tab key to navigate to the required Program name field or another required dropdown field
- Once you blur one of the required fields, return to the errored field, you should be able to hear that the error message is read
- You should be able to hear the error message when it's visible and the field is focused

### Notes

See [this PR](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/1476) for similar work in MFP

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/components/fields/DropdownField.test.tsx
+++ b/services/ui-src/src/components/fields/DropdownField.test.tsx
@@ -96,6 +96,15 @@ const dropdownComponentWithYoYCopyNoSubmittedReports = (
   </RouterWrappedComponent>
 );
 
+const dropdownComponentWithHint = (
+  <DropdownField
+    name="testDropdown"
+    label="test-dropdown-label"
+    hint="Select the option that applies"
+    options={mockDropdownOptions}
+  />
+);
+
 describe("<DropdownField />", () => {
   describe("Test DropdownField basic functionality", () => {
     test("Dropdown renders", () => {
@@ -103,6 +112,7 @@ describe("<DropdownField />", () => {
       render(dropdownComponentWithOptions);
       const dropdown = screen.getByLabelText("test-dropdown-label");
       expect(dropdown).toBeVisible();
+      expect(dropdown).toHaveAttribute("aria-invalid", "false");
     });
 
     test("DropdownField calls onChange function successfully", async () => {
@@ -310,15 +320,7 @@ describe("<DropdownField />", () => {
     });
   });
 
-  testA11yAct(dropdownComponentWithOptions, () => {
-    mockGetValues(undefined);
-  });
-
   describe("Dropdown connects error message to select for accessibility", () => {
-    afterEach(() => {
-      jest.clearAllMocks();
-    });
-
     test("Sets aria-invalid and aria-describedby, and renders the error text, when validation fails", () => {
       mockUseFormContext.mockImplementation((): any => ({
         ...mockRhfMethods,
@@ -334,13 +336,29 @@ describe("<DropdownField />", () => {
       const dropdown = screen.getByLabelText("test-dropdown-label");
 
       expect(dropdown).toHaveAttribute("aria-invalid", "true");
+      // no hint present, so aria-describedby only includes the error id
+      expect(dropdown).toHaveAttribute(
+        "aria-describedby",
+        "testDropdown__error"
+      );
+      expect(dropdown).toHaveClass("ds-c-field", "ds-c-field--error");
 
-      const describedById = dropdown.getAttribute("aria-describedby");
-      expect(describedById).toContain("testDropdown__error");
-
-      const errorEl = document.getElementById("testDropdown__error");
-      expect(errorEl).toBeInTheDocument();
-      expect(errorEl).toHaveTextContent("A response is required");
+      const errorEl = screen.getByText("A response is required");
+      expect(errorEl).toHaveAttribute("id", "testDropdown__error");
     });
+
+    test("Includes the hint id in aria-describedby when a hint is present", () => {
+      mockGetValues(undefined);
+      render(dropdownComponentWithHint);
+      const dropdown = screen.getByLabelText("test-dropdown-label");
+      expect(dropdown).toHaveAttribute(
+        "aria-describedby",
+        "testDropdown__error testDropdown-hint"
+      );
+    });
+  });
+
+  testA11yAct(dropdownComponentWithOptions, () => {
+    mockGetValues(undefined);
   });
 });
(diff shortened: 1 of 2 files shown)


services/ui-src/src/components/fields/DropdownField.test.tsx
```diff
@@ -313,4 +313,34 @@ describe("<DropdownField />", () => {
...
+          },
+        },
+      }));
+
+      render(dropdownComponentWithOptions);
+      const dropdown = screen.getByLabelText("test-dropdown-label");
+
+      expect(dropdown).toHaveAttribute("aria-invalid", "true");
+
+      const describedById = dropdown.getAttribute("aria-describedby");
```
bangbay-bluetiger: This assertion can be written like the one above it:
```
expect(dropdown).toHaveAttribute("aria-describedby", "testDropdown__error");
```

services/ui-src/src/components/fields/DropdownField.test.tsx
```diff
@@ -313,4 +313,34 @@ describe("<DropdownField />", () => {
...
+      expect(describedById).toContain("testDropdown__error");
+
+      const errorEl = document.getElementById("testDropdown__error");
```
bangbay-bluetiger: I would avoid `document.getElementById` and use React testing library methods. You could do something like this:
```
const errorEl = screen.getByText("A response is required");
expect(errorEl).toHaveAttribute("id", "testDropdown__error");
```

services/ui-src/src/components/fields/DropdownField.test.tsx
```diff
@@ -313,4 +313,34 @@ describe("<DropdownField />", () => {
   testA11yAct(dropdownComponentWithOptions, () => {
     mockGetValues(undefined);
   });
+
+  describe("Dropdown connects error message to select for accessibility", () => {
+    afterEach(() => {
+      jest.clearAllMocks();
```
bangbay-bluetiger: There aren't any calls that need to be reset in this block, so you can remove this `afterEach`.

services/ui-src/src/components/fields/DropdownField.test.tsx
```diff
@@ -313,4 +313,34 @@ describe("<DropdownField />", () => {
   testA11yAct(dropdownComponentWithOptions, () => {
     mockGetValues(undefined);
   });
+
+  describe("Dropdown connects error message to select for accessibility", () => {
```
bangbay-bluetiger, line 323: Nit: I would move this block before `testA11yAct`, as that's usually the last block in tests.

services/ui-src/src/components/fields/DropdownField.test.tsx
```diff
@@ -313,4 +313,34 @@ describe("<DropdownField />", () => {
...
+          errors: {
+            testDropdown: { value: { message: "A response is required" } },
+          },
+        },
+      }));
+
+      render(dropdownComponentWithOptions);
+      const dropdown = screen.getByLabelText("test-dropdown-label");
+
+      expect(dropdown).toHaveAttribute("aria-invalid", "true");
```
bangbay-bluetiger, line 338: I would add a test for the negative case. You could probably sneak it into one of the other test blocks:
```
expect(dropdown).toHaveAttribute("aria-invalid", "false");
```

services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -214,17 +222,18 @@ export const DropdownField = ({
       <Label htmlFor={name} id={`${name}-label`}>
         {labelText || ""}
       </Label>
-      {parsedHint && <Hint id={name}>{parsedHint}</Hint>}
-      {errorMessage && <InlineError>{errorMessage}</InlineError>}
+      {parsedHint && <Hint id={`${name}-hint`}>{parsedHint}</Hint>}
```
bangbay-bluetiger, line 225: I would also add a test for the hint. You can add a hint prop here:

https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/b2ed1d403c0fbdee076b555c94fe3b43bfef4345/services/ui-src/src/components/fields/DropdownField.test.tsx#L46-L52

Then your assertion can be something like:
```
expect(dropdown).toHaveAttribute("aria-describedby", "testDropdown__error testDropdown-hint");
```

services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -204,7 +210,9 @@ export const DropdownField = ({
   const formErrorState: AnyObject = form?.formState?.errors;
   const errorMessage = formErrorState?.[name]?.value.message;
   const parsedHint = hint && parseCustomHtml(hint);
+  const ariaDescribedBy = `${name}__error${parsedHint ? ` ${name}-hint` : ""}`;
   const nestedChildClasses = nested ? "nested ds-c-choice__checkedChild" : "";
+  const selectClasses = `ds-c-field${errorMessage ? " ds-c-field--error" : ""}`;
```
bangbay-bluetiger, line 215: I'd test for this, too. Could add this to your test block:
```
expect(dropdown).toHaveClass("ds-c-field ds-c-field--error");
```

bangbay-bluetiger (commented):
Thanks for adding the test case! Some suggestions to make it more robust.

amanda-forrester:
> Thanks for adding the test case! Some suggestions to make it more robust.

@bangbay-bluetiger Awesome, thank you! I'm still adjusting to writing FE tests, so I really appreciate the tips!

services/ui-src/src/components/fields/DropdownField.tsx
amanda-forrester: @bangbay-bluetiger I don't think it will break anything by adding the hint prop to `dropdownComponentWithOptions` but it feels a bit clearer to me to go ahead and make a new `const dropdownComponentWithHint` and then add a separate test for the hint text, just to keep the hint testing isolated. Thoughts?

bangbay-bluetiger: That sounds fine to me. You could also an assertion for when there's no hint.

amanda-forrester:
@bangbay-bluetiger I updated the tests!

bangbay-bluetiger (approved):
🎉

cy-by (approved):
Error message correctly being read via `aria-describedby`, tested in [deployed env](https://d3ftx38x62gu64.cloudfront.net/).

Image

ntsummers1 (approved):

PR #12847, opened 2026-07-31
[CMDCT-6442] NAAAR Plan Compliance Fix

ntsummers1:
### Description

So this PR fixes 2 problems, the tier 3 issue & a possible edge case that can trigger the exact same issue. I'll try and break this down as best as possible:

1. Tier 3 Fix

Rhode Island cannot submit their NAAAR. In III. Plan compliance, when they tick **"III.C.1 Plan is non-compliant for this standard"**, the child question **"III.C.2a Plan deficiencies: analyses used to identify deficiencies with this standard"** renders with **no checkboxes at all**, yet still shows a required-field error.

The reason is because the field data show the following:
```
plans[0] = { id: "6235f12-…-38a36a8d7e8c", name: "UnitedHealthcare Insurance Company" }
analysis_method_applicable_plans =  { key: "analysis_method_applicable_plans-6235f12-…-38a36a8d7e8c", value: "RIte Smiles" 
```

When we compare the data, we're looking at the values, not the id's. And obviously RIte Smiles =/= UnitedHealthcare Insurance Company. We should be looking at the ids so that we can match up (6235f12-…-38a36a8d7e8c).

This PR fixes that so they use the ID instead of the name when doing a comparison to make sure we render the correct checkboxes.

"But Nick, how'd this happen in the first place?" I hear you ask. And TBH... I dont know. I'm still investigating. They shouldn't have been able to be different in the first place so we still have an issue elsewhere. This will fix it though for now. 

2. Fun fact, theres a second part to this! You can actually trigger this in a different way! So with 2+ plans a standard can utilize a method that isn't applied to the plan being reported on, and thus one of the plans will have nothing in it! For now, I've temporarily made the question become optional, with the hint text changing to "No analysis methods apply to both this plan and this standard. To report analyses here, return to “I.D Analysis methods” to apply a method to this plan, or to “II. Program-level access and network adequacy standards” to update the methods used for this standard."

### Related ticket(s)

CMDCT-6442 & #CS5632764

---
### How to test

This can replicate part 2 of the above description
#### How to replicate a possible empty question "III.C.2a Plan deficiencies: analyses used to identify deficiencies with this standard" (Do this on the main branch if you want to see the problem, and this branch to see the fix!)

1. Create a NAAAR
2. **I.B Add plans**: Add two rows, `Plan A` and `Plan B`
3. **I.C Provider type coverage**: tick **Dental**.
4. **I.D Analysis methods**:  all 7 must be filled
   - **Geomapping**: Enter → *Yes* → any frequency → tick **Plan A only** → Save
   - **Secret Shopper: Appointment Availability**: → Enter → *Yes* → any frequency → tick **Plan B only** → Save
   - The remaining 5 → Enter → *No* → Save
5. **II. Program-level standards** → **Add standard**:
   - II.A.1 Provider type → **Dental**
   - II.A.3 Standard type → **Maximum time or distance** → II.A.4 description (required, any text)
   - II.A.5 Analysis method(s) utilized → tick **Geomapping only** (only the two "Yes" methods appear here)
   - II.A.6 Population → **Pediatric**; II.A.7 Region → **Statewide** → Save
6. **III. Plan compliance** → **Enter** on **`Plan B`**.
7. III.A.1 → **"No, the plan does not comply on all standards…"**
8. In the 438.68 table row → **Enter** → the standard → **Enter**
9. Tick **"III.C.1 Plan is non-compliant for this standard"**
10. **III.C.2a renders its heading and hint with zero checkboxes.**
11. **Save & return** → nothing happens; a red *"Select at least one response"* appears under III.C.2a with no
    control to satisfy it.

### Notes

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/utils/forms/naaarPlanCompliance.ts
+++ b/services/ui-src/src/utils/forms/naaarPlanCompliance.ts
@@ -62,9 +62,7 @@ export const addAnalysisMethods = (
    */
   const isAppliedToSelectedPlan = (analysisMethod: AnalysisMethodEntity) =>
     (analysisMethod?.analysis_method_applicable_plans ?? []).some(
-      (plan) =>
-        typeof plan?.key === "string" &&
-        plan.key.split(applicablePlansPrefix).pop() === selectedPlanId
+      (plan) => plan.key.split(applicablePlansPrefix).pop() === selectedPlanId
     );
 
   const associatedAnalysisMethodsWithSelectedPlan: AnalysisMethodChoice[] =
--- a/services/ui-src/src/utils/forms/naaarPlanCompliance.ts
+++ b/services/ui-src/src/utils/forms/naaarPlanCompliance.ts
@@ -22,6 +22,21 @@ const applicablePlansPrefix = "analysis_method_applicable_plans-";
 // An analysis method reduced to what the compliance form needs to render a choice
 type AnalysisMethodChoice = { id: string; name?: string };
 
+const isAppliedToPlan = (
+  analysisMethod: AnalysisMethodEntity,
+  planId: string
+) =>
+  (analysisMethod.analysis_method_applicable_plans ?? []).some(
+    (plan) => plan.key === `${applicablePlansPrefix}${planId}`
+  );
+
+const toAnalysisMethodChoice = (
+  analysisMethod: AnalysisMethodEntity
+): AnalysisMethodChoice => ({
+  id: analysisMethod.id,
+  name: analysisMethod.custom_analysis_method_name ?? analysisMethod.name,
+});
+
 export const hasComplianceDetails = (
   exceptionsNonCompliance: string[],
   standardKeyPrefix: string,
@@ -39,7 +54,7 @@ export const addAnalysisMethods = (
   form: FormJson,
   standardKeyPrefix: string,
   selectedStandard: EntityShape,
-  createdAnalysisMethods?: AnalysisMethodEntity[],
+  createdAnalysisMethods: AnalysisMethodEntity[] = [],
   selectedPlanId?: string
 ) => {
   // First we'll create a copy of the form to make future changes to
@@ -60,20 +75,11 @@ export const addAnalysisMethods = (
    * },
    * ]
    */
-  const isAppliedToSelectedPlan = (analysisMethod: AnalysisMethodEntity) =>
-    (analysisMethod?.analysis_method_applicable_plans ?? []).some(
-      (plan) => plan.key.split(applicablePlansPrefix).pop() === selectedPlanId
-    );
-
   const associatedAnalysisMethodsWithSelectedPlan: AnalysisMethodChoice[] =
     selectedPlanId
-      ? (createdAnalysisMethods ?? [])
-          .filter(isAppliedToSelectedPlan)
-          .map((analysisMethod) => ({
-            id: analysisMethod.id,
-            name:
-              analysisMethod.custom_analysis_method_name ?? analysisMethod.name,
-          }))
+      ? createdAnalysisMethods
+          .filter((method) => isAppliedToPlan(method, selectedPlanId))
+          .map(toAnalysisMethodChoice)
       : [];
 
   // Step 2: Grab all the Analysis Methods associated with the selected Standard
--- a/services/ui-src/src/utils/forms/naaarPlanCompliance.ts
+++ b/services/ui-src/src/utils/forms/naaarPlanCompliance.ts
@@ -150,9 +150,7 @@ export const addAnalysisMethods = (
            */
           obj.validation = ValidationType.CHECKBOX_OPTIONAL;
           obj.props.hint =
-            "No analysis methods apply to both this plan and this standard. " +
-            "To report analyses here, return to “I.D Analysis methods” to apply a method to this plan, " +
-            "or to “II. Program-level access and network adequacy standards” to update the methods used for this standard.";
+            "No analysis methods apply to both this plan and this standard. To report analyses here, return to “I.D Analysis methods” to apply a method to this plan, or to “II. Program-level access and network adequacy standards” to update the methods used for this standard.";
         }
       }
     });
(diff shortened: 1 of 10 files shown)


karla-vm (dismissed):
Tested this and confirmed functionality on the deployed branch!

karla-vm (dismissed):

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
```diff
@@ -128,11 +135,23 @@ export const addAnalysisMethods = (
...
+        if (choices.length === 0) {
+          /*
+           * No analysis method is applied to both this plan (I.D) and this standard (II).
+           * Rendering zero checkboxes under a required schema makes the form unsubmittable
+           * with no visible control, so downgrade to optional and explain the empty list.
+           */
+          obj.validation = ValidationType.CHECKBOX_OPTIONAL;
+          obj.props.hint =
+            "No analysis methods apply to both this plan and this standard. To report analyses here, return to “I.D Analysis methods” to apply a method to this plan, or to “II. Program-level access and network adequacy standards” to update the methods used for this standard.";
+        }
```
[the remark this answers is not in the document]
ntsummers1: Noted, but I _am_ trying to catch all choice.length === 0 cases in this output so I don't think this is helpful

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
```diff
@@ -55,20 +60,21 @@ export const addAnalysisMethods = (
...
-        if (method.value == selectedPlanName) {
-          return {
+  const isAppliedToSelectedPlan = (analysisMethod: AnalysisMethodEntity) =>
+    (analysisMethod?.analysis_method_applicable_plans ?? []).some(
+      (plan) => plan.key.split(applicablePlansPrefix).pop() === selectedPlanId
+    );
+
+  const associatedAnalysisMethodsWithSelectedPlan: AnalysisMethodChoice[] =
+    selectedPlanId
+      ? (createdAnalysisMethods ?? [])
```
bangbay-bluetiger: Can this be with the conditional above?

```
const associatedAnalysisMethodsWithSelectedPlan: AnalysisMethodChoice[] =
  selectedPlanId && createdAnalysisMethods
    ? createdAnalysisMethods
        .filter(isAppliedToSelectedPlan)
        .map((analysisMethod) => ({
          id: analysisMethod.id,
          name:
            analysisMethod.custom_analysis_method_name ?? analysisMethod.name,
        }))
    : [];
```

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
```diff
@@ -55,20 +60,21 @@ export const addAnalysisMethods = (
    * },
    * ]
    */
-  const associatedAnalysisMethodsWithSelectedPlan: AnyObject[] =
-    createdAnalysisMethods?.map((analysisMethod: AnyObject) => {
-      if (!analysisMethod?.analysis_method_applicable_plans) return [];
-      for (const method of analysisMethod.analysis_method_applicable_plans) {
-        if (method.value == selectedPlanName) {
-          return {
+  const isAppliedToSelectedPlan = (analysisMethod: AnalysisMethodEntity) =>
```
bangbay-bluetiger: I think it would help with readability to have an early return:

```
  const isAppliedToSelectedPlan = (analysisMethod: AnalysisMethodEntity) => {
    const plans = analysisMethod?.analysis_method_applicable_plans;
    if (!plans) return false;

    return plans.some(
      (plan) => plan.key.split(applicablePlansPrefix).pop() === selectedPlanId
    );
  };

```

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
```diff
@@ -102,12 +108,13 @@ export const addAnalysisMethods = (
    * ]
    * Since Geomapping is an Analysis Method in both the Plan and the Standard
    */
-  const associatedMethodsBetweenStandardsAndPlan =
-    associatedAnalysisMethodsWithSelectedPlan?.filter((plan) =>
-      utilizedAnalysisMethodsWithSelectedStandard?.some((standard) =>
-        standard?.key?.endsWith(plan.id)
+  const associatedMethodsBetweenStandardsAndPlan = analysisMethodsKey
```
bangbay-bluetiger: Does this need to check for `analysisMethodsKey`? Looks to me that above it `utilizedAnalysisMethodsWithSelectedStandard` starts off as an empty array and only has values if `anaysisMethodsKey` exists.

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
```diff
@@ -102,12 +108,13 @@ export const addAnalysisMethods = (
...
+    ? associatedAnalysisMethodsWithSelectedPlan.filter((method) =>
+        utilizedAnalysisMethodsWithSelectedStandard?.some(
```
bangbay-bluetiger: I think `utilizedAnalysisMethodsWithSelectedStandard?` can have `?` removed since it'll always have a value.

bangbay-bluetiger (dismissed):
Tested and looks good to me. Non-blocking comments for code readability.

bangbay-bluetiger (dismissed):

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
ntsummers1: Broke this out a little more. Hopefully that makes it more readable. Good call out!

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
ntsummers1: It does not! Good catch

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
ntsummers1: Made happen :)

services/ui-src/src/utils/forms/naaarPlanCompliance.ts
ntsummers1: See above comment. Tried to make both functions a little easier!

bangbay-bluetiger (approved):

PR #12851, opened 2026-08-05
[CMDCT-5231] Pre-load quality measures by state

bangbay-bluetiger:
### Description

This PR builds upon work from https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/12563 to pre-load MCPAR quality measures based on state and program.

Changes:

- Store quality measures in separate files per state, to keep each file smaller and more maintainable
- Add tests to compare the number of programs and measures to expected count (count taken from CSV)
- Add tests to compare program names in `services/app-api/utils/data/mcparQualityMeasuresList.ts` to `services/ui-src/src/programLists/mcparProgramList.ts`, so that they match up
- Update MCPAR seed fixture with new fields and to look up quality measures by program correctly

Screenshot 2026-08-05 at 11 25 07 AM

### Related ticket(s)

CMDCT-5231

---
### How to test

1. Run MCR locally or go to https://d3vsc8p0lxw0lj.cloudfront.net/
2. Login as `stateuser@test.com`
3. Create a MCPAR
4. Select `Minnesota Senior Care Plus (MSC+)`, `PMAP`, or `Special Needs Basic Care (SNBC)` as your program (`Minnesota Senior Health Options (MSHO)` has no pre-loaded measures)
5. Go to Section D: Plan-Level Indicators -> VII: Quality Measures -> Measures and Results
6. You should see measures matching these values for the program you chose:

https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/3231318cbe080d4897b1382d0cd3a91eb154f34e/services/app-api/utils/data/qualityMeasures/minnesota.ts#L5-L365

### Notes

To convert the CSV to JavaScript, I used AI to create a script that I could run locally. The initial output of the conversion was to a single file (~15,000 lines). I cut and paste each state section by hand to its individual file.

For help identifying typos, I used [cspell](https://github.com/streetsidesoftware/cspell) on the state files.

---
### Pre-review checklist

---
#### Review


services/app-api/utils/data/mcparQualityMeasuresList.test.ts
```diff
@@ -0,0 +1,268 @@
+import { State } from "../types";
+import { mcparQualityMeasuresList } from "./mcparQualityMeasuresList";
+import { mcparProgramList } from "../../../ui-src/src/programLists/mcparProgramList";
```
bangbay-bluetiger, line 3: Imported the list from `ui-src` so that we don't have to duplicate it in `app-api`.

services/app-api/tsconfig.json
```diff
@@ -24,5 +24,6 @@
   "ts-node": {
     "files": true
   },
-  "include": ["db/custom.d.ts", "handlers", "utils"]
+  "include": ["db/custom.d.ts", "handlers", "utils"],
+  "exclude": ["utils/data/mcparQualityMeasuresList.test.ts"]
```
bangbay-bluetiger, line 28: Excluded `mcparQualityMeasuresList.test.ts` because importing `mcparProgramList` from `ui-src` would cause the build to fail from `rootDir` issues. This seemed to be an okay trade-off so that we could reuse `mcparProgramList`.

britt-mo (commented):
For my own understanding, this change is only for states and DC? Not for territories?
Also just to add for documentation here, I already asked about whether the ability to edit the program type after form creation should be disabled now with the change to measures. Bangbay asked Anna, and she'll see what the BOs say.

bangbay-bluetiger:
> For my own understanding, this change is only for states and DC? Not for territories? Also just to add for documentation here, I already asked about whether the ability to edit the program type after form creation should be disabled now with the change to measures. Bangbay asked Anna, and she'll see what the BOs say.

I don't know that the territories won't ever have pre-loaded measures, but we could update the states constant to include them just in case. MFP has them listed:

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/app-api/utils/types/other.ts#L66-L126

services/app-api/utils/data/qualityMeasures/colorado.ts
```diff
@@ -0,0 +1,262 @@
...
+      measure_name: "Contraceptive Care - All Women (HEDIS CCW)",
+      measure_identifier: measureIdentifiers.yes,
+      measure_identifierCmit: "1002",
+    },
+    {
+      measure_name:
+        "Adherence To Antipsychotic Medications For Individuals With Schizophrenia (HEDIS SAA)",
+      measure_identifier: measureIdentifiers.yes,
+      measure_identifierCmit: "18",
+    },
```
ntsummers1, line 141: This is duplicated, I think? It looks like theres a few of them in here (22 from my count, though I could be wrong) that are duplicated.

services/app-api/utils/data/qualityMeasures/michigan.ts
```diff
@@ -0,0 +1,896 @@
...
+        "Glycemic Status Assessment for Patients With Diabetes (HEDIS GSD)",
+      measure_identifier: measureIdentifiers.yes,
+      measure_identifierCmit: "1820",
+    },
+    {
+      measure_name:
+        "HCBS CAHPS: Choosing the Services that Matter to You Composite",
+      measure_identifier: measureIdentifiers.yes,
+      measure_identifierCmit: "95-01",
+    },
```
ntsummers1, line 177: Are we allowed to accept these values?

services/app-api/utils/data/mcparQualityMeasuresList.test.ts
ntsummers1: Unfortunately I believe we need this duplicated across both. Thats a downside of our architecture. Crossing over between ui-src/app-api/ui/etc causes issues with the tests not accurately running and I believe has implications on deployed branches.

ntsummers1 (commented):
Leaving some comments for now - Still testing

services/app-api/utils/data/qualityMeasures/colorado.ts
bangbay-bluetiger: In the CSV, this measure name is there 38 times across 24 states/programs. I think it's correct to see it duplicated.

services/app-api/utils/data/qualityMeasures/michigan.ts
bangbay-bluetiger: We have `measure_identifierCmit` as a number field, but it looks like several CMIT values in the CSV have dashes. We should check if that means we need to change the field type.

services/app-api/utils/data/mcparQualityMeasuresList.test.ts
bangbay-bluetiger: I tried addressing that by excluding this test file from the build. I thought it would be an okay trade-off, since this test file doesn't test code, just that what we've entered in our data files match what we expect. It was more helpful when I first started the task because it helped find mismatches between the CSV and what we already had in the program list, which were all typos.

services/app-api/utils/types/data.ts
```diff
@@ -0,0 +1,17 @@
+import { State } from "./other";
+
+type Measure = {
+  measure_name: string;
+  measure_identifier?: any;
```
ntsummers1: Does swapping the any type here to Choice[] make sense?

ntsummers1:
I think I'm seeing a bug with Copy over that use reports that were created with the flag turned off, and then are copied while the flag is turned on:

**With the flag turned off**
- Login as Stateuser@test.com
- Create a new MCPAR report making sure you select PMAP
- Fill out the report, but in Section D: Topic VII: Quality & Performance Measures make sure you add one measure so you have an old measure.
- Submit the report

**With the flag turned on**
- Select Existing Report and PMAP and choose the old report you created
- Open Section D: Topic VII: Quality & Performance Measures and see that the table is empty. The old format measure is gone but also none of the pmap measures were added.

services/app-api/utils/data/qualityMeasures/michigan.ts
bangbay-bluetiger: We also have `measure_identifierCbe` as a number field but all of the pre-loaded measures with CBE numbers have dots in them. I think it'd be safer to change the fields to text fields.

bangbay-bluetiger: Looking at https://p4qm.org/measures and https://cmit.cms.gov/cmit/, I definitely see CMIT/CBE numbers with dashes or alpha characters.

services/app-api/utils/types/data.ts
bangbay-bluetiger: I made it `[Choice]` since it should only be one choice value in an array.

karla-vm (approved):
Tested on the deployed branch and read through the comments, I think this looks good.

ajaitasaini (approved):
tested with SNBC and saw corresponding measures!

PR #12854, opened 2026-08-06
CMDCT-6382: NAAAR content findings

karla-vm:
### Description

Updating content for the NAAAR, which includes a new frontend validation to display a custom checkbox error message for the Plan Compliance section questions `III.B.2` through `III.B.4`

### Related ticket(s)

CMDCT-6382

---
### How to test

- Log in as a state user
- Create a NAAAR report
- Add a plan
- Select any provider type coverage
- For at least one (1) of the analysis methods, select "Yes" and enter the frequency of analysis
- Add a standard and select the analysis method from the previous step
- Navigate to `/naaar/plan-compliance`
- Enter a plan
- For question **B. Assurance of plan compliance for 438.206**, select "No, the plan does not comply with all standards based on all analyses or exceptions granted"
- Enter the nested question
- Scroll down and click **Save & return** without entering any data

You should see the following error message:

Screenshot 2026-08-06 at 2 30 14 PM

If you select at least one (1) checkbox for questions `III.B.2` through `III.B.4`, the error message should go away.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/app-api/utils/validation/validation.ts
+++ b/services/app-api/utils/validation/validation.ts
@@ -99,9 +99,13 @@ export const mapValidationTypesToSchema = (fieldValidationTypes: AnyObject) => {
       // if standard validation type, set corresponding schema from map
       if (typeof fieldValidation === "string") {
         const correspondingSchema = schemaMap[fieldValidation];
-        if (correspondingSchema) {
-          validationSchema[key] = correspondingSchema;
-        }
+        if (!correspondingSchema) return;
+
+        validationSchema[key] =
+          typeof correspondingSchema === "function"
+            ? correspondingSchema()
+            : correspondingSchema;
+
         return;
       }
 
@@ -134,9 +138,10 @@ export const makeNestedFieldSchema = (fieldValidationObject: AnyObject) => {
     );
   } else {
     const correspondingSchema = schemaMap[type];
-    const fieldValidationSchema = options
-      ? correspondingSchema(options)
-      : correspondingSchema;
+    const fieldValidationSchema =
+      typeof correspondingSchema === "function"
+        ? correspondingSchema(options)
+        : correspondingSchema;
     return nested(() => fieldValidationSchema, parentFieldName, parentOptionId);
   }
 };
--- a/services/app-api/utils/validation/validation.ts
+++ b/services/app-api/utils/validation/validation.ts
@@ -119,6 +119,7 @@ export const mapValidationTypesToSchema = (fieldValidationTypes: AnyObject) => {
       const getSchema = dependentSchemas[fieldValidation.type];
       if (getSchema) {
         validationSchema[key] = getSchema(fieldValidation);
+        return;
       }
     }
   );
@@ -127,8 +128,7 @@ export const mapValidationTypesToSchema = (fieldValidationTypes: AnyObject) => {
 
 // return created nested field schema
 export const makeNestedFieldSchema = (fieldValidationObject: AnyObject) => {
-  const { options, type, parentFieldName, parentOptionId } =
-    fieldValidationObject;
+  const { type, parentFieldName, parentOptionId } = fieldValidationObject;
   const getSchema = dependentSchemas[type];
   if (getSchema) {
     return nested(
@@ -137,11 +137,7 @@ export const makeNestedFieldSchema = (fieldValidationObject: AnyObject) => {
       parentOptionId
     );
   } else {
-    const correspondingSchema = schemaMap[type];
-    const fieldValidationSchema =
-      typeof correspondingSchema === "function"
-        ? correspondingSchema(options)
-        : correspondingSchema;
+    const fieldValidationSchema = schemaMap[type];
     return nested(() => fieldValidationSchema, parentFieldName, parentOptionId);
   }
 };
--- a/services/app-api/utils/validation/validation.ts
+++ b/services/app-api/utils/validation/validation.ts
@@ -123,7 +123,8 @@ export const mapValidationTypesToSchema = (fieldValidationTypes: AnyObject) => {
 
 // return created nested field schema
 export const makeNestedFieldSchema = (fieldValidationObject: AnyObject) => {
-  const { type, parentFieldName, parentOptionId } = fieldValidationObject;
+  const { options, type, parentFieldName, parentOptionId } =
+    fieldValidationObject;
   const getSchema = dependentSchemas[type];
   if (getSchema) {
     return nested(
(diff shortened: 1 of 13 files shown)


services/ui-src/src/utils/validation/schemas.ts
```diff
@@ -391,6 +396,7 @@ export const schemaMap: any = {
   checkboxOneOptional: checkboxOneOptional(),
   checkboxOptional: checkboxOptional(),
   checkboxSingle: checkboxSingle(),
+  checkboxPlanCompliance: checkboxPlanCompliance(),
```
bangbay-bluetiger: Alternatively, instead of creating a new validation schema, we could modify the checkbox schema to accept `options`. This is something that we've done for a few schemas in MFP.

First modify validations to accept options:
https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/utils/validation/validation.ts#L22-L27
https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/utils/validation/validation.ts#L63-L65

Then, pass in options in the route. Instead of `maxLength` and `TEXT_CUSTOM`, we'd pass in `errorMessage` and `CHECKBOX`:
https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/app-api/forms/routes/wp/flags/wpSarRelease2025/state-or-territory-specific-initiatives/initiatives.ts#L227-L232

Then, pass in options in the schemaMap:
https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/utils/validation/schemas.ts#L354

Then, use the options in the schema definition. In this case, we'd look for an `errorMessage` in options, but if it doesn't exist, fall back to the default message:
https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/utils/validation/schemas.ts#L43-L47

This is also a pattern we can consider using for future custom error messaging. Thoughts?

britt-mo (approved):
Tested in ephemeral env and worked as described

ntsummers1:
Think I have bugs/things to fix. Investigating

services/app-api/utils/validation/validation.ts
```diff
@@ -105,8 +105,17 @@ export const mapValidationTypesToSchema = (fieldValidationTypes: AnyObject) => {
...
+        const correspondingSchema = schemaMap[fieldValidation.type];
+        if (correspondingSchema) {
+          validationSchema[key] = correspondingSchema(fieldValidation.options);
+        }
+        return;
+      }
+
       // if nested validation type, make and set nested schema
-      if (fieldValidation.nested) {
+      else if (fieldValidation.nested) {
```
bangbay-bluetiger, line 118: We'll need to add this block in `makeNestedFieldSchema` if we ever use the new type (or any other validation with options) on a nested field:

```
const fieldValidationSchema = typeof correspondingSchema === "function"
  ? correspondingSchema(options)
  : correspondingSchema;
```
I can see it being necessary if we port over the text character limit validation, but it doesn't need to happen in this PR.

services/ui-src/src/utils/validation/schemas.ts
```diff
@@ -319,6 +319,11 @@ export const checkbox = () =>
     .min(1, error.REQUIRED_CHECKBOX)
     .of(object({ key: text(), value: text() }))
     .required(error.REQUIRED_CHECKBOX);
+export const checkboxCustom = (options: ChoiceOptions) =>
+  array()
+    .min(1, options.errorMessage)
```
bangbay-bluetiger, line 324: Since `errorMessage` is defined as optional in `ChoiceOptions`, might want to add a fallback message here.

ntsummers1:
Bug found: 

If you mark the section as "Yes, the plan complies on all standards based on all analyses" and save it and leave. When you come back and click "No, the plan does not comply with all standards based on all analyses or exceptions granted" the form crashes. Confirmed against the dev branch this doesn't currently happen

https://github.com/user-attachments/assets/2f579dc4-a189-42a3-9695-296fc48fb7c3

bangbay-bluetiger (approved):
Thanks for using the options pattern! A couple of non-blocking comments.

amanda-forrester:
> Bug found:
> 
> If you mark the section as "Yes, the plan complies on all standards based on all analyses" and save it and leave. When you come back and click "No, the plan does not comply with all standards based on all analyses or exceptions granted" the form crashes. Confirmed against the dev branch this doesn't currently happen
> 
>  Screen.Recording.2026-08-10.at.4.46.44.PM.mov

@ntsummers1 It didn't crash for me when I left the page and came back and clicked "No,...", I checked out the branch, did not use the ephemeral env. Not sure what else could be different, but I will keep poking around!

ntsummers1:
> > Bug found:
> > If you mark the section as "Yes, the plan complies on all standards based on all analyses" and save it and leave. When you come back and click "No, the plan does not comply with all standards based on all analyses or exceptions granted" the form crashes. Confirmed against the dev branch this doesn't currently happen
> > Screen.Recording.2026-08-10.at.4.46.44.PM.mov
> 
> @ntsummers1 It didn't crash for me when I left the page and came back and clicked "No,...", I checked out the branch, did not use the ephemeral env. Not sure what else could be different, but I will keep poking around!

Interesting.. Can anyone verify in the deployed branch it also crashes for them? Steps here:

- Login as Stateuser1@test.com
- Open the 
"Accountable Care Organization Partnership Plan (ACPP) | PAHP | 08/10/2026 | Sammy States | In revision" report. (The first one in the list, not the second!
- Navigate to the Plan Compliance page
- Enter "SB"
- For B, Select "No"
- Attempt to enter B

amanda-forrester (commented):
It worked for me! I'll be curious about the bug Nick found when he was testing.

bangbay-bluetiger:
> Bug found:
> 
> If you mark the section as "Yes, the plan complies on all standards based on all analyses" and save it and leave. When you come back and click "No, the plan does not comply with all standards based on all analyses or exceptions granted" the form crashes. Confirmed against the dev branch this doesn't currently happen
> 
>  Screen.Recording.2026-08-10.at.4.46.44.PM.mov

@ntsummers1 I looked at that report and it looks like you created it when the deployed code was still using `checkbox` as the validation type with `options`. The finalized version of the code removed `options` from `checkbox` and instead uses `checkboxCustom` as the validation type. This shouldn't be an issue with existing reports outside of this branch, as `checkbox` with `options` has never been used before.

Screenshot 2026-08-10 at 7 00 10 PM

karla-vm:
> Thanks for using the options pattern! A couple of non-blocking comments.

I'll make note of these for a follow-up PR! thank you, Bangbay!

